PR #89, opened 2026-04-29
feat(client): self-upgrade CronJob (closes #69)

### LINKED ISSUE #69: Helm chart auto-update — deployed clients stay frozen on installed version
## Problem

When a customer deploys the tracebloc client via `helm install`, their cluster is pinned to whatever chart version was current at install time. Once the workspace is running, there is **no mechanism** to:

- Poll the published Helm repo (`https://tracebloc.github.io/client`) for newer chart versions
- Notify the operator when a new version is available
- Apply chart upgrades automatically

The image-tag layer inside the chart partially handles in-place updates: when `images.jobsManager.digest` is unset, `imagePullPolicy: Always` re-pulls the floating `prod`/`dev` tag — but only when a pod restarts, and pods do not restart on their own. So image updates are also bottlenecked behind a chart upgrade trigger that never fires.

Net result: customers sit on the version they first installed (e.g. `client` 1.0.8) indefinitely, missing security fixes, MySQL stability fixes (v1.1.0), and other improvements.

## Where

- `tracebloc/client` Helm chart (`client/`)
- Install entry: `scripts/lib/install-client-helm.sh` runs `helm repo update && helm upgrade --install` once at install time only — never again

## Fix

Ship a chart-side mechanism that closes the loop. Headline approach: a CronJob inside the chart that:

1. Periodically (e.g. daily) runs `helm repo update` against `tracebloc.github.io/client`
2. Compares the published chart version against the deployed Release version
3. Either notifies the operator (workspace UI banner / email / `kubectl describe`) or self-applies, based on the chosen UX
4. Reports the result back to the tracebloc backend so the version is visible in the workspace UI

## Options

- **(A) Notify-and-approve** — CronJob detects a new version, surfaces it in the workspace UI / emails the admin, customer clicks "apply". Safer for breaking changes; aligns with how enterprise infrastructure typically handles updates.
- **(B) Auto-upgrade** — CronJob runs `helm upgrade` directly when a new chart is detected. Lowest friction, highest risk for breaking-change rollouts on customer infra.
- **(C) Hybrid** — auto-apply patch versions (1.1.0 → 1.1.1), notify-and-approve for minor/major bumps. Closest to standard SaaS update patterns.

## Why

Every existing customer is silently drifting away from the latest secure, stable version. Security work (training-pod hardening from #57, the security-docs branch) and stability fixes (MySQL kill-loop fix in v1.1.0) only land for *new* deployments. There is no remediation channel for already-deployed clients short of asking each customer to run `helm upgrade` manually.

## Benefits

- Existing customers receive security and stability fixes without manual operator action
- Removes the support burden of nudging customers to upgrade after every release
- Visibility — backend knows what version each customer is running, enabling distribution dashboards
- Foundation for a maintenance / SLA story in enterprise sales conversations

## Improvement

After this lands, every running client either auto-upgrades or shows a clear "new version available" prompt within ~24h of a release. Backend can also report client-version distribution (how many customers on 1.0.8 vs 1.1.0).

## Open decisions before implementation

- **UX:** notify-and-approve vs auto-upgrade vs hybrid
- **Where the notification surfaces:** workspace UI in `frontend-app` (needs `backend` support too), email, or both
- **Cadence:** daily / hourly / Helm hook
- **Priority label**

## Status

Open

saadqbal:
## Summary

- Closes #69. Ships a `<release>-auto-upgrade` CronJob that polls `https://tracebloc.github.io/client` daily and runs `helm upgrade --reuse-values` when a newer chart version is published, so deployed clients no longer freeze on the version they first installed.
- Implements **option B (auto-upgrade)** from the issue — picked over notify-and-approve / hybrid because the issue's whole point is "customers don't run upgrade manually".
- Bumps chart **1.2.3 → 1.3.0**.

## What's in the chart now

| Resource | Name | Notes |
|---|---|---|
| `CronJob` | `<release>-auto-upgrade` | `concurrencyPolicy: Forbid`, default schedule `23 2 * * *` (UTC), `backoffLimit: 2` |
| `ConfigMap` | `<release>-auto-upgrade` | Holds the upgrade shell script |
| `ServiceAccount` | `<release>-auto-upgrade` | In the release namespace |
| `ClusterRoleBinding` | `<release>-auto-upgrade` | Bound to the built-in `cluster-admin` |

The Pod satisfies PSA `restricted`: `runAsNonRoot: true`, `runAsUser/Group: 1000`, `RuntimeDefault` seccomp, `allowPrivilegeEscalation: false`, `readOnlyRootFilesystem: true`, `capabilities.drop: [ALL]`. `HOME` and `HELM_*_HOME` are redirected into a `/tmp` emptyDir so helm can write its caches.

The version compare uses `sort -V`, not string compare — `1.10.0 > 1.9.0`.

## Trade-offs surfaced for review

1. **Default `enabled: true`.** The issue is "customers freeze on the version they installed". Default-off recreates the bug for everyone who installs 1.3.0 and never re-runs helm. Operator can opt out via `autoUpgrade.enabled: false`, or pause via `autoUpgrade.suspend: true`.
2. **`cluster-admin` not a narrow custom role.** The chart already templates cluster-scoped resources (`PriorityClass`, `StorageClass`, `ClusterRole`/`Binding`, optionally `Namespace`). A curated role would silently break the day a future chart version adds a new resource kind on already-deployed clients. Trust boundary is documented in `values.yaml` next to the `autoUpgrade:` block.
3. **Backend reporting deferred.** The issue mentions reporting back to the tracebloc backend so the workspace UI knows what version each customer runs. No endpoint exists yet — will land as a follow-up once the contract is defined. Not blocking the upgrade loop.
4. **Image pinned to `alpine/helm:3.16.4`.** No `jq` dep — script parses helm's YAML output with awk only. `latest` is rejected by the schema so behaviour can't drift from under us.

## Test plan

- [x] `helm lint client` clean (default + AKS + bare-metal value files)
- [x] `helm template stg client …` renders 4 new docs by default; renders 0 with `autoUpgrade.enabled=false`
- [x] `helm unittest client` — 12 suites, **116 tests** pass (11 of those new in `tests/auto_upgrade_test.yaml`)
- [ ] **Reviewer to verify on a live cluster** (out of scope for this PR but the natural next step):
  - [ ] `helm install` 1.3.0 fresh; observe `<release>-auto-upgrade` CronJob created
  - [ ] Manually trigger the Job (`kubectl create job … --from=cronjob/<release>-auto-upgrade`); confirm it logs `already at latest; nothing to do`
  - [ ] Publish a 1.3.1 to `gh-pages`; trigger the Job again; observe `helm upgrade complete: 1.3.0 -> 1.3.1`
  - [ ] Verify reused values: `helm get values <release>` still shows `clientId` / `clientPassword` (not stripped by `--reuse-values`)
  - [ ] Verify on a bare-metal cluster (no CSI) since the script's `helm upgrade --wait --timeout 10m` interacts with PVC bind timing

## Follow-ups (separate tickets)

- Backend reporting endpoint + auth wiring (chart side: optional `autoUpgrade.reportUrl`).
- Optional per-customer override on what major-version bumps trigger auto-apply vs notify (option C / hybrid). Not needed for v1.

---

> [!NOTE]
> **High Risk**
> Introduces a default-on CronJob that can mutate cluster-scoped resources and is bound to `cluster-admin`, so a compromised pod or chart repo could lead to full cluster takeover.
> 
> **Overview**
> Adds an **auto-upgrade mechanism** to the `client` Helm chart: when `autoUpgrade.enabled` (default **true**) it installs a `<release>-auto-upgrade` ConfigMap+CronJob that periodically checks the published chart repo and runs `helm upgrade --reset-then-reuse-values` to move the release to the latest chart version.
> 
> This also introduces the supporting `ServiceAccount` + `ClusterRoleBinding` (bound to built-in `cluster-admin`), new `autoUpgrade` values and schema validation (including rejecting `image.tag: latest`), updates install notes/migration docs, adds helm-unittest coverage, and bumps the chart/app version to `1.3.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b402aca81817f65998fad9e0a21bb24101fc8d6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.2.3
-appVersion: "1.2.3"
+version: 1.3.0
+appVersion: "1.3.0"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/MIGRATION.md
+++ b/client/MIGRATION.md
@@ -2,6 +2,31 @@
 
 This guide explains how to migrate from the legacy per-platform charts (`aks/`, `bm/`, `eks/`, `oc/`) to the unified `client/` chart.
 
+## Upgrading to 1.3.0 — self-upgrade CronJob lands on by default
+
+Releases of 1.3.0+ install a `<release>-auto-upgrade` CronJob that polls
+`https://tracebloc.github.io/client` daily and runs `helm upgrade --reuse-values`
+when a newer chart version is published. This closes
+[tracebloc/client#69](https://github.com/tracebloc/client/issues/69) — older
+deployed clients stop drifting from the latest secure / stable release.
+
+The upgrader's ServiceAccount is bound to the built-in `cluster-admin`
+ClusterRole because the chart already templates cluster-scoped resources
+(`PriorityClass`, `StorageClass`, `ClusterRole`/`Binding`, optionally
+`Namespace`); a curated narrower role would silently break the day a future
+chart version adds a new resource kind.
+
+To opt out and keep the manual approval gate you had on 1.2.x:
+
+```yaml
+# values-overrides.yaml
+autoUpgrade:
+  enabled: false
+```
+
+Or for a one-shot pause without removing the resources, set
+`autoUpgrade.suspend: true`.
+
 ## What Changed
 
 | Legacy | Unified |
--- a/client/MIGRATION.md
+++ b/client/MIGRATION.md
@@ -5,10 +5,24 @@ This guide explains how to migrate from the legacy per-platform charts (`aks/`, 
 ## Upgrading to 1.3.0 — self-upgrade CronJob lands on by default
 
 Releases of 1.3.0+ install a `<release>-auto-upgrade` CronJob that polls
-`https://tracebloc.github.io/client` daily and runs `helm upgrade --reuse-values`
-when a newer chart version is published. This closes
-[tracebloc/client#69](https://github.com/tracebloc/client/issues/69) — older
-deployed clients stop drifting from the latest secure / stable release.
+`https://tracebloc.github.io/client` daily and runs
+`helm upgrade --reset-then-reuse-values` when a newer chart version is
+published. This closes [tracebloc/client#69](https://github.com/tracebloc/client/issues/69) —
+older deployed clients stop drifting from the latest secure / stable release.
+
+> **Operator note for the 1.x → 1.3.0 jump.** Use `--reset-then-reuse-values`
+> on the *manual* upgrade command too, not plain `--reuse-values`. The new
+> `autoUpgrade` block was added in 1.3.0; with `--reuse-values` Helm reuses
+> the last release's *computed* values, which don't contain `autoUpgrade`,
+> and the new templates fail with `nil pointer evaluating interface {}.enabled`.
+> Once you're on 1.3.0+ the CronJob handles future bumps with the correct
+> flag itself.
+>
+> ```bash
+> helm upgrade <release> tracebloc/client \
+>   -n <namespace> --version 1.3.0 \
+>   --reset-then-reuse-values
+> ```
 
 The upgrader's ServiceAccount is bound to the built-in `cluster-admin`
 ClusterRole because the chart already templates cluster-scoped resources
--- a/client/templates/NOTES.txt
+++ b/client/templates/NOTES.txt
@@ -23,6 +23,11 @@
 {{- if .Values.openshift.scc.enabled }}
   {{ "\033[1;34m" }}OpenShift SCC:{{ "\033[0m" }}    {{ "\033[0;33m" }}tracebloc-resource-monitor-{{ .Release.Name }}{{ "\033[0m" }}
 {{- end }}
+{{- if .Values.autoUpgrade.enabled }}
+  {{ "\033[1;34m" }}Auto-upgrade:{{ "\033[0m" }}     {{ "\033[1;32m" }}ON{{ "\033[0m" }} {{ "\033[0;90m" }}(CronJob {{ .Release.Name }}-auto-upgrade, schedule "{{ .Values.autoUpgrade.schedule }}", repo {{ .Values.autoUpgrade.repoUrl }}){{ "\033[0m" }}
+{{- else }}
+  {{ "\033[1;34m" }}Auto-upgrade:{{ "\033[0m" }}     {{ "\033[1;33m" }}OFF{{ "\033[0m" }} {{ "\033[0;90m" }}(release pinned to chart {{ .Chart.Version }}; set autoUpgrade.enabled=true to receive future fixes automatically){{ "\033[0m" }}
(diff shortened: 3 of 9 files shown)


saadqbal:
## Dev-cluster verification — `tb-client-dev-templates` (EKS, eu-central-1)

Verified end-to-end against the existing `tracebloc` release in `tracebloc-templates` (was on `client-1.1.0`).

### One bug caught + fixed during verification

A dry-run of the manual 1.1.0 → 1.3.0 jump with plain `--reuse-values` failed:

```
template: client/templates/auto-upgrade-rbac.yaml:1:14:
executing "..." at <.Values.autoUpgrade.enabled>:
nil pointer evaluating interface {}.enabled
```

Root cause: `--reuse-values` reuses the previous release's *computed* values, so any new top-level key added in the upgraded chart (here, the whole `autoUpgrade` block) is missing from the merged values and templates that reference it blow up. Fix in commit b402aca:
- In-chart upgrade script switched to `--reset-then-reuse-values` (helm 3.14+, available in our pinned `alpine/helm:3.16.4`)
- `MIGRATION.md` tells operators to use the same flag for the *manual* 1.x → 1.3.0 jump
- Subsequent chart bumps go through the CronJob, which now uses the right flag itself

### Verification results (with the fix)

`helm upgrade tracebloc ./client -n tracebloc-templates --reset-then-reuse-values --wait`:
- ✅ Release moved 1.1.0 → 1.3.0, status `deployed` (revision 8)
- ✅ Pre-existing `mysql-client` and `tracebloc-jobs-manager` pods stayed Running, no restart (non-disruptive upgrade)
- ✅ PVCs unchanged (`client-pvc`, `client-logs-pvc`, `mysql-pvc` all bound)
- ✅ NOTES.txt prints the new `Auto-upgrade: ON …` line

Auto-upgrade resources:
```
cronjob.batch/tracebloc-auto-upgrade   23 2 * * *   <none>   False   0   <none>   2m26s
serviceaccount/tracebloc-auto-upgrade  0   2m38s
configmap/tracebloc-auto-upgrade       1   2m34s
clusterrolebinding/tracebloc-auto-upgrade   ClusterRole/cluster-admin   2m32s
```

Manually fired the Job (`kubectl create job --from=cronjob/tracebloc-auto-upgrade …`); Pod ran under PSA `restricted` (non-root, RO root, dropped caps, RuntimeDefault seccomp) and exited Succeeded with these logs:

```
[auto-upgrade] release=tracebloc namespace=tracebloc-templates repo=https://tracebloc.github.io/client
[auto-upgrade] current=1.3.0 latest=1.2.3
[auto-upgrade] deployed version is ahead of repo (current=1.3.0 > latest=1.2.3); skipping
```

That exercises every code path the CronJob will execute against the public repo (helm repo add/update, search-repo YAML parse, helm-list YAML parse, semver compare via `sort -V`); the only branch-specific line for the "newer found" case is the `helm upgrade --reset-then-reuse-values --version $LATEST` invocation, which we just exercised manually for the bootstrap upgrade itself.

Ready for review.

saqlainsyed007 (approved):

PR #215, opened 2026-06-05
Post-install self-verification of CLI usability (#738)

LukasWodka:
Closes #738

## Summary
After Step 5 installs the tracebloc CLI, the installer told the user *"open a new terminal so it's on your PATH"* without ever proving a fresh terminal would find it — the cli#61 failure mode (binary lands in `~/.local/bin`, which a brand-new shell does not have on PATH), undetected until a customer hits it. This adds a post-install **self-verification** that probes a fresh shell for `tracebloc` and either prints a **verified** verdict or the **exact shell-correct PATH fix** for the user's actual `$SHELL`. Stays **non-fatal** (the client is already connected by Step 5). Mirrored on Windows.

## Related
Closes #738 · Part of #736 (install-journey epic)

## Type of change
- [x] Feature
- [x] Bug fix

## Test plan

**Verified locally (macOS, bats 1.13.0):**
- `bash -n scripts/lib/install-cli.sh` — passes
- `bats scripts/tests/install-cli.bats` — **8/8 pass**, including:
  - fresh-shell success reports a *verified* verdict (asserts `tracebloc version` proof, asserts the old "open a new terminal" line is gone, asserts the dataset-push command is NOT duplicated here)
  - CLI-missing-from-fresh-shell prints an actionable, shell-correct PATH hint (asserts the exact `export PATH="...local/bin:$PATH"` line + `source ~/.zshrc` for zsh)
  - fish gets a fish-correct fix (`fish_add_path` + `~/.config/fish/config.fish`, not a POSIX `export`)
  - verification failure is non-fatal (status 0), incl. a stress test under the orchestrator's `set -e`
- `bats scripts/tests/summary.bats` — 15/15 pass (unchanged; confirms next-steps stay consistent and the verified command is not duplicated)
- Unit-checked the rc-routing helper across zsh / bash-linux / bash-darwin / fish / sh / dash — all match the spec routing.

**Pre-existing baseline failures (NOT introduced here):** `validate_config`, `_extract_yaml_value`, and two Docker-engine RHEL/Amazon-Linux tests fail identically on clean `origin/develop` (macOS env-specific); untouched by this PR.

**Needs CI:**
- **Pester** (`scripts/tests/install-k8s.Tests.ps1`) — pwsh is not installed on this machine, so the new `Test-TraceblocCli` Describe block (verified verdict / actionable hint / non-fatal when `RefreshPath` throws) and the updated `Install-TraceblocCli` cases were **not run locally**. Needs the Windows/Pester CI job to confirm.
- **shellcheck** — not installed locally; relies on CI lint.

## Deployment notes
None. Behavior change is limited to the install-time messaging of Step 5; no env vars, migrations, or rollout ordering.

## Checklist
- [x] Tests added / updated and passing locally (bats; Pester needs CI)
- [x] Docs updated if behavior or config changed (installer self-describes; no separate docs)
- [x] For security-sensitive paths: appropriate reviewer requested (assigned @saadqbal)

### DIFF
--- a/scripts/lib/install-cli.sh
+++ b/scripts/lib/install-cli.sh
@@ -100,11 +100,21 @@ _verify_tracebloc_cli() {
   # `|| true` so a hiccup in rc-resolution can't trip the orchestrator's set -e.
   local rc; rc="$(_cli_rc_for_shell || true)"
   local export_line; export_line="$(_cli_path_export_line || true)"
-  success "tracebloc CLI installed — one step to put it on your PATH in this terminal:"
-  hint "  ${export_line}"
-  hint "  source ${rc}"
-  info "After that, the 'tracebloc dataset push ./data' step below will work here."
-  info "(New terminals pick it up automatically once ${rc} has the line above.)"
+  local sh_name; sh_name="$(basename "${SHELL:-/bin/sh}")"
+  success "tracebloc CLI installed — put it on your PATH:"
+  if [[ "$sh_name" == "fish" ]]; then
+    # fish_add_path persists (a universal var) AND applies to this shell — no
+    # `source` needed, unlike a POSIX rc edit.
+    hint "  ${export_line}"
+  else
+    # Append the line to the rc, then load it: fixes THIS terminal and every
+    # new one in a single copy-pasteable step (the old code printed a bare
+    # `export` that fixed only this shell, then `source`d an rc that didn't
+    # yet contain the line — so nothing persisted).
+    hint "  echo '${export_line}' >> ${rc}"
+    hint "  source ${rc}"
+  fi
+  info "Then the 'tracebloc dataset push ./data' step below will work."
   return 0
 }
 
--- a/scripts/lib/install-cli.sh
+++ b/scripts/lib/install-cli.sh
@@ -22,6 +22,92 @@
 
 TRACEBLOC_CLI_INSTALL_URL="https://github.com/tracebloc/cli/releases/latest/download/install.sh"
 
+# Where the CLI's own installer drops the binary when /usr/local/bin isn't
+# writable (see cli's install.sh) — the dir we tell the user to put on PATH.
+TRACEBLOC_CLI_FALLBACK_BIN="${HOME}/.local/bin"
+
+# Which rc file a *fresh* interactive shell of the user's $SHELL actually reads,
+# so the PATH fix we print sources the right file. Mirrors how the cli's
+# install.sh routes guidance, but resolved per-shell here:
+#   zsh           → ~/.zshrc
+#   bash + Linux  → ~/.bashrc      (a fresh non-login bash reads ~/.bashrc,
+#                                    NOT ~/.profile — this is the failure mode)
+#   bash + macOS  → ~/.bash_profile
+#   fish          → ~/.config/fish/config.fish
+#   anything else → ~/.profile     (POSIX sh fallback)
+_cli_rc_for_shell() {
+  local sh_name; sh_name="$(basename "${SHELL:-/bin/sh}")"
+  case "$sh_name" in
+    zsh)  echo "${HOME}/.zshrc" ;;
+    bash)
+      if [[ "${OS:-$(uname -s)}" == "Darwin" ]]; then
+        echo "${HOME}/.bash_profile"
+      else
+        echo "${HOME}/.bashrc"
+      fi
+      ;;
+    fish) echo "${HOME}/.config/fish/config.fish" ;;
+    *)    echo "${HOME}/.profile" ;;
+  esac
+}
+
+# The shell-correct line a fish user must add differs (no POSIX `export`).
+_cli_path_export_line() {
+  local sh_name; sh_name="$(basename "${SHELL:-/bin/sh}")"
+  if [[ "$sh_name" == "fish" ]]; then
+    echo "fish_add_path \"${TRACEBLOC_CLI_FALLBACK_BIN}\""
+  else
+    echo "export PATH=\"${TRACEBLOC_CLI_FALLBACK_BIN}:\$PATH\""
+  fi
+}
+
+# Does a *fresh* shell resolve `tracebloc` on its PATH? This is the real test
+# the success message has, until now, only asserted by hope: a brand-new
+# terminal must find the binary. We probe two ways because they read different
+# startup files:
+#   1. login shell    ("$SHELL" -lic)  → ~/.profile / ~/.zprofile / ~/.bash_profile
+#   2. non-login shell ("$SHELL" -ic)  → ~/.bashrc / ~/.zshrc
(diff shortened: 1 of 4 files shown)


saadqbal (dismissed):

aptracebloc (dismissed):

scripts/lib/install-cli.sh
```diff
@@ -22,6 +22,92 @@
...
+# startup files:
+#   1. login shell    ("$SHELL" -lic)  → ~/.profile / ~/.zprofile / ~/.bash_profile
+#   2. non-login shell ("$SHELL" -ic)  → ~/.bashrc / ~/.zshrc
+# A pass requires BOTH (cli#61 was "works in my login shell, missing in a plain
+# `bash` subshell"). Indirected into its own function so the bats suite can stub
+# it without spawning real shells. Never fatal: returns non-zero on "not found".
+_cli_on_fresh_path() {
+  local shell_bin="${SHELL:-/bin/sh}"
+  "$shell_bin" -lic 'command -v tracebloc' >/dev/null 2>&1 || return 1
+  "$shell_bin" -ic  'command -v tracebloc' >/dev/null 2>&1 || return 1
```
saadqbal, line 76: **Fresh-shell probe: fish-specific false-negative risk.** *(updated — dash checks out fine)*

The probe runs `command -v tracebloc` via `"$SHELL" -lic`. I verified the non-fish shells: `zsh`/`bash`/`dash` all accept `-lic` and resolve `command -v` (the `dash: can't access tty` notice is harmless and gets swallowed by the probe's `2>/dev/null`). So this narrows to **fish**:

- fish's `command` builtin doesn't implement `-v` (it uses `-s`/`--search`, or `type -q`). So `command -v tracebloc` likely errors under fish → the **miss branch fires even when `tracebloc` is already on PATH** → a fish user is told to "fix" a PATH that's fine, via the `fish_add_path` line `_cli_path_export_line` carefully produced for them.

I couldn't confirm this here (no fish on the review machine), so flagging rather than asserting. Quick check on any box with fish — a non-zero/error confirms it:
```
fish -lic 'command -v ls'; echo "status=$status"
```
The failure direction is safe (non-fatal, guidance still fish-correct), so not a blocker — but it's the one path with **no test coverage** (`_cli_on_fresh_path` is stubbed everywhere). Fix options: per-shell query (`fish -lic 'type -q tracebloc'`), or gate the probe to bash/zsh/dash and fall through to the miss-branch guidance for fish — same safe outcome, intentional rather than incidental.

scripts/lib/install-cli.sh
```diff
@@ -22,6 +22,92 @@
...
+  # ~/.local/bin, which isn't on PATH). Tell the user precisely how to fix it
+  # for THEIR shell — not a generic "open a new terminal" that won't help.
+  # `|| true` so a hiccup in rc-resolution can't trip the orchestrator's set -e.
+  local rc; rc="$(_cli_rc_for_shell || true)"
+  local export_line; export_line="$(_cli_path_export_line || true)"
+  success "tracebloc CLI installed — one step to put it on your PATH in this terminal:"
+  hint "  ${export_line}"
+  hint "  source ${rc}"
+  info "After that, the 'tracebloc dataset push ./data' step below will work here."
+  info "(New terminals pick it up automatically once ${rc} has the line above.)"
```
saadqbal: **The printed PATH fix mixes "fix now" with "persist", and over-instructs fish.**

- `export PATH=…` fixes only the *current* shell; the following `source ${rc}` then re-runs an rc that does **not** contain the line — so neither command persists the fix, yet the closing line implies `${rc}` should contain it (the user is never told to write it there).
- For **fish**, `${export_line}` is `fish_add_path "…"`, which already persists (universal var) **and** applies to the running shell — the extra `source …/config.fish` is redundant.

Since #738 is about replacing vague guidance with *correct* guidance, suggest making POSIX shells append-then-source (correct for this terminal **and** new ones) and letting fish stand on its one self-persisting command:

```suggestion
  local sh_name; sh_name="$(basename "${SHELL:-/bin/sh}")"
  success "tracebloc CLI installed — put it on your PATH:"
  if [[ "$sh_name" == "fish" ]]; then
    # fish_add_path persists (universal var) AND applies to this shell — no source needed.
    hint "  ${export_line}"
  else
    # Append to the rc, then load it: fixes this terminal and every new one.
    hint "  echo '${export_line}' >> ${rc}"
    hint "  source ${rc}"
  fi
  info "Then the 'tracebloc dataset push ./data' step below will work."
```

`rc`/`export_line` are still computed just above, so they're in scope. Test impact: the **fish** bats test's `source …/config.fish` assertion must be dropped (the fix is that fish needs no source); the **zsh** test still passes (`export PATH="$HOME/.local/bin:$PATH"` remains a substring of the `echo '…' >>` line, and `source $HOME/.zshrc` is still printed).

saadqbal (commented):
Two non-blocking follow-ups from the review, both on `scripts/lib/install-cli.sh` — they bear on #738's core promise of *correct, shell-specific* guidance. CI is green and every failure mode here is non-fatal, so neither blocks merge.

scripts/lib/install-cli.sh
saadqbal: Addressed in 9ec5a5d — POSIX shells now print `echo '<export>' >> ~/.rc` then `source ~/.rc` (fixes this terminal *and* every new one in one step), and fish prints only `fish_add_path "…"` (no `source`, since it self-persists). Tests updated; bats 8/8, `shellcheck --severity=error` gate clean.

scripts/lib/install-cli.sh
saadqbal: Withdrawing this — I ran the actual probe under real fish 3.4.1 and 4.0.2 (Alpine containers): `fish -lic 'command -v ls'` returns `/bin/ls` and exits 0, and the bundled `-lic` flags parse fine. fish's `command` builtin *does* implement `-v`, so the probe resolves `tracebloc` correctly under fish — my assumption otherwise was wrong, and no change is needed. Resolving.

saadqbal (approved):
LGTM. ✅

Reviewed the post-install CLI self-verification (#738):

- **#2 — PATH-fix guidance** (addressed in `9ec5a5d`): POSIX shells now `echo '<export>' >> ~/.rc` + `source ~/.rc` (persists for this terminal and new ones); fish uses its self-persisting `fish_add_path` with no bogus `source`. `bats` 8/8, `shellcheck --severity=error` gate clean, `bash -n` clean.
- **#1 — fish fresh-shell probe**: verified against real fish **3.4.1** and **4.0.2** that `command -v` resolves correctly and `-lic` parses — non-issue, **withdrawn**.

Both review threads resolved, CI green. Design is sound: non-fatal after Step 5, two-shell (login + non-login) probe matching the cli#61 failure mode, and faithfully mirrored on Windows via `RefreshPath`. 👍

saadqbal (approved):

PR #218, opened 2026-06-06
test(drift): CI guard for source-of-truth drift (backend hosts + workload names)

LukasWodka:
## What

Net-new CI **drift check** — fails when two sources of truth that must agree have silently diverged. From the test-hardening analysis; the only roadmap item with no existing PR. Tracks backend#746.

Two checks (`scripts/tests/check-drift.sh`):

**1. Backend API host parity** — the dev/stg/prod hosts are hardcoded in three files (`preflight.sh::_pf_backend_host`, `install-client-helm.sh::_backend_url`, `install-k8s.ps1::Get-BackendUrl`). Asserts all three carry the identical `*.tracebloc.io` host set.

**2. Workload-name contract** — `summary.sh` (readiness wait, `rollout status deployment/<name>`) and `diagnose.sh` (`--diagnose` bundle, `logs deploy/<name>` + `daemonset/<name>`) reference `mysql-client`, `<release>-jobs-manager`, `<release>-requests-proxy`, `tracebloc-resource-monitor` by name. A chart rename breaks readiness + diagnostics **silently** — same class as the `app=manager` selector bug. Renders the chart (`helm template`) and asserts each name exists, and that the scripts still reference each (so the contract can't go stale on either side).

## How
- `.github/workflows/drift-checks.yaml` runs on `scripts/**` OR `client/**` (either side moving is the drift), sets up helm, runs the check.
- `scripts/tests/check-drift.bats` — 8 cases (parity match / mismatch / removed; contract drop; render-missing; helm-absent skip).
- `check-drift.sh` added to the static job's shellcheck list.

## Testing
- `bash scripts/tests/check-drift.sh` → green on develop (both checks pass; the chart render confirms the four workload names).
- `bats scripts/tests/check-drift.bats` → 8/8.
- Full suite: no new failures (same 4 pre-existing macOS-env failures).
- shellcheck (error severity) clean on the new script.

## Follow-ups (separate)
- Cross-repo selector drift: Mintlify `docs` `-l app=…` selectors ↔ chart labels (the original `app=manager` surface) — needs a cross-repo guard.
- Preflight thresholds ↔ docs sizing numbers (cross-repo).
- Promote the `Source-of-truth drift` job to a required status check once proven stable.

Refs tracebloc/backend#746

### DIFF
--- a/.github/workflows/drift-checks.yaml
+++ b/.github/workflows/drift-checks.yaml
@@ -1,0 +1,37 @@
+name: Drift checks
+
+# Source-of-truth drift: values duplicated across files (the backend API hosts)
+# or coupled across artifacts (the chart's workload names <-> the names the
+# readiness-wait and --diagnose bundle grep for). Mocked unit tests can't see
+# these — a rename ships green and breaks in the field. This guards both sides.
+# Triggers on scripts/ OR client/ because either side moving is the drift.
+on:
+  push:
+    branches: [main, develop, openshift]
+    paths:
+      - 'scripts/**'
+      - 'client/**'
+      - '.github/workflows/drift-checks.yaml'
+  pull_request:
+    branches: [main, develop, openshift]
+    paths:
+      - 'scripts/**'
+      - 'client/**'
+      - '.github/workflows/drift-checks.yaml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  drift:
+    name: Source-of-truth drift
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


saadqbal (approved):
Solid net-new drift guard — CI-only, can't affect runtime. Check 1 (backend host parity across preflight.sh / install-client-helm.sh / install-k8s.ps1) and Check 2 (workload-name contract: 2a scripts ↔ 2b `helm template` render) both look correct, and sourcing is side-effect-safe (set + main gated on direct execution) so the bats suite can exercise the helpers. 8/8 bats green. LGTM.

PR #279, opened 2026-06-24
feat(installer): reorder — sign in + provision the client before Helm (#838)

LukasWodka:
## What

RFC-0001 **#838** ([epic backend#830](https://github.com/tracebloc/backend/issues/830)): reorder the installer so the client is **provisioned before Helm** — the minted credential + derived namespace feed the chart.

New flow in `install-k8s.sh`:
```
create_cluster → provision_client (Step 3) → install_client_helm (Step 4/5)
```

## `provision.sh` (new) — `provision_client`

- **Browser-auth path:** install the CLI **(now FATAL** — it mints the credential, vs the old non-fatal post-Helm Step 5) → `tracebloc login` (device flow) → `tracebloc client create --credential-file <f>` → **source the 0600 file** → export `TRACEBLOC_CLIENT_ID` + `TRACEBLOC_CLIENT_PASSWORD` + `TB_NAMESPACE` into `install_client_helm`'s **existing** non-interactive env contract. The credential is **never printed**; the transient file is removed after sourcing **and on any error/signal** (a failed mint never leaves a secret-bearing file behind).
- **Adopt (re-run):** no fresh credential (the existing one stands) — hands only `TB_NAMESPACE` and lets `install_client_helm` reconcile from the local `values.yaml`.
- **Dual-mode preserved:** `TRACEBLOC_VALUES_FILE` or pre-supplied `TRACEBLOC_CLIENT_ID`+`PASSWORD` **skip sign-in** entirely (still install the CLI, non-fatal) — exactly the prior behavior. Unattended installs use this path.

Also: `install.sh` fetches `provision.sh`; `common.sh` `print_roadmap` renumbered to match (3 sign-in/provision · 4 install · 5 connect) and `install_cleanup` shreds the transient credential on exit; the step headers + the trailing CLI install were updated/removed.

## Tests

`scripts/tests/provision.bats` (8): dual-mode skip (creds + values), mint hands all three env vars, adopt hands only the namespace, **fatal** on missing CLI / failed login / no credential file, and **a failed `create` leaves no credential file behind**. The `install-cli`, `install-client-helm`, and `common` (`install_cleanup`) suites stay green; `shellcheck` clean on `provision.sh`. (The 2 suite failures — `validate_config: valid config passes` and `_extract_yaml '' escape` — are **pre-existing on develop**, unrelated to this change.)

## Gate + deferred

- **Runs end-to-end once the CLI release ships `login` + `create` + `--credential-file`** (cli#104 merged; `install-cli.sh` fetches `releases/latest`).
- Adversarially reviewed (reorder/dual-mode safety, credential-handoff security, bats+integration); the should-fix items (cred-file cleanup on error, stale roadmap) are folded in.
- **Deferred** (noted, follow-ups): a non-interactive **enroll-token login** for fully unattended browser-auth installs; the **R7 orphan-resume** edge (adopt on a rebuilt host with no local `values.yaml` → it currently falls through to the existing credential prompt).

### DIFF
--- a/scripts/lib/provision.sh
+++ b/scripts/lib/provision.sh
@@ -56,12 +56,20 @@ provision_client() {
   # between mint and the explicit removal below — the secret must never linger.
   _PROVISION_CRED_FILE="$cred_file"
   rm -f "$cred_file"
-  if ! tracebloc client create --yes --credential-file "$cred_file" >>"${LOG_FILE:-/dev/null}" 2>&1; then
+  # umask 077 so the credential file lands 0600 even if a future CLI build
+  # regresses on its explicit chmod — defense in depth (cli#104 already sets 0600).
+  if ! ( umask 077; tracebloc client create --yes --credential-file "$cred_file" ) >>"${LOG_FILE:-/dev/null}" 2>&1; then
     rm -f "$cred_file"   # remove any partial the failed create may have written
     error "Provisioning the client failed — see ${LOG_FILE:-the install log}. Re-run to retry."
   fi
   [[ -f "$cred_file" ]] || error "client create did not write the credential file ($cred_file)."
 
+  # The credential file is the sole source of truth for what was provisioned.
+  # Clear any pre-existing values from the environment first: the mint case does
+  # NOT write TRACEBLOC_CLIENT_ADOPTED, so a stale ADOPTED=1 left in the env would
+  # otherwise misroute a fresh mint into the adopt branch and drop the just-minted
+  # credential. Same reasoning for id/password/namespace — only the file wins.
+  unset TRACEBLOC_CLIENT_ID TRACEBLOC_CLIENT_PASSWORD TB_NAMESPACE TRACEBLOC_CLIENT_ADOPTED
   # shellcheck disable=SC1090
   source "$cred_file"
   rm -f "$cred_file"
--- a/scripts/lib/provision.sh
+++ b/scripts/lib/provision.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+# =============================================================================
+#  provision.sh — RFC-0001 #838: provision the client BEFORE Helm.
+#
+#  In the browser-auth path this installs the tracebloc CLI (FATAL — it mints the
+#  credential), signs in (`tracebloc login`, the device flow), and `tracebloc
+#  client create`s to mint the machine credential + derive the namespace. The
+#  credential is written to a 0600 file by `--credential-file` and sourced here
+#  (never printed), then handed to the Helm step via the SAME env contract
+#  install_client_helm already consumes:
+#      TRACEBLOC_CLIENT_ID + TRACEBLOC_CLIENT_PASSWORD + TB_NAMESPACE
+#
+#  DUAL-MODE (unchanged, one deprecation cycle): when the operator pre-supplies a
+#  values file (TRACEBLOC_VALUES_FILE) or credentials (TRACEBLOC_CLIENT_ID +
+#  TRACEBLOC_CLIENT_PASSWORD), browser sign-in is skipped and those paths feed
+#  Helm as before. Unattended installs use dual-mode — the device-flow `login` is
+#  interactive; a non-interactive enroll-token login is a CLI follow-up.
+# =============================================================================
+
+# _provisioning_preset: true when the operator already supplied a values file or
+# credentials, so browser sign-in is skipped and install_client_helm uses them.
+_provisioning_preset() {
+  [[ -n "${TRACEBLOC_VALUES_FILE:-}" ]] && return 0
+  [[ -n "${TRACEBLOC_CLIENT_ID:-}" && -n "${TRACEBLOC_CLIENT_PASSWORD:-}" ]] && return 0
+  return 1
+}
+
+provision_client() {
+  step 3 5 "Sign in and provision this client"
+
+  if _provisioning_preset; then
+    info "Using the credentials you supplied — skipping browser sign-in."
+    # Still install the CLI (non-fatal) so the operator has it for `data ingest`.
+    install_tracebloc_cli
+    return 0
+  fi
+
+  # Browser-auth path. The CLI is REQUIRED here (it mints the credential), so the
+  # install is effectively FATAL — unlike the old post-Helm, non-fatal Step 5.
+  install_tracebloc_cli
+  # The CLI installer may drop the binary in ~/.local/bin (not yet on this
+  # process's PATH); make it resolvable so the login/create calls below work.
+  export PATH="${HOME}/.local/bin:${PATH}"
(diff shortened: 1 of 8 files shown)


saadqbal (commented):
## Review summary

Read the diff + surrounding code, ran the suite and shellcheck locally (isolated worktree), and verified the cross-repo CLI contract. **Careful, well-tested PR with honest disclosure of its gates.** All 8 `provision.bats` pass, `install-cli`/`common` suites green, `provision.sh` shellcheck-clean, and the 2 suite failures it calls out are genuinely pre-existing (the diff doesn't touch `validate_config` / `_extract_yaml_value`).

**Verified good**
- Env handoff matches `install_client_helm`'s existing `_noninteractive_creds` path (ID + PASSWORD → `=1`; adopt unsets both → values.yaml path).
- CLI contract matches exactly on cli `develop`: mint→ID+PASSWORD+TB_NAMESPACE (stdout suppressed), adopt→ID+TB_NAMESPACE+`ADOPTED=1` (no password), file 0600, parent dir auto-created.
- No ordering bug: `setup_log_file()` mkdir's `HOST_DATA_DIR` early in `main()`, before `provision_client` writes the cred file there.
- `install_tracebloc_cli` always returns 0 → the non-fatal claim holds under `set -e`; the browser path's `has tracebloc || error` is the real fatal gate.
- Credential cleanup: explicit `rm` on success + error paths, `unset` after, EXIT-trap backstop. "failed create leaves no file" test confirms it.
- Step renumber (1,2 → 3 → 4,5) consistent; `install_tracebloc_cli` step-framing correctly relocated to the caller, no orphaned caller.

Two items flagged inline (release sequencing for `main`; legacy-machine namespace divergence).

**Minor / nits**
- `verify_credentials` on a *freshly-minted* credential is fatal on `invalid`(400)/`inactive`(401) — worth confirming create→auth is read-your-writes consistent on the backend.
- `trap install_cleanup EXIT` is EXIT-only; "removed on any error/signal" is best-effort (SIGINT/SIGTERM in the brief mint→`rm` window is bash-version-dependent). `EXIT INT TERM` would make the claim true. Low severity.
- Defense-in-depth: `( umask 077; tracebloc client create … )` would keep the file 0600 even against a future CLI regression. Optional — cli#104 guarantees 0600 today.
- Test gap: the new `_PROVISION_CRED_FILE` removal inside `install_cleanup` (trap backstop) isn't directly tested; `provision.bats` covers the explicit `rm` paths only.
- Adopt comment says it "reconciles from the local values.yaml," but that path actually re-prompts interactively ("Use previous settings as defaults?" + Client ID/password with values.yaml defaults). Functional, but the comment oversells it as silent.

Net: good for `develop` once the namespace question is settled, with the CLI-release gate as a hard precondition for `main`.

_Reviewed with Claude Code._

scripts/lib/provision.sh
```diff
@@ -0,0 +1,87 @@
...
+  install_tracebloc_cli
+  # The CLI installer may drop the binary in ~/.local/bin (not yet on this
+  # process's PATH); make it resolvable so the login/create calls below work.
+  export PATH="${HOME}/.local/bin:${PATH}"
+  has tracebloc || error "The tracebloc CLI is required to provision this client but isn't available after install. Install it (curl -fsSL ${TRACEBLOC_CLI_INSTALL_URL} | sh), open a new shell, and re-run."
+
+  # Sign in (device flow — approve in a browser). Unattended installs use the
+  # dual-mode credential path above.
+  info "Sign in to tracebloc — approve this machine in your browser when prompted…"
+  tracebloc login || error "Sign-in didn't complete — re-run the installer to try again."
```
saadqbal, line 74: ⚠️ **Release-sequencing gate.** This makes `tracebloc login` fatal on the default interactive path, but the latest *published* CLI release is **v0.3.1 (2026-06-08)** — it predates cli#104 and has neither `login` nor `client create --credential-file`. Since `install-cli.sh` installs `releases/latest`, the browser-auth path can't work until a CLI release shipping both lands.

Fine on `develop` (not the install endpoint), but **don't promote to `main`** (the `curl|bash` endpoint) until that CLI release is published, or the default interactive install breaks right here. Verified the commands do exist on cli `develop` (`root.go:85` `newLoginCmd()`; `client.go:63-65` `--yes` / `--credential-file`), so it's purely a publish-ordering gate, not a contract mismatch.

scripts/lib/provision.sh
```diff
@@ -0,0 +1,87 @@
...
+    error "Provisioning the client failed — see ${LOG_FILE:-the install log}. Re-run to retry."
+  fi
+  [[ -f "$cred_file" ]] || error "client create did not write the credential file ($cred_file)."
+
+  # shellcheck disable=SC1090
+  source "$cred_file"
+  rm -f "$cred_file"
+  unset _PROVISION_CRED_FILE
+
+  if [[ "${TRACEBLOC_CLIENT_ADOPTED:-}" == "1" ]]; then
```
saadqbal, line 103: ⚠️ **Legacy re-run can create a duplicate release.** This changes the namespace *source* from the old hardcoded `tracebloc` default to the minted/adopted client slug. On a machine previously installed with the *old* installer (release in namespace `tracebloc`), an adopt that returns a slug namespace ≠ `tracebloc` makes `install_client_helm` run `helm upgrade --install <slug> -n <slug>` → a **second** release beside the orphaned `tracebloc` one (duplicate client, double capacity). The one-client-per-machine guard keys on `clientId`, not namespace, so it won't catch a same-client / different-namespace duplicate.

The guard already computes `existing_ns` — reusing it to reconcile in place on adopt would close this. Worth confirming what namespace the backend assigns legacy clients on adopt. Lower-frequency since fleets normally upgrade via the in-cluster auto-upgrade cronjob rather than a re-run — but the failure is silent.

scripts/lib/provision.sh
saadqbal: Addressed in 40ad207. `provision.sh` now probes the installed CLI for `login` + `client create` (`--help`, side-effect-free; cobra exits non-zero on an unknown command) before taking the browser-auth path. If the CLI is too old — i.e. the latest release still lags this installer — it falls back to the existing manual-credential prompt in `install_client_helm` instead of hard-failing on an unknown `tracebloc login`, and auto-engages the browser flow once the cli#104 release ships. So the reorder no longer hard-depends on CLI-release ordering; the `main`-promotion note remains a nicety, not a correctness gate. Regression test: `a CLI too old to provision falls back to manual sign-in (not fatal)`.

scripts/lib/provision.sh
saadqbal: Addressed in 40ad207. The one-client guard now detects the **same** client already installed under a **different** namespace (the legacy fixed-`tracebloc` ns vs. the #838 minted slug) and upgrades that release **in place** — reusing the guard's already-computed `existing_ns` — instead of forking a second release under the new namespace. Fresh installs still use the minted slug; a different-client install still hits the existing "Refusing to replace" guard. So no silent duplicate / capacity double-book can occur. Regression test: `same client in a different namespace -> upgrades in place, no duplicate`.

saadqbal (approved):

PR #326, opened 2026-07-09
fix(installer): survive curl|bash — retry name prompt (customer-reported), read creds from terminal, guard pkg-index refresh

LukasWodka:
## Summary

Installer robustness for the `curl … | bash` path — three fixes that stop opaque `set -e` aborts, including the **customer-reported onboarding failure of 2026-07-09** (Moritz, external).

### 1. `fix(installer)` — retry the client-name prompt so tty type-ahead can't abort provisioning ⭐ customer-reported
`provision_client` read the client name with `IFS= read -r client_name </dev/tty || true` — one shot, status swallowed by `|| true`, no retry, no fallback (location silently defaults to the detected zone, which is why only the *name* failed hard). Any empty/failed read → the fatal *"A name for this client is required to provision it."* The customer saw both prompts print, neither captured, install aborted.

Most likely trigger: **tty type-ahead** — during the ~minute browser-approval wait the CLI reads nothing, so a stray newline queued in the terminal is consumed by the read as an empty name. (An adversarial verification pass ruled out the background-process-group and "login drains the tty" theories: the reader is foreground and `tracebloc login` never touches the tty — the empty read is an environmental dead/queued-input condition, not our process state.)

Fix: read the name in a bounded retry loop that **re-prompts on an empty line** (a queued blank is skipped, not accepted) and **breaks on a failed read** (`rc≠0` = EOF / no live input, which re-prompting can't fix) so the actionable *set `TRACEBLOC_CLIENT_NAME`* error still fires for genuinely non-interactive runs. Reads route through `TB_TTY` (same seam as the credential reads below). +2 regression tests.

### 2. `fix(installer)` — read credential prompts from the terminal (`TB_TTY`)
The Step-5 credential prompt read stdin, which under `curl … | bash` is the piped script → EOF → `set -e` abort. Now reads from `TB_TTY` (=`/dev/tty`), with an actionable error when there's no terminal *and* no env creds.

### 3. `fix(installer)` — don't abort on a flaky package-index refresh
`spin_cmd "…" $PM_UPDATE` was unguarded under `set -e` (a flaky mirror aborted the whole install, though the per-package installs below are already guarded). Now `|| warn`.

Regenerated `scripts/manifest.sha256` (R8 supply-chain) for all three.

## Test plan
- `bats scripts/tests/provision.bats` — 18/18 green (incl. the 2 new type-ahead / dead-input tests).
- `bats scripts/tests/install-client-helm.bats` + `setup-linux.bats` — green for the touched paths.
- Two full-suite failures (`_extract_yaml_value: single-quoted with '' escape`, `validate_config: valid config passes`) are **pre-existing on `develop`** and macOS-bash-3.2-specific (untouched code) — verified by stashing my changes; Linux CI (`standard-checks.yml` runs bats + shellcheck) is authoritative.

## Immediate customer unblock (independent of merge)
```bash
curl -fsSL https://raw.githubusercontent.com/tracebloc/client/main/scripts/install.sh \
  | TRACEBLOC_CLIENT_NAME="moritz-macbook" TRACEBLOC_CLIENT_LOCATION=DE bash
```
Pre-setting the name skips the prompt entirely (the read only runs `if [[ -z "$client_name" ]]`).

---

> [!NOTE]
> **Low Risk**
> Changes are confined to installer shell UX and error handling (no auth backend or cluster logic); behavior for env-supplied credentials and unattended installs is preserved, with new regression tests.
> 
> **Overview**
> Hardens the bash installer for **`curl … | bash`** and flaky Linux package mirrors so failures surface as clear guidance instead of opaque **`set -e`** exits.
> 
> **Provisioning (`provision.sh`)** routes interactive reads through **`TB_TTY`** (default **`/dev/tty`**) and **retries the client-name prompt** on empty lines (fixes customer-reported type-ahead after browser approval) while **stopping on EOF** so non-interactive runs still hit the **`TRACEBLOC_CLIENT_NAME`** error.
> 
> **Helm credential step (`install-client-helm.sh`)** adds the same **`TB_TTY`** pattern, **`_tty_available` / `_no_interactive_creds_die`**, and per-read guards so missing or dead terminals fail with **`TRACEBLOC_CLIENT_ID` / `TRACEBLOC_CLIENT_PASSWORD`** instructions instead of mid-**`read`** aborts; “use previous settings” only runs when a TTY exists.
> 
> **Linux deps (`setup-linux.sh`)** wraps **`$PM_UPDATE`** with **`|| warn`** so a failed index refresh does not kill the whole install while per-package installs remain best-effort.
> 
> **`scripts/manifest.sha256`** is updated for the touched libs; **bats** cover no-TTY, EOF-on-readable-TTY, and name type-ahead regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 508a3782657decb68f3bfe9a7dc6cf88220b73ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/install-client-helm.sh
```diff
@@ -408,20 +432,20 @@ install_client_helm() {
   local _cred_attempt=0 _cred_max=5 _cred_status
   while true; do
     if [[ -n "$default_client_id" ]]; then
-      read -r -p "  Client ID [${default_client_id}]: " TB_CLIENT_ID_INPUT
+      read -r -p "  Client ID [${default_client_id}]: " TB_CLIENT_ID_INPUT <"$TB_TTY"
       TB_CLIENT_ID="${TB_CLIENT_ID_INPUT:-$default_client_id}"
     else
-      read -r -p "  Client ID: " TB_CLIENT_ID
+      read -r -p "  Client ID: " TB_CLIENT_ID <"$TB_TTY"
```
saadqbal: One gap: these credential reads (438/435/444/448, and the "Use previous settings" read at 358) go through `$TB_TTY` but have no EOF guard, unlike the name read in provision.sh (`|| _name_read_ok=0`). `_tty_available` only checks `-r`, so on a readable-but-dead-input tty (non-PTY ssh, some IDE terminals, a queued/drained tty — the same class this PR documents for provision.sh) it returns true, we enter the loop, and the first standalone `read … <"$TB_TTY"` hits EOF and aborts under `set -e` — the exact opaque failure this PR is killing for the name prompt, just left in place for creds.

Either guard each read the way provision.sh does, or have `_tty_available` reject a tty that yields no input. Happy path is fine; only bites the dead-input case.

LukasWodka: Good catch — fixed in 6770524. Factored the guidance into `_no_interactive_creds_die` and added `|| _no_interactive_creds_die` to all five `$TB_TTY` reads (358/435/438/444/448), so a readable-but-dead-input tty now fails fast with the env-var guidance instead of aborting mid-read under `set -e` — matching the `provision.sh` name-read behavior. Added a regression test (readable `/dev/stdin` backed by `/dev/null` → EOF → actionable error, no helm). Happy path unchanged.

saadqbal (approved):

PR #333, opened 2026-07-10
fix(installer): run the helm client install behind a spinner (no more silent ~10-15s)

LukasWodka:
### What
`Step 4/5 · Installing tracebloc client` ran `helm upgrade --install` (and the in-place reconcile) **blocking, with all output redirected to the log** — so after the step header the terminal sat frozen for ~10-15s (helm render + apply + image pull) with no feedback.

### Fix
Wrap both helm calls in the existing `spin_cmd` helper (`scripts/lib/common.sh`) — the same one the Docker/Homebrew installs already use:
- animated spinner + "Installing the tracebloc client…" / "Reconciling the existing client…"
- output still streamed to `$LOG_FILE`
- on failure: the log tail to stderr + the same `error` exit ("Check the log for details")

Honours RFC-0002 §2 *progress on every wait* (the principle already applied to the CLI's post-submit waits).

### Test
- `bash -n` + `shellcheck -S warning` clean on `install-client-helm.sh` and `common.sh`
- `install-client-helm.bats` green — 33 install tests, incl. every reconcile / adopt path
- Failure path verified: exits non-zero, full helm output preserved in `$LOG_FILE`

---

> [!NOTE]
> **Low Risk**
> Installer UX only; helm arguments and failure handling stay equivalent, with no auth or cluster logic changes.
> 
> **Overview**
> **Helm client install and adopt reconcile** no longer leave the terminal silent for ~10–15s. Both `helm upgrade` paths in `install-client-helm.sh` now use the shared **`spin_cmd`** helper (same as Docker/Homebrew steps) with messages *Installing the tracebloc client…* and *Reconciling the existing client…*.
> 
> The previous **mktemp + full output redirect** pattern is removed; **`spin_cmd`** still streams helm output to **`$LOG_FILE`**, prints a log tail on failure, and exits with the same user-facing error. **`scripts/manifest.sha256`** is updated for the changed installer script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ccf2538fb47a2406924d20c04bbbd8aef45ae569. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (approved):
@LukasWodka — **Approved.** Small, net-positive change: it swaps the bespoke `mktemp`/`cat`/`rm` capture at two call sites for `spin_cmd`, the installer's dominant pattern (18 uses in `setup-linux.sh`). The hand-rolled capture was the odd one out, so this makes the helm install consistent with every other long step and puts progress on the ~10-15s window that was a frozen terminal before.

What I verified:
- **Trust-root hash is correct** — recomputed `sha256(scripts/lib/install-client-helm.sh)` on the branch = `825c4507…`, matches the updated `manifest.sha256` byte-for-byte. Since this is an R8 signed-install path, a wrong hash would break cosign verification; it's right.
- **No correctness regression** — `spin` preserves helm's exit code via `wait "$pid"`; both call sites keep the `if ! …; then error …` shape; helm reads no stdin so backgrounding is safe; the `if !` context correctly suspends `set -e` inside the spinner. `setup_log_file` sets `LOG_FILE` (and `exec > >(tee …)`) before these run, so `spin_cmd`'s output lands in the same file `error` points the user at.
- **Behavioral delta is benign** — on failure the terminal now shows the last 10 log lines instead of the full dump; the full output is still in `LOG_FILE`, and with the global `tee` it's actually less duplicated than before.

Nits — none blocking, all pre-existing `spin_cmd` properties this PR merely inherits:
- ⚪ No `[[ -t 1 ]]` TTY guard and no `trap … INT TERM` to restore the cursor (`tput cnorm`) or kill the backgrounded child if the user Ctrl-Cs mid-spin. This PR puts `spin_cmd` on the single longest blocking step — the most likely place for an interrupt — so it's the highest-value reason to harden `spin_cmd` if that's ever done.
- ⚪ A helm error whose salient line is >10 lines from the end shows only in the log, not on screen. Acceptable given the log pointer.

Touches Asad's code-owned `scripts/lib/` + `manifest.sha256`, so his review is the intended backstop — but the gate isn't enforced on `develop`, so you're clear to merge on this approval.

PR #339, opened 2026-07-12
feat(installer): stop-and-check gate — recognize an existing install

LukasWodka:
## Summary

Adds a read-only **"stop-and-check" gate** to the installer so **re-running it on an already-set-up machine no longer drags the user through full provisioning**. On a verifiably healthy machine it prints one line and hands off to `tracebloc` (the home screen), then exits 0. Otherwise it falls through to today's 5-step flow, unchanged.

This is **PR 2 of 2** (PR 1 = the status-aware `tracebloc` home screen, tracebloc/cli#244). Together they fix the *"I just wanted to use the CLI / reinstall, and suddenly a whole new environment gets provisioned"* problem.

## Behavior

The gate runs after the banner, before the 5-step roadmap. `assess_existing_install()` classifies the machine (read-only, bounded, never fatal):

- **healthy** — cluster running **AND** a tracebloc release present **AND all core workloads Ready** (`mysql-client` + `jobs-manager` + `requests-proxy`) **AND** the CLI present → prints `✓ Already set up on this machine — no need to run the installer again.`, then runs `tracebloc` (the home screen) and `exit 0`. Never re-provisions, never re-prompts login.
- **degraded** — cluster stopped, a workload not Ready, or the CLI missing → prints an honest one-liner naming what's off ("your secure environment is stopped / still starting up / the CLI isn't installed"), then falls through to today's flow to reconcile.
- **fresh** — no cluster, or no tracebloc release → today's 5-step install, unchanged.

`--force` / `--reinstall` (or `TRACEBLOC_FORCE_REINSTALL=1`) bypasses the gate entirely. Non-interactive (`curl | bash`): healthy still short-circuits (the hand-off is output-only, reads no stdin); degraded/fresh run the normal flow.

## Safety (this is a public, cosign-verified installer)

- **`assess` is strictly read-only** — `k3d cluster list`, `kubectl get`, `helm list` (via the reused `detect_installed_client`), `has`. It never starts the cluster, runs `helm upgrade`, mints credentials, or writes anything, and it does **not** call the mutating `_handle_existing_cluster`.
- **No false "healthy."** Every probe degrades to non-healthy on any error / timeout / missing tool / uncertainty — a machine that needs setup is never told "already set up." (Mutation-proven.)
- **Manifest integrity.** `scripts/lib/assess.sh` is added to both bootstrap `FILES` arrays and `manifest.sha256` is regenerated (`gen-manifest.sh --check` → up to date), so the cosign bootstrap still verifies; `install-bootstrap.bats` file-count lists updated (15 → 16).
- The healthy path uses `exit 0` (not `exec`) so the `install_cleanup` EXIT trap still runs — a verified no-op at gate time (before any credential / sudo-keepalive is set).
- **`provision.sh` untouched**, to stay clear of the active #838 provision rewrite.

## Single source of truth for "ready"

Extracted the client-workload deployment list into `_client_workload_deployments()` (`common.sh`); both `wait_for_client_ready` (the post-install gate) and the assess readiness probe consume it, so the installer's definition of "ready" and the gate's **can't drift**.

## Code review

Adversarial review before opening — both invariants (no false-healthy; non-mutating + manifest-consistent) confirmed and mutation-proven. Fixed in the second commit:
- **[should-fix]** the readiness gate checked only `jobs-manager`; a machine with jobs-manager up but `requests-proxy` (training egress) or `mysql-client` down could be called "healthy" → now requires **all three**, via the shared helper.
- **[should-fix]** an unguarded awk-pipe assignment that could abort the installer under `set -e`/pipefail (SIGPIPE) → guarded to match its jq sibling.

## Deferred (fast-follow)

Surgical per-layer repair of a degraded machine (start-only / CLI-only / reconcile-only, and skipping re-login when already authenticated) — deliberately left out to keep this PR off `provision.sh` and clear of #838. Rebases cleanly once #838 lands.

## Test plan

- `scripts/tests/assess.bats` (31 tests): each probe (cluster-running jq + awk fallback; workload-ready per-deployment incl. `requests-proxy` / `mysql-client` down; CLI present); all classify branches; the hand-off (runs `tracebloc`, `exit 0`, fallback); and the `main()` branch (healthy → hand-off + exit 0; `--force` bypass; degraded/fresh → flow). Assertions are fail-loud (grep-backed + single-bracket status) to dodge the macOS bats `[[ ]]` blindspot; the healthy short-circuit and the all-three-workloads gate are **mutation-verified**.
- `shellcheck --severity=error` clean; `bash -n` clean; `gen-manifest.sh --check` up to date.
- Full suite: only 2 **pre-existing** macOS-local failures (reproduce on pristine `develop`, unrelated files) — **Linux CI is authoritative** for bats.

## Checklist

- [x] Targets `develop`
- [x] Read-only assess; no false-healthy (mutation-proven)
- [x] Manifest regenerated + bootstrap `FILES` in sync
- [x] `provision.sh` untouched (#838 coordination)
- [x] Tests added + mutation-checked
- [ ] Human review (Bugbot + reviewer on Monday)

---

Pairs with **tracebloc/cli#244**. Hand-off degrades gracefully: shows today's home screen on the released CLI, the new status-aware one once cli#244 ships.

---

> [!NOTE]
> **Medium Risk**
> Changes behavior of the cosign-verified public installer entrypoint; a false healthy would skip needed setup, though the design biases toward fall-through on any doubt and adds extensive bats coverage.
> 
> **Overview**
> Adds a **read-only stop-and-check gate** so re-running the installer on an already-healthy machine skips full provisioning and hands off to `tracebloc` instead.
> 
> New **`scripts/lib/assess.sh`** classifies the host as **fresh**, **degraded**, or **healthy** using bounded `k3d`/`kubectl`/`helm` probes (never mutating). Only **healthy** short-circuits: running cluster, tracebloc release, all core workloads ready, and CLI present. Uncertainty always falls through to the normal install flow—never a false “already set up.”
> 
> **`install-k8s.sh`** invokes the gate after the banner; **`--force` / `--reinstall`** and **`TRACEBLOC_FORCE_REINSTALL=1`** bypass it. **`assess.sh`** is guarded when missing (stale bootstrap) like other optional libs.
> 
> **`_client_workload_deployments()`** in **`common.sh`** centralizes readiness (`mysql-client`, jobs-manager, requests-proxy); **`wait_for_client_ready`** and the assess gate both use it so “ready” cannot drift.
> 
> Bootstrap **`FILES`**, **`gen-manifest.sh`**, **`manifest.sha256`**, help text, and **`assess.bats`** / bootstrap tests updated for the new module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18b113a42df3580a0ddb7c53111f3e4fb6d6f5c0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (approved):
@LukasWodka — **Approved.** The core invariants hold: the gate is read-only, every probe fails safe toward the normal flow, the `set -euo pipefail` patterns are correct, and the `exit 0` hand-off is safe (detail in my comment above).

Caveats — follow-up, none blocking:
- 🟡 the healthy hand-off runs bare `tracebloc` on the inherited curl-pipe stdin; `|| true` guards a non-zero exit but not a blocking read — `tracebloc </dev/null || true` is a cheap hardening if you're already in the file, otherwise fine as a follow-up.
- 🟡 the `--force`/`--reinstall` → `TB_FORCE_REINSTALL` arg loop is untested (bats presets the var directly).
- ⚪ "release present" could in principle match a non-tracebloc `client-*` helm release; header "writes nothing" is slightly stronger than literal (`detect_installed_client` writes a temp file).

This sits squarely in the R8 trust root (`install.sh`, `install-k8s.sh`, `gen-manifest.sh`, `lib/*`), so Asad's review is the intended backstop — but the gate isn't enforced, so you're clear to merge on my approval.

PR #337, opened 2026-07-11
fix(proxy): NO_PROXY covers the cloud metadata IP (chart + both installers)

LukasWodka:
## Summary
The auto-augmented `NO_PROXY` lists omitted the cloud metadata IP `169.254.169.254`, so on proxy tenants metadata traffic would be sent to the corporate proxy (broken IMDS access; instance credentials transiting a middlebox for no reason). Adds the IP to the chart helper (`tracebloc.proxyEnv`), the bash installer defaults (`_augment_no_proxy`), and the Windows defaults (`Get-EffectiveNoProxy`) — the three lists stay in lockstep — and regenerates `manifest.sha256` for the two touched scripts. Chart bumped 1.9.3 → 1.9.4.

This is the chart/installer half of backend#803 item J; the runtime half (jobs-manager/ingestor NO_PROXY strings) ships separately in client-runtime alongside the #120 fix.

## Related
Ref tracebloc/backend#803 (item J). Ref tracebloc/client-runtime#120.

## Type of change
- [x] Bug fix

## Test plan
- `helm unittest client`: 25 suites / **267 tests pass**, including the three updated exact-value `NO_PROXY` assertions in `client/tests/proxy_env_test.yaml`
- `bats scripts/tests/cluster.bats`: **27 tests pass**, including a new `169.254.169.254` assertion on `_augment_no_proxy`
- Pester assertion added to `Get-EffectiveNoProxy` (runs in CI's Windows job)
- `scripts/gen-manifest.sh` regenerated — exactly the `cluster.sh` + `install-k8s.ps1` digests changed; CI `--check` gate stays green
- Note: `helm lint client` fails on develop already (schema requires non-empty `clientId`/`clientPassword`) — pre-existing, unrelated

## Checklist
- [x] Tests added / updated and passing locally
- [x] No secrets / credentials in the diff

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Narrow proxy-bypass list change with installer/chart parity and test coverage; no auth or data-model changes.
> 
> **Overview**
> Adds **`169.254.169.254`** (link-local cloud metadata) to the auto-augmented **`NO_PROXY`** lists so corporate-proxy tenants do not send IMDS traffic through the HTTP proxy.
> 
[...]

PR #340, opened 2026-07-12
fix(installer): Windows one-client guard reads clientId via ConvertFrom-Json (#200)

LukasWodka:
## Summary
The Windows one-client-per-machine guard only recognized an existing install when `helm get values` happened to render `clientId` in **double quotes** (`clientId:\s*"([^"]+)"`). Helm re-serializes stored values on `get`, so typical clientIds come back **unquoted** — the guard silently matched nothing and the installer would re-point the machine to a different Client ID, defeating the protection from #192.

Fix: read the values as **JSON** (`helm get values <release> -n <ns> -o json | ConvertFrom-Json`) and take `.clientId` directly — quoting disappears at that layer, and it matches the adjacent comment that already claimed ConvertFrom-Json. Handles `null` (release with no user values), a missing `clientId` key, and unparsable output from one release without aborting the scan of the remaining releases. Mirrors what the bash installer's `_extract_yaml_value` tolerance achieves (`scripts/lib/install-client-helm.sh`), but without regex-scraping YAML at all.

Tests: the two existing guard tests now serve real-helm-shaped output (JSON for `-o json`, unquoted YAML otherwise — so a regression back to YAML-scraping fails them), plus new Pester cases covering unquoted / single-quoted / double-quoted YAML views, a `null`-values release that must not abort the scan, and values without a `clientId` key that must not trip the guard.

`scripts/install-k8s.ps1` is on the signed-manifest surface — `scripts/manifest.sha256` regenerated via `scripts/gen-manifest.sh` (only the install-k8s.ps1 line changed; `--check` passes).

> **Note:** open PR #337 also touches `scripts/install-k8s.ps1` + `scripts/manifest.sha256` (+ the same test file) — whichever merges second gets a trivial manifest conflict; resolve by re-running `bash scripts/gen-manifest.sh`.

## Related
Closes #200 · Refs #192, #199

## Type of change
- [ ] Feature
- [x] Bug fix
- [ ] Tech-debt / refactor
- [ ] Docs
- [x] Security / hardening
- [ ] Breaking change

## Test plan
No `pwsh` on this machine, so the CI-equivalent suite ran in a `mcr.microsoft.com/powershell:lts-7.4` Linux container (same invocation as `installer-tests.yaml`):

- AST parse of both edited files — OK
- PSScriptAnalyzer at CI's error gate — no errors (warnings unchanged, all pre-existing lines)
- Pester >= 5.5 over `scripts/tests/` — **95 passed, 0 failed, 6 skipped** (skips are the `-Skip:(-not $IsWindows)` blocks, same as CI's Linux leg); all 7 one-client-guard tests green
- `bash scripts/gen-manifest.sh --check` — passes

The real-Windows leg runs in CI (`unit-pester`, windows-latest).

## Deployment notes
None — installer script only; ships with the next release artifact regeneration (manifest already updated).

## Checklist
- [ ] Docs updated if behavior or config changed — N/A (behavior restored to what the comment already claimed)
- [x] For security-sensitive paths: appropriate reviewer requested

---

> [!NOTE]
> **Medium Risk**
> Installer hardening on a security-sensitive guard; behavior change only blocks installs that should have been blocked, with good test coverage.
> 
> **Overview**
> Fixes **#200**: the Windows installer’s **one-client-per-machine** check no longer scrapes `helm get values` YAML with a double-quote regex. It now uses **`helm get values … -o json`** and **`ConvertFrom-Json`** to read **`.clientId`**, so unquoted Helm output can’t bypass the guard and silently allow replacing an existing client.
> 
> The scan is more resilient: failed or empty JSON, literal **`null`** (no user values), missing **`clientId`**, and bad JSON on one release **continue** to the next release instead of aborting or false-blocking.
> 
> **Pester** coverage mirrors real Helm (JSON for `-o json`, unquoted YAML otherwise), adds cases for YAML quoting styles, **`null`** releases, and values without **`clientId`**, and updates **`scripts/manifest.sha256`** for the signed **`install-k8s.ps1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c7ff2f39c4b5ba3be0965981ab0fcd774a2ef3fe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (dismissed):
@LukasWodka — **Approved.** The `ConvertFrom-Json` rewrite correctly restores the #200-broken guard and sidesteps every PowerShell trap (empty/`null` double-guarded, `foreach` instead of `.Count` so the array-vs-scalar quirk never bites); signed-manifest hash matches head (detail in my comment above).

Caveats — follow-up, none blocking:
- 🟡 namespace adoption for the same id in a different namespace isn't at parity with the bash installer — pre-existing, out of scope for #200.
- ⚪ **worth its own hardening issue:** the outer `} catch { }` fails the one-client guard *open* on a malformed `helm list` (a garbage response could let a second client through). Pre-existing — not introduced by this fix — but security-relevant, so I'd file it rather than lose it. The `catch { continue }` / non-zero-exit branches are also untested despite the PR body claiming them.

`install-k8s.ps1` is code-owned; gate isn't enforced, so you're clear to merge — Asad's review is the intended backstop.

aptracebloc (approved):

PR #137, opened 2026-05-20
Sync develop → main for v1.3.5 chart release

saadqbal:
## Summary

Brings main up to date with develop in preparation for tagging the
v1.3.5 chart release. Matches the established sync pattern (see
\`chore: sync main → develop after misrouted docs PRs (#108)\` for
the reverse direction; this direction is precedented in earlier
\"Merge pull request #N from tracebloc/develop\" merge commits).

After this merges, the next step is tagging \`v1.3.5\` against main,
which fires release-helm-chart.yaml and publishes BOTH the parent
\`client\` chart and the \`ingestor\` subchart to
tracebloc.github.io/client.

## What's in this sync

Since the last develop → main sync, develop has gained:

- **Declarative ingestion subchart & end-to-end UX**:
  - #123 ingestor Helm subchart + companion RBAC/service/authz
  - #126 \`INGESTOR_IMAGE_DIGEST\` env wiring on jobs-manager
  - #128 ingestor chart auto-resolves jobs-manager endpoint to release namespace
  - #131 parent client chart owns the shared ingestor ServiceAccount
  - #132 idempotency key defaults to install-time stamp (not release revision)
  - #133 ingestor subchart default \`serviceAccount.create=false\` (0.2.0, breaking)
  - #134 bump \`images.ingestor.digest\` to v0.3.0 + chart to 1.3.5
  - #136 publish ingestor subchart alongside parent chart in release workflow
- **Operational hardening**:
  - #112 auto-upgrade cronjob runs hourly at :23
  - #117 drop push-tags trigger from release-helm-chart workflow
  - #118 / #119 / #120 / #122 request-proxy wiring
  - #124 nil-guard ingestionAuthz access for \`--reuse-values\` upgrade path

## Conflict resolution

One conflict in \`client/Chart.yaml\`: main had \`version: 1.3.2\`,
develop had \`version: 1.3.5\`. Resolved by taking develop's version
since we're tagging v1.3.5 immediately after this merges.
\`MIGRATION.md\`, \`values.yaml\`, and \`docs/INSTALL.md\` auto-merged
cleanly.

## Test plan

- [x] All four platform-specific lints pass on the merged tree
      (\`client/ci/{aks,bm,eks,oc}-values.yaml\`)
- [x] \`helm lint --strict ./ingestor\` passes
- [ ] After merge → tag \`v1.3.5\` → release-helm-chart.yaml builds + pushes
[...]

PR #355, opened 2026-07-15
feat(installer): stop prompting for location during register — auto-derive from timezone (#354)

saadqbal:
Closes #354.

## What
The installer's **Register this machine** step prompted for **both** a client name and a location zone. This drops the location prompt: we now prompt for the **name only** and derive the carbon zone silently from the system timezone. When detection yields nothing, we provision with **no** location rather than asking.

## Why it's safe now
The location prompt was an explicit `INTERIM` measure (`provision.sh`), kept only because the *deployed* CLI hard-required `--name`/`--location` without a TTY. **cli#137 has since shipped** — in CLI `v0.9.2` and `main` — so `tracebloc client create` now auto-names `<firstname>-NN` and treats `--location` as fully optional (never prompted, never required; sends nothing when absent, backend records no location). The installer pulls the CLI from `releases/latest`, so it already has this behavior. No CLI-repo change is needed.

## Behavior
| Case | Before | After |
|---|---|---|
| Interactive install | prompts name **and** location (Enter to accept detected zone) | prompts **name only**; zone auto-derived from timezone |
| Timezone not detectable | required-entry loop until a zone is typed | provisions with **no** `--location`, no prompt |
| `TRACEBLOC_CLIENT_LOCATION` set | used, prompt skipped | used, still overrides (unchanged) |
| Unattended (env name) | silent auto-derive | silent auto-derive (unchanged) |

## Changes
- `scripts/lib/provision.sh`: remove the interactive location block (detected-zone accept + required-entry loop); refresh the now-stale interim comments and the `_report_create_failure` hint that referenced the removed prompt.
- `scripts/tests/provision.bats`: +2 tests (auto-derive from timezone → `--location`; no detection → no `--location`, never prompts). Full suite green (23/23).
- `scripts/manifest.sha256`: regenerated (R8 Static-analysis gate).

shellcheck clean.

---

> [!NOTE]
> **Low Risk**
> Installer UX and provisioning-args logic only; name prompting and env overrides are unchanged, with new bats tests locking behavior.
> 
> **Overview**
> Interactive **Register this machine** now asks only for the **client name**; carbon zone is **never** prompted.
> 
> When `TRACEBLOC_CLIENT_LOCATION` is unset, `provision.sh` silently derives a zone from the system timezone and passes `--location` only when one is found. If detection fails, it calls `client create` **without** `--location` instead of blocking on a required-entry loop. Invalid-zone errors now tell users to set `TRACEBLOC_CLIENT_LOCATION` rather than referring to the removed prompt.
> 
> `provision.bats` adds coverage for silent auto-derive and no-location provisioning; `manifest.sha256` is updated for the changed script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ed7c27dbb9772b021d005c413ad7885112f525d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (approved):
@saadqbal — **Approved.** Correctly-sequenced removal of the interim location prompt (the "KEEP THIS PROMPT until cli#137 ships" block from #341) — and importantly you waited for cli#137 to actually *release*, not just merge. I verified the load-bearing bits:

- **✅ cli#137 is shipped, not just closed.** At the **v0.9.2** tag (released today, the latest the installer pulls), `client create`'s `--location` is a plain optional `StringVar` with **no `MarkFlagRequired`** — "optional … never prompted, never required … omitted if unset," request field `omitempty` (backend records no location, not a silent default), plus auto-naming. So a no-detection `curl | bash` install sending `client create --name X` with no `--location` now succeeds — the exact breakage the interim comment guarded against is gone.
- **✅ Manifest hash matches** — recomputed `provision.sh` = `bf283e1…` == `manifest.sha256`. R8 intact.
- **✅ `--location` conditionally passed** — `[[ -n "$client_location" ]] && _create_args+=(--location "$client_location")`; never `--location ""`, clean array construction.
- **✅ Clean removal** — parses, no orphaned refs to the deleted locals/prompt strings, `_detect_location_zone` still defined + used by the silent fallback, and `_report_create_failure`'s hint was updated in lockstep to point at `TRACEBLOC_CLIENT_LOCATION`.
- **✅ Tests pin both paths** — detection → `--location FR` silently, no prompt; no-detection → no `--location`, no prompt, non-fatal. Both feed only a name on stdin, so a lingering prompt would hang the test — good structural guard. (bats isn't installed locally; I parse-checked + logic-reviewed, CI runs the suite.)

## ⚪ Minor (non-blocking)
- No explicit CLI-version gate for optional-location: safety rests on "latest release carries cli#137" (v0.9.2 does). A run that pins an *older* `TRACEBLOC_CLI_VERSION` **and** has no detectable zone **and** no `TRACEBLOC_CLIENT_LOCATION` would send a `--location`-less `client create` to a pre-#137 CLI and fail. Edge case — the env override covers it and `_report_create_failure` surfaces the real error rather than failing blind — but a one-line note wouldn't hurt.

Touches your code-owned `provision.sh` + manifest (R8); you're the author, so this is the approving second set of eyes. Clear to merge.

PR #362, opened 2026-07-21
Installer: clean setup output + fix name-prompt escape garble

LukasWodka:
## Summary
Two installer-polish items on the macOS/Linux install path (`scripts/install-k8s.sh` + libs), plus Windows parity (`install-k8s.ps1`).

### 1. Setup output copy — matches the agreed spec
Before, the setup step printed off-guideline noise (a stray `Homebrew`, `System tools ready (k3d, helm, kubectl)`, tool-version dumps, and a `verified on your PATH` claim). Collapsed to exactly:

```
tracebloc needs your password once to set up Docker and a few tools.
Password:
✔ Docker ready
✔ System tools ready
✔ tracebloc CLI updated (v0.9.2 → v0.9.3) — run `tb` to use it
```

- `common.sh` — password hint `install` → `set up`.
- `setup-macos.sh` — Homebrew lines → `log()` (hidden); drop the `(k3d, helm, kubectl)` parenthetical.
- `install-cli.sh` — collapse the *already-present / re-running / installing / ready* pileup into **one** `✔` line: `updated (vX → vY)` when the CLI was upgraded (prior version captured before the installer runs), else `up to date` / `ready`. Edge-case lines (installed-but-not-yet-on-PATH) keep their actionable, shell-correct PATH-fix hints.
- `install-k8s.ps1` — Windows parity for the same three.

### 2. Name-prompt escape garble (customer-reported 2026-07-20)
Arrow keys at the *"Name this client"* prompt injected raw `ESC[D`/`ESC[A` bytes that slug-ified into a garbage name (`d-d-d-a-a-a`) when passed to `client create`. The credential prompt already stripped these; the name prompt did not.

- `provision.sh` — `read -e` (readline line-editing: arrows move the cursor instead of emitting escapes) + strip anything that still slips through, using the same helper the credential path uses.
- Reword the prompt to **"Name your secure environment"** (terminology SoT).
- Move `_strip_paste_garbage` from `install-client-helm.sh` → `common.sh` so the credential path and the name prompt share one sanitizer (removes a fragile cross-file dependency; fixes bats isolation).

## Test plan
- `bats scripts/tests/*.bats` — green except two failures that **pre-date this change** and reproduce on clean `develop` (`validate_config: valid config passes`, `_extract_yaml_value: single-quoted with '' escape` — local bash/locale artifacts, unrelated to this diff).
- `install-cli.bats` assertions updated for the new `✔` copy; Pester assertions in `install-k8s.Tests.ps1` likewise (pwsh not runnable locally — Windows verified in CI).
- `bash -n` + `shellcheck` clean on all changed libs.

## Notes
- `read -e` is belt-and-suspenders *with* the strip, not instead of it — a non-readline paste can still deliver escapes.
- CLI-side hardening (`tracebloc client create --name` rejecting/stripping control chars as defense-in-depth) is a **separate follow-up in the `cli` repo**.

---

> [!NOTE]
> **Low Risk**
> Changes are installer messaging, input sanitization, and non-fatal CLI install UX; no auth, data-plane, or cluster logic changes.
> 
> **Overview**
> **Installer output** is aligned to a shorter spec: the sudo hint says “set up” instead of “install”; macOS Homebrew status moves to `log()` instead of green checks; system-tools success drops the tool list parenthetical; CLI install drops redundant “already present / re-running” chatter and ends with **one** ✔ (`ready`, `up to date`, or `updated (vX → vY)`), preferring `tb` when that alias exists. CLI install on bash uses `spin` instead of `spin_cmd` so recoverable failures stay soft. **Windows** (`install-k8s.ps1`) mirrors the CLI success wording.
> 
> **Name prompt fix:** provisioning rewords the prompt to “Name your secure environment”, uses plain `read` (avoids garbled echo under tee), and runs input through **`_strip_paste_garbage`** moved from `install-client-helm.sh` into **`common.sh`** so credentials and the name prompt share one sanitizer.
> 
> **Tests:** `install-cli.bats` and `install-k8s.Tests.ps1` expectations updated; `manifest.sha256` hashes refreshed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 274b9b486c0a17b4219f38bb929581207531957b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (approved):
Verified — approving.

- **Name-garble fix is correct:** plain `read -r` + `_strip_paste_garbage`. The strip is the real fix (readline `read -e` was reverted because it echoes to stderr and garbles under the installer's `tee` redirect — Bugbot's resolved "readline echo breaks under tee").
- **The shared sanitizer mirrors the CLI side (#364) exactly** — same CSI pattern `ESC\[[0-9;]*[A-Za-z~]`, same order (CSI sequences iteratively → literal paste markers → C0/DEL last), UTF-8 preserved. Moving it into `common.sh` so the credential and name paths share one sanitizer is the right consolidation. #362 (strip at the installer source) + #364 (strip at the CLI `--name` boundary) are the two consistent halves of defense-in-depth for this customer bug.
- **Manifest integrity verified:** I recomputed the sha256 of all three changed lib scripts on the branch and they match `manifest.sha256` exactly (provision `725a85e4…`, common `1e035f5d…`, install-cli `50711ae0…`), so the installer's self-verification stays valid. CI green, Bugbot clean.

**One doc nit (please fix the description):** the body says the fix uses `read -e`, but the code reverted to plain `read -r` (the Bugbot tee fix). Worth correcting so the description matches the shipped code.

LGTM.

PR #370, opened 2026-07-22
feat(install): host capability/privilege probe + audit (RFC 0001, #1171)

LukasWodka:
## What

Foundation for the least-privilege install (RFC 0001) — the **sensing + reporting** layer. New `scripts/lib/probe.sh`:

- **Side-effect-free probes** (read-only, never-fatal):
  - `docker info` exit 0 = a runtime is usable *as this user* (binary + daemon + socket permission in one call — no image pull).
  - cgroup v2 (`/sys/fs/cgroup/cgroup.controllers`) + unprivileged userns (`/proc/sys/user/max_user_namespaces`, `kernel.unprivileged_userns_clone`).
  - privilege posture: `id -u` / `command -v sudo` / `sudo -n true` → `root | sudo_nopw | sudo_pw | no_sudo`.
- **`classify_install_tier`** → the lowest workable tier: **0** (runtime usable) · **1** (no runtime but kernel is rootless-capable) · **2** (old/locked kernel, or non-Linux with no runtime). A usable runtime always wins Tier 0.
- **`render_host_audit`** → the "Host check" panel from the RFC.
- **`--verify` (`TB_PROBE_VERIFY=1`)** gates the only image-pulling probe (`hello-world`); it is never on the default path.

## Wiring (deliberately conservative)

Tier **routing** is the follow-up (#1172). Surfacing a "Tier 0 — no privileged steps" conclusion in the main flow *before* routing honors it would contradict the still-current sudo path — so this PR wires the audit only into the **read-only `--diagnose`** path (a console panel + a plain `## install tier` readout in the support bundle). **Nothing in the install path changes behaviour yet.** #1172 will consume `INSTALL_TIER` to branch the flow and surface the panel where it's actionable.

## Integrity / tests

- `probe.sh` registered in **both** `FILES` arrays (`install.sh` + `gen-manifest.sh`); R8 manifest regenerated (`--check` green).
- Guarded `source` in `install-k8s.sh` (a stale bootstrap that didn't fetch it degrades — `--diagnose` just omits the section).
- **15 new bats tests** (`scripts/tests/probe.bats`): tier truth table, the four privilege postures, the read-only guarantee (asserts no `docker run`/`pull` on the default path), and the audit panel. Full suite **318 green**; `shellcheck --severity=warning` clean; `bash -n` clean.

Part of the least-privilege install epic (#1168). **Stack:** first of the Wave-0 tickets; #1172/#1173 build on this.

[...]

PR #372, opened 2026-07-22
fix(install): root-aware sudo (A2, #1173)

LukasWodka:
## What (A2)

The installer primed and ran ~70 privileged steps as `sudo <cmd>`. Two real users hit the wall:

- **A root container / VM** — sudo is unnecessary, and often *absent* on minimal images, so `sudo <cmd>` died with `sudo: command not found` even though `<cmd>` would have run fine. (`id -u` returned `0`, yet the install failed.)
- **"sudo isn't installed"** was reported as **"no sudo access"** — a different and misleading fix.

## How

Shadow `sudo` with a function in `common.sh`:

```sh
sudo() {
  if [ "$(id -u)" -eq 0 ]; then "$@"            # root: run directly, no sudo binary
  elif _have_sudo_bin;      then _real_sudo "$@" # non-root: the real sudo
  else return 127; fi                            # mimic sudo's not-found (best-effort friendly)
}
```

- **Every existing `sudo <cmd>` call site is fixed with no edit** — the diff is confined to `common.sh` (only `preflight_sudo` and the sole `sudo -v`/`sudo -n` option-forms needed touching, via `_real_sudo`).
- The no-sudo branch returns non-zero rather than exiting, so best-effort sites like `sudo modprobe … || fallback` still take their fallback. `preflight_sudo` aborts up front with an accurate message on the required path.
- `preflight_sudo` now: root → return 0 (no priming); non-root + no sudo → **accurate** error; non-root + passwordless → return 0; else prime once via `_real_sudo -v`.
- `type -P` finds the real binary past the shadow (a bare `command -v sudo` would resolve to the function). Helper seams `_have_sudo_bin` / `_real_sudo` make every branch unit-testable without a real sudo.

## Tests

- 6 new `common.bats` cases: root runs directly (no real sudo), non-root defers to real sudo, no-sudo returns 127, and the three `preflight_sudo` postures (root / no-sudo error / passwordless).
- Full suite adds **no new failures** (3 failures pre-exist on clean develop: macOS bash-3.2 blindspot + locally-unstamped `DEFAULT_REF`; green on Linux CI). No new shellcheck findings; R8 manifest regenerated.

Part of the least-privilege install epic (#1168), Wave 1. Independent of #1171 (self-contained; #1172 routing will consume both).

Closes tracebloc/backend#1173

---

> [!NOTE]
> **Medium Risk**
> Changes how every privileged installer step runs (root vs sudo vs missing sudo); scope is shell helpers and messaging but mistakes could affect Docker/system package setup on diverse Linux hosts.
> 
> **Overview**
> **Root-aware privileged execution (RFC 0001 A2)** — `common.sh` now shadows `sudo` so existing `sudo <cmd>` call sites run the command directly when uid 0, delegate to the real binary when present, or return 127 when neither applies (preserving `sudo … || fallback` behavior). Helpers `_have_sudo_bin` / `_real_sudo` and `export -f` keep nested `bash -c` subshells on the same path.
> 
> **`preflight_sudo`** short-circuits on root, fails early with an explicit “sudo isn’t installed” message when appropriate, and uses `_real_sudo` for `-v` / `-n` and the credential keepalive.
> 
> **`probe.sh`** `_probe_privilege` uses those same helpers so host audit no longer treats the shadow as a real sudo binary (Bugbot #372).
> 
> Bats coverage for sudo/preflight branches, subshell inheritance, and `set -e` safety on old bash; probe privilege tests updated; `manifest.sha256` refreshed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 71f8d91ddb2d76e10980bbb772e291df8b56e469. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


divyasinghds (approved):

divyasinghds (dismissed):
Approve. Root-aware sudo shadow. Verified no option-led (sudo -flag) call sites exist, so the root branch's "$@" is safe; -n/-v correctly route through _real_sudo. export -f / env_reset interaction reasoned through correctly. Highest blast radius in the stack but well-tested. Optional follow-up: a CI guard forbidding 'sudo -' outside _real_sudo to prevent regressions.

divyasinghds (approved):
Re-reviewed at head 71f8d91 (3 new commits since my prior approval on 43cb51c; #370 has since merged to develop, so probe.sh is now in-tree). The new changes are correct and one fixes a real bug:

- `probe.sh` `_probe_privilege` now detects the real sudo **binary** (`_have_sudo_bin`, via `type -P`) and bypasses the shadow (`_real_sudo`) instead of `has sudo` / a bare `sudo -n true`. This is the exact shadow interaction I raised on first pass: once the A2 `sudo()` function exists, `command -v sudo` resolves to it and a bare `sudo -n true` returns 127 on a sudo-less host — so the posture would be misreported as `sudo_pw` instead of `no_sudo`, which would defeat #374's `Tier 2 + no_sudo` fail-fast and let a genuinely-stuck host proceed and crash mid-install. Correctly fixed.
- `_have_sudo_bin` dropped the `[ -n "$(type -P sudo)" ]` command substitution for a whole-body `type -P sudo >/dev/null 2>&1` — `set -e`-safe on bash <4.4 (Amazon Linux 2's 4.2 / macOS 3.2), stderr silenced, and still finds the real binary past the shadow. Dedicated regression test added.

Coupling is sound: `common.sh` is sourced unconditionally before `probe.sh`, and this PR ships both edits plus the regenerated manifest atomically. CI green (29 pass, 1 skip); `probe.bats` / `common.bats` updated to match. Approve.

PR #377, opened 2026-07-22
feat(install): prepare-host admin step (#1178, client)

LukasWodka:
## What (#1178 — prepare-host, client side)

The standalone, **admin-run Tier-2 step**. An administrator runs it once on a host a researcher can't install on unprivileged; afterwards the researcher installs at **Tier 0** with no admin.

- Invoked as `curl -fsSL https://tracebloc.io/i.sh | bash -s -- prepare-host` (the bootstrap already forwards `"$@"` to `install-k8s.sh`, so no bootstrap change was needed) or, later, `tracebloc prepare-host`.
- `run_prepare_host` does **only the privileged prerequisites**, reusing the exact functions the full install uses: `preflight_sudo` + `install_docker_engine` + `install_system_deps`. It then grants the researcher (`SUDO_USER`, or `TB_PREPARE_USER`) **docker-group** access — so their later install is Tier 0.
- It **never** mints a credential, creates a cluster, or installs the CLI — an admin can safely run it on a shared host without provisioning anything as themselves.
- Terminal mode in `main()` (mirrors `--diagnose`). Non-Linux gets a Docker-Desktop / WSL2 pointer.

## Tests

- 3 new `setup-linux.bats`: installs the runtime prereqs + adds the researcher to the docker group (and does **not** create a cluster / install the CLI); best-effort when there's no target user; non-Linux errors with the right pointer.
- Full suite adds **no new failures**; shellcheck clean; R8 manifest current.

## Scope

Client half of #1178. The `cli` `tracebloc prepare-host` subcommand (a thin wrapper that runs `… | bash -s -- prepare-host`, like `tracebloc upgrade`) is a separate cli PR. The rootless/userns-enable nuance for hosts that need Tier 1 lives with #1177.

## Stack

Stacked on **#1175** → #1172 → #1171. Base is `feat/lpi-tier0`; merge in order.

Part of the least-privilege install epic (#1168), Wave 2.

Part of tracebloc/backend#1178 (client half; cli half to follow).

---

> [!NOTE]
> **Medium Risk**
> Runs privileged package/Docker setup and `usermod` on shared hosts; behavior is narrowed to an explicit admin step with guarded messaging and tests, but mis-set `TB_PREPARE_USER` could still leave researchers without socket access until fixed manually.
> 
> **Overview**
> Adds a **one-time administrator** path so shared Linux hosts can be prepped without running the full tracebloc install as the admin.
> 
> **`prepare-host` / `--prepare-host`** is a terminal mode in `install-k8s.sh` (like `--diagnose`): it skips provisioning, cluster creation, and CLI install, uses a lightweight EXIT trap to stop the sudo keepalive from orphaning (#377), and calls **`run_prepare_host`** in `setup-linux.sh`.
> 
> **`run_prepare_host`** runs the same privileged pieces as a full install (`preflight_sudo`, Docker engine, system deps) and optionally adds a named researcher to the **`docker`** group via **`TB_PREPARE_USER`** only—no fallback to **`SUDO_USER`** (#377). Success text promises a no-admin install only when **`usermod`** actually succeeded.
> 
> Five new **`setup-linux.bats`** cases cover happy path, missing user, failed grant, admin-not-granted, and non-Linux. **`manifest.sha256`** is updated for the touched scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a625b09ead20ca1e218872251f302f30362f3be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/setup-linux.sh
```diff
@@ -495,3 +495,50 @@ install_linux() {
...
+run_prepare_host() {
+  if [[ "$OS" != "Linux" ]]; then
+    error "prepare-host is for Linux hosts. On macOS/Windows, install Docker Desktop (or enable WSL2) as an administrator, then run the installer normally."
+  fi
+  export DEBIAN_FRONTEND=noninteractive NEEDRESTART_MODE=a NEEDRESTART_SUSPEND=1
+
+  step_header a "Preparing this host for tracebloc (one-time administrator step)"
+  if declare -F host_audit >/dev/null 2>&1; then host_audit; fi
+  echo ""
+
```
divyasinghds, line 593: **docker-group grant targets the wrong user in the primary flow.**

The intended scenario (per the cli#394 help) is: an **admin** preps a shared/HPC box so a **different** researcher — who has no sudo — can then install at Tier 0. But this line resolves the target from `SUDO_USER`:

- When a sudoer admin runs `sudo tracebloc prepare-host` (the common case), `SUDO_USER` is the **admin**, so we `usermod -aG docker <admin>` and print *"Added `<admin>` … they can now install with no admin."* The researcher is never granted access, and the message reads as success.
- The `TB_PREPARE_USER` tip only prints when `target` is empty/root, so in exactly this case the admin never sees the escape hatch.
- The `SUDO_USER`-is-the-researcher case essentially never happens in the real workflow: if the researcher had sudo, they wouldn't need an admin to prep the host.
- `tracebloc prepare-host` (cli#394) is `cobra.NoArgs` with no `--user` flag and no mention of `TB_PREPARE_USER` in its help — so there's no discoverable way to name the researcher.

**Suggested fix:** add a `tracebloc prepare-host <user>` arg (or `--user`) plumbed to `TB_PREPARE_USER` in cli#394, and/or always surface the `TB_PREPARE_USER` hint (including on the sudoer-admin path) rather than silently treating `SUDO_USER` as the researcher. Everything else in this PR looks correct.

divyasinghds (approved):
Approve. My earlier finding on the docker-group targeting is fully resolved:

- `local target="${TB_PREPARE_USER:-}"` — the `$SUDO_USER` fallback is gone, with a comment explaining SUDO_USER is the admin, not the researcher.
- New `granted` flag gates the closing message, so it no longer falsely promises "no administrator rights" when no grant happened (no TB_PREPARE_USER, or usermod failed) — that was a secondary half of the same problem, and it's handled.
- Four new tests cover it, including one explicitly asserting SUDO_USER (the admin) is NOT added, and the usermod-fails path.

The rest still checks out: reuses the exact privileged install functions (no drift), terminal mode with the keepalive-reaping EXIT trap, non-Linux pointer. `$OS` is set at source time in common.sh so the guard is safe despite skipping run_preflight. CI green. LGTM.

PR #443, opened 2026-07-27
fix(installer): make the bats suite green on stock macOS (bash 3.2 + tmpdir symlink + host-CLI leak)

LukasWodka:
## Summary
Three bats tests failed when the suite runs locally on macOS but passed in ubuntu CI (verified on a clean origin/develop checkout) — noise that masks real regressions for anyone running `bats scripts/tests/*.bats` before pushing. All three are local-environment artifacts; this fixes each at its actual root cause and silences the one bats warning the suite emitted. Suite is now 468/468 green on stock macOS (bash 3.2, bats 1.13) with zero warnings, and unchanged on Linux.

## Root causes & fixes

**1. `_extract_yaml_value: single-quoted with '' escape` (install-client-helm.bats) — bash 3.2 bug in the lib code.**
In `${var//pat/rep}`, macOS's bash 3.2 keeps the backslash when the replacement is an escaped quote: the `''` de-escape turned `'a''b'` into `a\'b` (bash 4+/5 gives `a'b`). Fixed by building pattern/replacement from a `$sq` variable — identical semantics on every bash. The mirror-image escape site (`' -> ''` for `TB_CLIENT_PASSWORD_ESCAPED`) had the same latent bug and gets the same two-line fix. `scripts/manifest.sha256` regenerated since this lib is on the R8 integrity surface (`gen-manifest.sh --check` passes).

**2. `validate_config: valid config passes` (common.bats) — tmpdir symlink shape.**
macOS puts `BATS_TEST_TMPDIR` under the `/var -> /private/var` symlink. `validate_config` resolves the data dir via `cd -P` (symlinks resolved) but compared it against the unresolved `$HOME`, spuriously failing the under-`$HOME` check. Fixed the test by pre-resolving `HOME` with `cd -P` — the exact pattern (and comment rationale) the sibling tilde tests in the same file already use; this first test simply predated them.

**3. `un-stamped DEFAULT_REF fails closed` (install-bootstrap.bats) — host CLI leaking into the sandbox.**
This is the only bootstrap test that sets no `REF`/`BRANCH`, so `_tb_bail_ok` stays armed — and `run_boot`'s `PATH="$BIN:$PATH"` let the dev box's real `/usr/local/bin/tracebloc` satisfy `command -v tracebloc`. The test then ran the host's real `tracebloc doctor` and, the box being healthy, exec'd the real CLI, which exited 0 before the REF gate was ever reached (CI has no tracebloc binary, so it never saw this). Added `run_boot_hermetic` — `PATH=$BIN` alone (setup already symlinks the needed coreutils for exactly this pattern) plus a sandboxed `HOME`, since the bootstrap re-prepends `~/.local/bin` to PATH. Beyond determinism, this stops the suite from executing a real CLI binary on dev machines.

**4. bats `BW01` warning on `sudo modprobe overlay` (common.bats).**
The `sudo(): non-root without sudo returns 127` test deliberately expects 127, which bats flags as "command not found". Switched to `run -127` plus the documented `bats_require_minimum_version 1.5.0` declaration (CI's apt bats is 1.10, fine with both).

## Related
Local dev-environment fix — no tracked issue.

## Type of change
- [x] Bug fix
- [x] Tech-debt / refactor

## Test plan
- macOS (bash 3.2.57, bats 1.13.0): full suite `bats scripts/tests/*.bats` — 468/468 pass, zero warnings (was: 3 failures + BW01)
- `scripts/gen-manifest.sh --check` — manifest up to date
- CI-equivalent lint locally: `bash -n` over all scripts + `shellcheck --severity=error` on the same file set as the Lint job — clean
- `bash scripts/check-style.sh` — clean
- Behavior spot-check of the substitution fix on bash 3.2: `'a''b'` unescapes to `a'b`; `a'b` escapes to `a''b`

## Checklist
- [x] Docs updated if behavior or config changed (N/A)
- [x] For security-sensitive paths: appropriate reviewer requested (manifest regenerated; no gate/verify logic touched)
- [x] Terminal output follows STYLE.md — `bash scripts/check-style.sh` passes
- [x] Cross-repo issues use `Fixes tracebloc/<repo>#N` (N/A)

---

> [!NOTE]
> **Low Risk**
> Changes are installer script portability, values-file quoting hardening, and test isolation; manifest hash updated but no verify/bootstrap gate logic changed.
> 
> **Overview**
> Makes the installer bats suite reliable on **stock macOS (bash 3.2)** while tightening how credentials land in generated Helm values.
> 
> **Helm values / YAML:** Adds shared `_yaml_sq_escape` / `_yaml_sq_unescape` so single-quoted YAML scalars work on bash 3.2 (escaped-quote replacements must use a variable, not `\'`). **`clientId`** is now written like **`clientPassword`**—single-quoted with escaping—instead of a double-quoted scalar that could break on `"` or `\`. **`scripts/manifest.sha256`** is updated for the changed lib on the integrity surface.
> 
> **Tests:** **`validate_config`** resolves **`HOME`** with `cd -P` so macOS `/var` → `/private/var` tmpdirs don’t fail the under-`$HOME` check. Bootstrap’s unstamped-ref test uses **`run_boot_hermetic`** so a host `tracebloc` CLI can’t bail out before the REF gate. **`run -127`** for the sudo mock test plus **`bats_require_minimum_version 1.7.0`**. New bats cover YAML escape round-trips and `clientId` quoting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f0872ad67eaa8c69260446e0fc8853f4f31e9de. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/install-client-helm.sh
```diff
@@ -791,7 +791,11 @@ install_client_helm() {
     TB_NAMESPACE="$existing_ns"
   fi
 
-  TB_CLIENT_PASSWORD_ESCAPED="${TB_CLIENT_PASSWORD//\'/\'\'}"
+  # ' -> '' via $_sq, not \': bash 3.2 keeps the backslash in an escaped-quote
+  # replacement (same constraint as _extract_yaml_value's unescape above), which
+  # would corrupt a password containing a quote into a\'\'b inside the values file.
+  local _sq="'"
+  TB_CLIENT_PASSWORD_ESCAPED="${TB_CLIENT_PASSWORD//$_sq/$_sq$_sq}"
```
saqlainsyed007: Robustness (non-blocking): the password value is now single-quote-escaped, but its sibling `clientId` on line 883 is interpolated into a **double-quoted** YAML scalar with no escaping — `clientId: "$TB_CLIENT_ID"`. `_sanitize_credential` only strips paste/non-printable chars, so a client ID containing a `"` or `\` would corrupt values.yaml — the same bug class this PR fixes for the password one line below. Largely gated by `verify_credentials` (valid IDs are UUIDs), so low severity, but the interpolation itself is unguarded.

scripts/tests/common.bats
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bats
 # Tests for scripts/lib/common.sh — config validation, the install_cleanup
 # CLIENT_STATE guard (#716), retry, has.
+bats_require_minimum_version 1.5.0   # for `run -<code>` (expected-exit syntax)
```
saqlainsyed007: `bats_require_minimum_version` was itself introduced in bats-core **1.7.0**, so requiring `1.5.0` here is internally inconsistent: on a bats 1.5.x–1.6.x, `run -<code>` is supported but *this line* is an undefined command and errors the file out before the guard can help. The effective floor is 1.7.0 — consider `bats_require_minimum_version 1.7.0` (CI's 1.10 and your 1.13 both satisfy it).

scripts/tests/common.bats
```diff
@@ -348,7 +352,7 @@ setup() {
 @test "sudo(): non-root without sudo returns 127 (best-effort friendly), never exits" {
   id() { echo 1000; }
   _have_sudo_bin() { return 1; }
-  run sudo modprobe overlay
+  run -127 sudo modprobe overlay   # -127: expected here — silences bats BW01
```
saqlainsyed007: With `run -127`, bats already asserts the status is 127 and fails the test otherwise — so the following `[ "$status" -eq 127 ]` (line 356) is now dead (can never fail if reached). Safe to drop it now that `run -127` enforces the expectation.

scripts/lib/install-client-helm.sh
```diff
@@ -791,7 +791,11 @@ install_client_helm() {
     TB_NAMESPACE="$existing_ns"
   fi
 
-  TB_CLIENT_PASSWORD_ESCAPED="${TB_CLIENT_PASSWORD//\'/\'\'}"
+  # ' -> '' via $_sq, not \': bash 3.2 keeps the backslash in an escaped-quote
+  # replacement (same constraint as _extract_yaml_value's unescape above), which
+  # would corrupt a password containing a quote into a\'\'b inside the values file.
+  local _sq="'"
```
saqlainsyed007: This `$_sq` single-quote escape is now the second copy of the same bash-3.2-safe idiom (the first is the *unescape* in `_extract_yaml_value`, ~line 148). Since both carry cross-referencing comments to stay in sync, a tiny shared helper (`_yaml_sq_escape` / `_yaml_sq_unescape`) would centralize the portability rule and prevent future drift. Non-blocking.

scripts/tests/install-bootstrap.bats
```diff
@@ -115,6 +115,15 @@ run_boot_no_cosign() {
   PATH="$BIN" run bash "$BOOT" "$@"
 }
 
+# Run hermetically: PATH=$BIN alone AND a sandboxed HOME. Needed by any test
```
saqlainsyed007: Minor wording: this says `run_boot_hermetic` is “Needed by any test that passes NO REF/BRANCH,” but the bailout tests at lines ~233/246/258 also pass no REF/BRANCH and correctly stay on `run_boot` — they're fine because each installs its own mock `tracebloc` into `$BIN`. The real distinguishing condition is “no REF/BRANCH **and** no mock tracebloc provided.” Worth tightening so a future reader doesn't think those tests are wrong.

saqlainsyed007 (commented):
Reviewed the escaping fix + test-portability changes. The core is sound and I verified the load-bearing claims locally: the bash 3.2 quote-corruption is real and the `$_sq` fix produces the correct `a''b` on 3.2.57; `manifest.sha256` matches the new file and `gen-manifest.sh --check` passes; the `validate_config` HOME/`cd -P` fix and the hermetic-runner isolation both target the right root cause. No blocking bugs.

A few non-blocking observations inline — #1 (clientId) and #2 (bats version floor) are the two worth a look.

scripts/lib/install-client-helm.sh
LukasWodka: Good catch — fixed in 4f0872a. You're right that it's the same bug class, and right that `_sanitize_credential` doesn't help: it only calls `_strip_paste_garbage`, which strips paste markers and C0 controls and leaves `"` and `\` untouched.

Both credentials now go through the shared escaper into **single-quoted** scalars:

```
clientId: '$TB_CLIENT_ID_ESCAPED'
clientPassword: '$TB_CLIENT_PASSWORD_ESCAPED'
```

In a single-quoted YAML scalar a `"` is literal, so it can no longer terminate the scalar early. Agreed on severity — `verify_credentials` gates real IDs to UUIDs, so this is hardening rather than a live break, but the interpolation itself is now safe either way.

Reading stays backward compatible: `_extract_yaml_value` still handles the double-quoted form, so values files written by older installers parse unchanged. I deliberately left the double-quoted **inputs** and the mocked `helm get values` output in the tests as-is to keep covering that path, and moved only the five assertions on the file the installer *generates*.

Tests added: clientId round-trips through a single-quoted scalar with a `'` in the value, a `"` no longer breaks the scalar, and a guard that the heredoc uses the escaper rather than raw interpolation.

scripts/tests/common.bats
LukasWodka: You're right, and it was self-defeating in exactly the way you describe. Fixed in 4f0872a — raised to `1.7.0`.

Confirmed against bats-core's changelog:

- **1.5.0** (2021-10-22) — *"Experimental: add return code checks to `run` via `!`/`-<N>` (#367, #507)"*
- **1.7.0** (2022-05-14) — *"bats_require_minimum_version to guard code that would not run on older versions (#587)"*

So on a 1.5.x–1.6.x bats, `run -127` would work but **that line** is an undefined command and takes the file down before the guard can do anything. 1.7.0 is the real floor, and as you note CI's 1.10 and my local 1.13 both satisfy it (verified locally — the file runs clean with `1.7.0`).

scripts/tests/common.bats
LukasWodka: Correct — dropped in 4f0872a. I verified it empirically rather than assuming, because it matters *where* bats fails:

```
@test "run -127 with a MISMATCHED status" {
  run -127 bash -c 'exit 3'
}
@test "run -127 mismatch: is the line after even reached?" {
  run -127 bash -c 'exit 3'
  echo "REACHED-THE-LINE-AFTER" >&3
  [ "$status" -eq 127 ]
}
```

```
not ok 2 run -127 with a MISMATCHED status
#   `run -127 bash -c 'exit 3'' failed, expected exit code 127, got 3
not ok 3 run -127 mismatch: is the line after even reached?
#   `run -127 bash -c 'exit 3'' failed, expected exit code 127, got 3
```

`REACHED-THE-LINE-AFTER` never prints — bats aborts at the `run` line, so the trailing assertion is unreachable, not merely redundant. Replaced it with a comment recording that `run -127` is what enforces the expectation.

scripts/lib/install-client-helm.sh
LukasWodka: Done in 4f0872a — agreed this was the right shape. Two one-liners now hold the rule:

```bash
_yaml_sq_escape() {                      # raw value -> body of a '...' scalar
  local _sq="'"
  printf '%s' "${1//$_sq/$_sq$_sq}"
}
_yaml_sq_unescape() {                    # body of a '...' scalar -> raw value
  local _sq="'"
  printf '%s' "${1//$_sq$_sq/$_sq}"
}
```

The bash-3.2 rationale is stated once above them instead of twice with cross-references, and `_extract_yaml_value` plus both credential call sites just call them. That also made your `clientId` finding a one-line fix rather than a third copy of the idiom.

Pinned with tests on both directions plus a round-trip over quote-heavy values (`a'b`, `'`, `''`, `it's a 'test'`, and a no-quote control).

scripts/tests/install-bootstrap.bats
LukasWodka: Fair point — you're right that the bail-out tests are fine for a different reason than my comment implied. Reworded in 4f0872a.

Verified: the three bail-out tests (~226/242/254) each `cat > "$BIN/tracebloc"` and `chmod +x` it, so they pass no REF/BRANCH *and* are hermetic by construction. The real distinguishing condition is "no REF/BRANCH **and** no mock `tracebloc` of its own", which is what the comment now says — plus an explicit note that tests providing their own mock correctly stay on plain `run_boot`.

aptracebloc (approved):
Approving — verified all three fixes: bash-3.2 YAML escaping via a variable-built replacement (the `\'` literal keeps the backslash on macOS system bash), `cd -P` tmpdir-symlink resolution, and `run_boot_hermetic` isolating HOME/PATH so a host tracebloc CLI can't pass the test for the wrong reason. Plus the clientId single-quote hardening. manifest.sha256 updated; CI fully green.

PR #451, opened 2026-07-28
fix(installer): execute-gate installed tools so a broken binary fails Step 1 (#411)

shujaatTracebloc:
Closes #411.

## Problem
The only post-install "verification" was a **log-only, failure-masking** interpolation (`Log "k3d: $(k3d version …)"` / `… 2>/dev/null || echo present`). So a corrupt or wrong-architecture binary — winget shims / partial installs skip the direct path's checksum verify, and brew delivers with no checksum of ours (#429) — still printed `✔ System tools` and only died at cluster-create in Step 2. Lukas's audit confirmed it's **all three OSes**.

## Fix (both halves + macOS)
- **Shared execute-gate helpers**: `assert_tool_runs` ([common.sh](scripts/lib/common.sh), bash) and `Assert-ToolRuns` ([install-k8s.ps1](scripts/install-k8s.ps1), PowerShell). Each **runs the tool's self-check**; on a non-zero exit or an exception it removes the located binary and fails loudly with an **arch-aware remedy**, so the tool step fails instead of the cluster step.
- Gate **kubectl / k3d / helm** after install on Linux, macOS, and Windows — on both the fresh-install and already-present paths. `System tools` success only prints once all gates pass.
- **kubectl is gated with `version --client`, not `--short`** (removed in kubectl 1.28+, which would false-fail the gate).
- New `check-drift.sh` parity check (`_drift_execute_gates`) so no installer can silently drop a gate.

## Acceptance criteria
- [x] A corrupt / wrong-arch `k3d.exe` fails Step 1 loudly with an actionable message; **Step 2 is never reached** (`Err`/`error` exits before it).
- [x] Coverage for the failure path — widened to **bats + Pester + the drift check**.

## Tests (+11)
- `common.bats` +2 (working tool passes & binary untouched / broken tool errors + removes the binary)
- `check-drift.bats` +2 (execute-gate parity)
- `install-k8s.Tests.ps1` +7 Pester (exit-nonzero / exception / binary-removal / arch remedy / static gates)

Verified locally (CI-identical):
- Pester full: **182 passed, 0 failed**, 8 skipped (Windows-only)
- all bats: 3 failures are pre-existing macOS-local flakes (my new tests all pass); ubuntu CI is the arbiter
- PSScriptAnalyzer: **0 Error-severity** · check-style / drift: clean · manifest regenerated

---

> [!NOTE]
> **Low Risk**
> Changes are confined to installer verification and drift tests; behavior is fail-fast with guarded binary removal, with no auth or runtime cluster logic touched.
> 
> **Overview**
> Replaces **log-only** post-install version checks (which could still show ✔ System tools when a binary was corrupt or wrong-arch) with **real execute gates** that run each tool’s version self-check and **fail Step 1** with an arch-aware message before cluster creation.
> 
> Adds **`assert_tool_runs`** in `common.sh` and **`Assert-ToolRuns`** in `install-k8s.ps1`, wired for **kubectl** (`version --client`), **k3d**, and **helm** (bare `version`, not `--short`) on **Linux**, **macOS**, and **Windows** for both fresh installs and already-on-PATH tools. Optional **`--rm` / `-BinPath`** removes only the installer-placed binary when it is the one that failed, so re-runs can self-heal without deleting winget/brew copies elsewhere on PATH.
> 
> Adds **`_drift_execute_gates`** in `check-drift.sh` plus bats/Pester coverage so no platform can drop a gate silently; updates `manifest.sha256` and Linux setup test mocks so gates don’t break install-path assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0f2626fb91ab4f713a7c8e10cfa8ffb3d102ddf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/common.sh
```diff
@@ -140,6 +140,27 @@ step_header()    { echo -e "  ${TB_HEADING}$1) $2${RESET}"; echo ""; }
...
+# would false-fail the gate). Usage: assert_tool_runs <name> <version-arg>...
+assert_tool_runs() {
+  local name="$1"; shift
+  local out
+  if out="$("$name" "$@" 2>&1)"; then
+    log "$name OK: $(printf '%s\n' "$out" | head -1)"
+    return 0
+  fi
+  local path; path="$(command -v "$name" 2>/dev/null || true)"
+  [[ -n "$path" && -w "$path" ]] && rm -f "$path" 2>/dev/null || true
```
saadqbal: This `rm` fires on the *already-present* path too (macOS gates all three unconditionally after `has`), so a broken brew-managed binary gets its symlink deleted — but `brew install` on re-run sees the formula already installed and won't relink, so the gate fails again → confusing loop. Note the Windows side deliberately omits `-BinPath` on the already-present calls (815) to avoid removing a tool it didn't install; the bash gate has no such guard. Consider only removing when we placed the binary (i.e. pass the dest path like the fresh-install callers, and skip removal otherwise).

scripts/lib/setup-macos.sh
```diff
@@ -253,20 +253,23 @@ install_docker_desktop() {
...
     spin_cmd "Installing system tools…" brew install k3d
   fi
-  log "k3d: $(k3d version | head -1 2>/dev/null || echo installed)"
+  assert_tool_runs k3d version
 
   if ! has helm; then
     spin_cmd "Installing system tools…" brew install helm
   fi
-  log "helm: $(helm version --short 2>/dev/null || echo installed)"
+  assert_tool_runs helm version --short
```
saadqbal: kubectl was moved off `--short` here because it was removed in 1.28+ and would false-fail the gate — but helm is still gated with `version --short` on all three OSes. Worth confirming the pinned Helm (v4.x) still supports `--short`; if it ever drops it the gate false-fails and deletes a perfectly good binary. `helm version --template` / bare `helm version` would be immune.

scripts/install-k8s.ps1
```diff
@@ -932,7 +962,7 @@ function Install-K3dAndHelm {
       RefreshPath
     }
   }
-  Log "k3d: $(k3d version | Select-Object -First 1)"
+  Assert-ToolRuns -Name "k3d" -VersionArgs @("version") -BinPath "$TOOL_DIR\k3d.exe"
```
saadqbal: This gate runs on the already-present path too, but always points `-BinPath` at `$TOOL_DIR\k3d.exe`. If k3d came from winget/choco (or was already on PATH), the broken binary is elsewhere, so Remove-Item is a no-op and `Has "k3d"` stays true on re-run → same failure every time. The remedy text does tell the user to uninstall the package-manager copy, so this is loud-but-stuck rather than silent; just flagging that the auto-remove can't actually self-heal that path.

scripts/tests/check-drift.sh
```diff
@@ -186,7 +186,34 @@ _drift_cli_contract() {
...
+    for t in kubectl k3d helm; do
+      grep -qE "assert_tool_runs $t\b" "$DRIFT_ROOT/$f" || \
+        _note "$f: no execute-gate for '$t' (assert_tool_runs $t ...) — #411"
+    done
+  done
+  local ps1="scripts/install-k8s.ps1"
+  if [[ ! -f "$DRIFT_ROOT/$ps1" ]]; then _note "$ps1 is missing"; fi
+  for t in kubectl k3d helm; do
+    grep -qE "Assert-ToolRuns -Name \"$t\"" "$DRIFT_ROOT/$ps1" || \
+      _note "$ps1: no execute-gate for '$t' (Assert-ToolRuns -Name \"$t\") — #411"
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 3dc866f. _drift_execute_gates no longer whole-file greps: it strips comment lines and matches the actual call (assert_tool_runs … <tool> version / Assert-ToolRuns -Name "<tool>"), so a commented-out or merely-mentioned gate no longer satisfies it. Added check-drift.bats cases for the --rm form and a commented-out gate. (Also avoided grep -q under pipefail — SIGPIPE would false-fail.)

scripts/install-k8s.ps1
```diff
@@ -969,7 +999,7 @@ function Install-K3dAndHelm {
 
     if (-not (Has "helm")) { Err "Helm could not be installed. Install manually from https://helm.sh/docs/intro/install/ and re-run." }
   }
-  Log "helm: $(cmd /c 'helm version --short 2>&1')"
+  Assert-ToolRuns -Name "helm" -VersionArgs @("version","--short") -BinPath "$TOOL_DIR\helm.exe"
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 3dc866f. -BinPath now tracks the direct-download dest only ($k3dDest/$helmDest, $null on the winget/present path), so the gate removes only a binary WE placed — never a winget/pre-existing copy elsewhere on PATH. kubectl already did this.

scripts/lib/common.sh
shujaatTracebloc: Fixed in 3dc866f. assert_tool_runs removal is now opt-in via a leading --rm <path>; only the Linux fresh-install callers pass it (the dest we placed). The already-present path and all macOS/brew calls pass no --rm, so a brew-managed binary is never deleted — no more wedged re-run. Added a bats case asserting no-removal without --rm.

scripts/lib/setup-macos.sh
shujaatTracebloc: Fixed in 3dc866f. helm is now gated with bare `helm version` on all three OSes (Linux/macOS/Windows) — immune if --short is ever dropped like kubectl's was.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 3dc866f. The k3d/helm PS gates now pass -BinPath $k3dDest/$helmDest, which is set only on the direct-download path and stays $null for winget/already-present — so Remove-Item no longer targets a TOOL_DIR path we didn't populate.

scripts/lib/setup-linux.sh
```diff
@@ -401,8 +401,11 @@ install_kubectl() {
       || error "Couldn't resolve the kubectl version from dl.k8s.io/release/stable.txt — check network connectivity to dl.k8s.io and re-run."
     spin_cmd "Installing system tools…" _fetch_kubectl "$KUBE_VER" "$ARCH_DL"
     log "kubectl $KUBE_VER installed."
+    # Fresh install: we placed it, so --rm lets a corrupt/wrong-arch download be
+    # cleared. Present path (else): don't remove a binary we didn't install (#411 review).
+    assert_tool_runs --rm "$TB_TOOLS_DIR/kubectl" kubectl version --client
   else
-    log "kubectl: $(kubectl version --client --short 2>/dev/null || echo present)"
+    assert_tool_runs kubectl version --client
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in f0f2626. All three tools now gate uniformly with --rm (bash) / -BinPath (ps1) pointing at our install dir on EVERY path, and the gate removes that path only when the binary that actually ran resolves to it (command -v … -ef on bash, (Get-Command).Source -eq on ps1). So a broken installer-placed kubectl/k3d — fresh OR left by a prior run — self-heals on re-run, while a pkg-manager/winget copy elsewhere on PATH is never deleted. Added a decoy-copy bats case and a ps1 resolved-elsewhere case.

saadqbal (approved):

PR #453, opened 2026-07-28
fix(installer): trust the corporate MITM CA in the k3d nodes (#424)

shujaatTracebloc:
Closes #424.

## Problem
Proxy *reachability* reaches the k3d nodes, but on a **TLS-inspecting (break-and-inspect)** network the nodes still don't *trust* the corporate CA — so every **in-node containerd pull** (rancher/k3s, ghcr.io, tracebloc images) fails x509. Because `helm upgrade` runs without `--wait` the failure is masked (companion #425) and surfaces as a root-cause-free *"an image couldn't be pulled."* This is the enterprise/hospital archetype, on all three OSes.

## Fix
**1. Inject the CA at create time (the core fix).** When `TRACEBLOC_CA_BUNDLE` (or `CURL_CA_BUNDLE`) is set, mount the bundle into every k3d node (`-v …:/etc/ssl/certs/tracebloc-mitm-ca.crt@all`) and write a `registries.yaml` (`configs.<registry>.tls.ca_file`) pointing containerd at it for `docker.io` / `registry-1.docker.io` / `ghcr.io`, hooked into the same `--config`/create path that already carries proxy env. Parity across `cluster.sh` (Linux/macOS) and `install-k8s.ps1` (Windows). A CA var set but **unreadable fails loudly**, not silently.

**2. Name the env var where the user hits the wall.** The TLS-interception preflight hint (both OSes), `docs/INSTALL.md`, and the PS `-Help` env-var list.

**3. CA-aware diagnosis.** `_diagnose_not_ready` / `Get-NotReadyState` detect `x509` / "certificate signed by unknown authority" pull events → a dedicated `image_pull_ca` state → *"the cluster does not trust your network's TLS-inspection CA"* + the exact remedy (never the generic pull error). Mirrored in `summary.sh` and `Print-Summary`.

**Drift gate.** New `check-drift.sh` parity check (`_drift_ca_trust`) so neither installer can drop the CA wiring for the other's OS.

## Design choices (as flagged)
- **Env var:** support both — prefer explicit `TRACEBLOC_CA_BUNDLE`, fall back to `CURL_CA_BUNDLE`.
- **Injection:** k3d `registries.yaml` `ca_file` as primary (targets the actual registry-pull failure) **plus** mounting the bundle into the nodes so the `ca_file` path resolves.

## Acceptance criteria
- [x] Behind a MITM proxy with the CA supplied: the nodes trust it for in-node pulls (k3d-native `registries.yaml` + node mount).
- [x] Without the CA: preflight **and** the diagnosis name the CA problem + the env var — never a generic pull error.

## Tests (+21)
`cluster.bats` +8 (resolver 4 cases, registries.yaml gen, create wires mount + `--registry-config`, no-CA path, unreadable→hard error), `summary.bats` +3 (x509→`image_pull_ca`, non-x509 stays `image_pull`, report names CA+env var), `check-drift.bats` +2, Pester +8.

Verified locally (CI-identical):
- Pester full: **183 passed, 0 failed**, 8 skipped (Windows-only)
- all bats: 3 failures are pre-existing macOS-local flakes (all 21 new tests pass); ubuntu CI is the arbiter
- PSScriptAnalyzer: **0 Error-severity** · check-style / drift: clean · manifest regenerated

*Note: a real break-and-inspect network can't be exercised in CI; the wiring is the documented k3d `registries.yaml` + node-mount mechanism and is unit-tested end-to-end.*

---

> [!NOTE]
> **Medium Risk**
> Changes cluster-create wiring and failure diagnosis across all installer paths; mistakes could affect enterprise installs or mis-route users, but scope is installer-only with strong unit/drift coverage and fail-closed CA handling.
> 
> **Overview**
> Fixes **TLS-inspecting (break-and-inspect)** installs where in-cluster **containerd** pulls failed **x509** and surfaced only as a generic image-pull error (#424).
> 
> When **`TRACEBLOC_CA_BUNDLE`** or **`CURL_CA_BUNDLE`** points at a readable PEM file, **cluster create** mounts it into every k3d node and passes a generated **`registries.yaml`** (`ca_file` for `docker.io`, `registry-1.docker.io`, `auth.docker.io`, `ghcr.io`). Unreadable or missing CA paths **fail loudly**; mounting without registry config is not allowed. Reusing an existing cluster **without** the CA mount triggers a **recreate** warning. Same behavior in **`cluster.sh`** and **`install-k8s.ps1`**, guarded by **`check-drift.sh`** `_drift_ca_trust`.
> 
> **Readiness/summary** adds **`image_pull_ca`**: x509 is detected only on pull-failure events, with remediation (delete cluster, re-run with the env var). Preflight TLS hints and **`docs/INSTALL.md`** document host vs node trust. Tests and **`manifest.sha256`** updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c56bffde73cca7628f706cb34214d17f2d47dda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/cluster.sh
```diff
@@ -111,6 +111,49 @@ _write_k3d_proxy_config() {
...
+# Echo the operator's CA bundle path (absolute) when TRACEBLOC_CA_BUNDLE or
+# CURL_CA_BUNDLE is set and readable. If a var is set but the file is unreadable,
+# echo the offending var NAME and return 2 — the caller turns that into a hard
+# error (a silent skip would drop them straight back into the x509 failure they
+# set the var to fix). Empty stdout + return 0 when no CA var is set.
+_resolve_ca_bundle() {
+  local var val
+  for var in TRACEBLOC_CA_BUNDLE CURL_CA_BUNDLE; do
+    val="${!var:-}"; [[ -z "$val" ]] && continue
+    if [[ ! -r "$val" ]]; then echo "$var"; return 2; fi
```
saadqbal: Non-blocking parity nit: this only checks `-r`, so a *directory* (some IT teams hand out a folder of PEMs) passes, then gets bind-mounted over the single file path and containerd can't read it as a `ca_file` — the exact silent "looks applied but still x509" case the rest of this PR kills. The PS `Resolve-CaBundle` already rejects it via `-PathType Leaf`; `[[ ! -r "$val" || ! -f "$val" ]]` here would match it.

saadqbal (dismissed):
Really careful PR — CA trust wired symmetrically across both installers, fail-loud on unreadable, a reuse-path guard, and a drift check to keep the two OS paths honest. Nicely tested too. One tiny parity nit inline, non-blocking. 👍

scripts/lib/cluster.sh
shujaatTracebloc: Good catch — fixed in c199c9f. `_resolve_ca_bundle` now requires a readable regular file: `[[ ! -r "$val" || ! -f "$val" ]]`, so a directory of PEMs returns the var name + rc 2 (hard error) instead of being bind-mounted over the single node ca_file path. Now matches PS `Resolve-CaBundle -PathType Leaf`. Added a directory-reject bats case.

scripts/lib/cluster.sh
```diff
@@ -510,6 +558,27 @@ _check_existing_cluster_proxy() {
...
+# exists WITHOUT it, a re-run reuses the cluster and the x509 pulls persist — so
+# the "set the CA and re-run" remedy silently does nothing. Warn and point at
+# recreate (Bugbot #424). The path mirrors _create_new_cluster's mount destination.
+_check_existing_cluster_ca() {
+  [[ -n "${TRACEBLOC_CA_BUNDLE:-}" || -n "${CURL_CA_BUNDLE:-}" ]] || return 0
+  local server_container="k3d-${CLUSTER_NAME}-server-0"
+  local mounts
+  mounts=$(docker inspect "$server_container" --format '{{range .Mounts}}{{println .Destination}}{{end}}' 2>/dev/null) || return 0
+  [[ -z "$mounts" ]] && return 0
+  if [[ "$mounts" != *"/etc/ssl/certs/tracebloc-mitm-ca.crt"* ]]; then
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in cb4bc78. `_check_existing_cluster_ca` now matches the mount destination as an exact whole line via `grep -qxF '/etc/ssl/certs/tracebloc-mitm-ca.crt'` instead of a substring test, so a longer path that merely embeds it (e.g. `…crt.bak`) is no longer treated as our CA mount and the recreate warning still fires. This mirrors the PS side, which already anchored with `(?m)^…\s\*$` (no PS change needed). Added a substring-embed bats case.

saadqbal (approved):

scripts/lib/cluster.sh
```diff
@@ -673,6 +745,29 @@ _create_new_cluster() {
     log "Propagating proxy settings to k3d nodes (authenticated proxies supported; NO_PROXY auto-augmented)."
   fi
 
+  # In-node CA trust for TLS-inspecting networks (#424): mount the operator's CA
```
saadqbal, line 748: Worth calling out: this + registries.yaml only fix containerd pulls *inside* the node. The k3d runtime images themselves (rancher/k3s, ghcr.io/k3d-io/k3d-tools, k3d-proxy) are pulled by the host Docker daemon, which this doesn't touch — so on a fresh machine behind break-and-inspect, `k3d cluster create` x509s at that host pull before the node ever boots, and that failure won't be classified as `image_pull_ca` (the diagnosis only runs post-create). On Linux, trusting the CA in the system store covers dockerd; on Docker Desktop (mac/Win) it doesn't reach the VM. Non-blocking, but the primary target scenario can still hit a raw x509 at create — worth at least a docs line.

scripts/lib/summary.sh
```diff
@@ -68,6 +68,14 @@ _diagnose_not_ready() {
   fi
   pods="$(kubectl get pods -n "$ns" --request-timeout=5s 2>/dev/null || true)"
   if printf '%s' "$pods" | grep -qiE 'ImagePullBackOff|ErrImagePull|InvalidImageName'; then
+    # On a TLS-inspecting network the pull fails x509 because the nodes don't trust
+    # the corporate CA (#424). Distinguish it from a generic pull error so the
+    # remedy can name the CA + the env var, not a vague "retry".
+    local events
+    events="$(kubectl get events -n "$ns" --request-timeout=5s 2>/dev/null || true)"
+    if printf '%s' "$events" | grep -qiE 'x509|certificate signed by unknown authority|tls: failed to verify'; then
```
saadqbal: Minor: this classifies `image_pull_ca` if *any* event in the namespace matches x509, not necessarily the pull that's actually failing. Low risk in a fresh install ns, but a stale/unrelated x509 event would steer someone into a delete+recreate for the wrong reason. Could scope it to the Failed pull event if you want to be strict.

saadqbal (dismissed):
Really careful PR — the CA resolver, fail-closed wiring, drift gate, and the bash/PS parity all hold up, and the test coverage is excellent. Approving. Two non-blocking notes inline (host-docker pull gap + the events-based CA classification).

scripts/lib/summary.sh
shujaatTracebloc: Fixed in 5c56bff. `_diagnose_not_ready` (and PS `Get-NotReadyState`) now filter events to the image-pull failure lines (`failed to pull` / `ErrImagePull`) before testing for x509, so a stale/unrelated x509 event elsewhere in the namespace no longer classifies as `image_pull_ca`. Added an unrelated-x509 test to both the bats and Pester suites.

scripts/lib/cluster.sh
shujaatTracebloc: Good call — agreed this is out of scope for #424 (which covers in-node containerd pulls). The host Docker daemon pulling k3d's own runtime images at create-time is a separate failure surface with a platform-split fix (Linux system store covers dockerd; Docker Desktop VM does not), so I filed it as a follow-up: #474. Keeping #453 focused on the in-node path.

saadqbal (approved):

PR #469, opened 2026-07-29
fix(installer): honest cosign bootstrap download + translated DISM failures

LukasWodka:
## Summary
The v1.9.7-rc.1 FR killed a **healthy** install: Windows PowerShell 5.1's progress overlay throttles `Invoke-WebRequest` massively and paints a "Writing request stream" banner, so the one big bootstrap download (pinned cosign, ~17 MB) ran ~4.5 minutes with zero output and got killed as frozen. This PR makes the bootstrap fast and visibly alive, fixes the `â€¦` mojibake at its actual root cause, and translates the raw DISM COMException on Server SKUs. **The security chain is unchanged** — same pins, same fail-closed paths, same verification order.

## What changed
1. **Progress-overlay silencing** in `Get-WithRetry` / `Get-Optional` (function-local `$ProgressPreference`, auto-reverts on return) — the classic 10–50× PS 5.1 IWR speedup.
2. **Liveness ticks for the cosign fetch**: `Get-OptionalWithTicks` runs the download in a background job (cwd pinned per #409, TLS 1.2 floor re-applied in the fresh process, progress silenced) while `Wait-JobWithTicks` prints a dim heartbeat dot every 3 s. Expectation lines before (`~17 MB … a few minutes on a slow network is normal` + the `winget install sigstore.cosign` tip), elapsed + `checksum-verified` confirmation after. The 30-min timeout only bounds a wedged transfer — a slow-but-moving download is never killed (that's the exact failure mode this fixes).
3. **Mojibake root cause**: it is *not* console output encoding — the release asset is served `application/octet-stream` with no charset, so PS 5.1's `irm` decodes the UTF-8 source as **Latin-1 before `iex`**; `-File` reads of BOM-less scripts are ANSI likewise. Fix: ASCII-only string literals in both installers (matches the existing `--` convention in install-k8s.ps1; glyphs stay as `[char]0xNNNN`). Locked in by a **tokenizer-based Pester test** over both files — which immediately caught one more real case (the `-Help` here-string em-dash).
4. **`Enable-OneVirtFeature`** (extracted from `Enable-VirtualisationFeatures`): catches DISM's raw COMException and names the real situation — feature package absent on this edition (Server SKUs have no `Microsoft-Hyper-V-All`) vs. enable failure (with a manual-remedy hint). Also a deliberate behavior fix: **$rebootNeeded is only set when a feature actually got enabled** — the old code demanded a reboot for a feature that never enabled, sending Server users into a reboot → re-run → same-error loop.

## Related
Closes #468

## Type of change
- [x] Bug fix

## Test plan
- `Invoke-Pester scripts/tests` — **212 passed / 0 failed** (pwsh 7.5.2, Pester 5.7.1), incl. new tests: `Wait-JobWithTicks` behavior (fast-finish + timeout-stop), download-UX invariants (progress silenced in both helpers; job re-applies TLS 1.2 + pins local cwd), ASCII string-literal hygiene for **both** installers, and 4 `Enable-OneVirtFeature` cases (already-enabled / newly-enabled / package-absent / enable-fails).
- `Invoke-ScriptAnalyzer` on both entrypoints: 0 errors.
- `bash scripts/check-style.sh`: clean. `scripts/gen-manifest.sh` regenerated (install-k8s.ps1 digest changed).
- End-to-end repro harness: `win-install-smoke.yml` in release-train runs the exact customer one-liner on windows-latest — will re-run it on the next rc containing this fix.

## Deployment notes
No env vars, no flags. Ships with the next client release; the released bootstrap is the stamped `install.ps1` release asset.

## Checklist
- [ ] Docs updated if behavior or config changed
- [x] For security-sensitive paths: appropriate reviewer requested (CODEOWNERS: @saadqbal)
- [x] Terminal output follows STYLE.md — tone helpers, no hardcoded colour; `bash scripts/check-style.sh` passes
- [x] Cross-repo issues use `Fixes tracebloc/<repo>#N` — same-repo `Closes #468` used here

---

> [!NOTE]
> **Low Risk**
> Installer UX and error-handling only; cosign/manifest verification order and fail-closed paths are unchanged per the PR.
> 
> **Overview**
> Fixes **healthy Windows installs being killed as “frozen”** during the bootstrap’s large cosign download and **Server SKU reboot loops** when optional features can’t be enabled.
> 
> **Bootstrap (`install.ps1`):** Silences PS 5.1’s `Invoke-WebRequest` progress overlay in download helpers, adds **`Get-OptionalWithTicks`** / **`Wait-JobWithTicks`** so the ~17 MB cosign fetch runs in a background job with periodic dots plus upfront timing guidance and post-download confirmation. Replaces Unicode punctuation in **user-visible string literals** with ASCII (`--`) so Latin-1 / ANSI decoding doesn’t show mojibake.
> 
> **K8s installer (`install-k8s.ps1`):** Introduces **`Enable-OneVirtFeature`** so missing packages (e.g. Hyper-V on Server) and enable failures surface as warnings instead of raw COMExceptions; **`$rebootNeeded` is set only when a feature actually enables**.
> 
> **Tests & manifest:** New Pester coverage for job ticks, download UX invariants, ASCII literal hygiene, and virt-feature cases; **`manifest.sha256`** digest updated for `install-k8s.ps1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c854391bd726fd02e8677064ce8cc37ad08b8cbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shujaatTracebloc (approved):
Reviewed code + logic + security chain against issue #468 — LGTM. ✅

**Delivers the full issue scope (#468)** — all four items plus one correctness fix:
1. Progress-overlay silencing in `Get-WithRetry` / `Get-Optional` (function-local, auto-reverts). ✓
2. Liveness: `Get-OptionalWithTicks` + `Wait-JobWithTicks` heartbeat, upfront ~17 MB expectation lines + the `winget install sigstore.cosign` tip, elapsed + checksum-verified confirmation. ✓
3. Mojibake fixed at the actual root cause — ASCII-only **string literals** (comments exempt, glyphs as `[char]0xNNNN`), which is more correct than the issue's `[Console]::OutputEncoding` suggestion since the corruption is the `irm`/`-File` Latin-1 decode, not console output. ✓
4. `Enable-OneVirtFeature` translates the raw DISM COMException (package-absent vs enable-fail) **and** only sets $rebootNeeded on an actual state change — kills the Server-SKU reboot→re-run→same-error loop. ✓

**Verified**
- Security chain unchanged: every hunk in `Resolve-Cosign` / `Confirm-ManifestSignature` / `Confirm-ScriptIntegrity` / `Invoke-Bootstrap` is an ASCII string swap or the fetch mechanism — pins, fail-closed `throw`s, checksum-before-use, and verification order all intact.
- `Get-OptionalWithTicks` is well-guarded: re-applies the TLS 1.2 floor in the fresh process, pins cwd (#409), drains the job's terminal error via `Receive-Job` in try/catch, times out only a wedged transfer, and confirms `Test-Path` on the output.
- `$edition` is defined (line 521) before the loop, so the not-available message renders correctly.
- `scripts/manifest.sha256` digest for install-k8s.ps1 recomputed and matches the branch (`d24bcac5…`).
- CI green (35 pass, 2 conditional skips); Cursor Bugbot clean on the latest commit `c854391`; branch current with develop; MERGEABLE.

**Non-blocking nit:** in the package-absent path `Enable-OneVirtFeature` prints `Enabling <label>...` before the `not available … skipping` line — slightly redundant sequence, cosmetic.

Note: this touches a security-sensitive path with **CODEOWNER @saadqbal** — my approval is additive; their review is still required to merge. Merge-order with #471 stands (whichever lands second re-runs gen-manifest.sh for the install-k8s.ps1 digest).

PR #476, opened 2026-07-29
fix(installer): failures surface the real error + log path + support-bundle hint (#423)

shujaatTracebloc:
## What & why
Fatal failures printed a generic red line (e.g. `✖ Failed to create compute environment.`) while the actionable detail — k3d's / helm's stderr, the real reason: image pull, proxy, port, WSL — went **only** to the transcript, and the transcript path itself was never shown on screen. Support then has to ask for a file nobody can find.

- **`Err` gains an optional `$Detail`**; new pure, unit-tested **`Get-ErrDetailLines`** renders — appended to *every* fatal error — the last ~5 non-empty lines of `$Detail`, the **log path**, and a **`-Diagnose`** support-bundle next step.
- **Cluster-create failure** passes k3d's stdout/stderr → the real reason shows on screen (the motivating case).
- **Helm** repo-add / reconcile / install failures pass helm's output via `$Detail` instead of embedding it in the message (removes duplicated log-path text).
- **Log path announced up front** in the banner (previously written only inside the log).

## Acceptance (from #423)
- [x] Every failure screen contains: real error excerpt, log path, next step — no generic-only failures
- [x] Log path is a first-class line at install start

## Tests
- Pester `Get-ErrDetailLines`: always names log path + `-Diagnose`; excerpt only when detail supplied; caps at last 5 non-empty lines; drops blanks; omits log line pre-log-init.
- Updated the `#385` helm-abort test to assert helm's output surfaces via the new `$Detail` path.

## Scope note
bash `install-k8s.sh` already dumps the real error to stderr on failure (`cat "$create_out" >&2`), so this Windows-specific gap doesn't exist there; keeping the change PS-only per the issue.

## Local verification
Pester 229/0/8 · check-style pass · drift clean · manifest regenerated.

Closes #423

---

> [!NOTE]
> **Low Risk**
> Installer UX and error formatting only on Windows; behavior is unit-tested and does not change install logic or credentials handling.
> 
> **Overview**
> Windows installer fatal errors now show **actionable context** instead of a generic red line only.
> 
> **`Err`** takes an optional **`$Detail`** and delegates to new **`Get-ErrDetailLines`**, which prints a `--- details ---` excerpt (last five non-empty lines, with PS 5.1 ErrorRecord chrome stripped), **`Full log:`** when logging is active, and **`Support bundle: re-run with -Diagnose`** on every failure. **k3d** create failures and **helm** repo-add / install / reconcile failures pass captured stdout/stderr through **`$Detail`**; duplicate inline log hints before **`Err`** are removed. The **install banner** announces the log path at start.
> 
> **Pester** covers **`Get-ErrDetailLines`** and the helm-abort test expects output via **`$Detail`**. **`manifest.sha256`** is updated for **`install-k8s.ps1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3f1639a0301bfc5f7412ebd4e527aaabb032634. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/install-k8s.ps1
```diff
@@ -54,7 +54,29 @@ if (-not $env:TB_PESTER) {
...
+}
+
+# $Detail (optional) is captured tool output (e.g. k3d/helm stderr); its last few
+# non-empty lines are surfaced on screen so the real reason isn't buried in the
+# log (#423). Every failure names the log path + -Diagnose regardless.
+function Err($m, $Detail)  {
+  Write-Host "  " -NoNewline; Write-Host ([char]0x2716) -ForegroundColor Red -NoNewline; Write-Host " $m" -ForegroundColor Red
+  foreach ($l in (Get-ErrDetailLines $Detail)) { Write-Host "  $l" -ForegroundColor DarkGray }
+  exit 1
+}
```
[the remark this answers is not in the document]
shujaatTracebloc: Addressed in f236cf0. I checked this empirically: when `Get-ErrDetailLines` returns a single line it does unwrap to a `System.String`, but PowerShell's `foreach` statement iterates a scalar string **once** (whole line), not per-character — I verified the exact bare `Err` loop prints `[Support bundle: re-run with -Diagnose]` as one line on pwsh 7, so the char-splitting doesn't reproduce. Still, I wrapped the enumeration in `@(...)` to make it unambiguous and future-proof, and added a regression test asserting the single-line case enumerates as one intact line (Count 1, length > 1).

scripts/install-k8s.ps1
```diff
@@ -54,7 +54,32 @@ if (-not $env:TB_PESTER) {
...
-function Err($m)           { Write-Host "  " -NoNewline; Write-Host ([char]0x2716) -ForegroundColor Red -NoNewline; Write-Host " $m" -ForegroundColor Red; exit 1 }
+# Build the trailing lines every fatal error shows (#423): a short excerpt of the
+# real tool output (last few non-empty lines — the actual reason, not a generic
+# line), then the log path and the -Diagnose support-bundle hint as first-class
+# next steps. Pure (no host writes / no exit) so it is unit-testable.
+function Get-ErrDetailLines([string]$Detail) {
+  $out = @()
+  if ($Detail) {
+    $lines = @($Detail -split "`r?`n" | ForEach-Object { $_.TrimEnd() } | Where-Object { $_ -ne "" } | Select-Object -Last 5)
+    if ($lines.Count) { $out += "--- details ---"; $out += $lines }
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in f5e4fa5 — and you'\''re right that this was a regression my change introduced (previously the full `$helmOutput` was dumped). `Get-ErrDetailLines` now strips the PS 5.1 ErrorRecord chrome (the `At <file>:<n> char:<n>` position line and the `+ …` / `+ CategoryInfo` / `+ FullyQualifiedErrorId` block) before taking the last-5 window, so the real `Error:` line survives instead of being crowded out by formatter noise. Added a regression test that feeds a simulated 5.1 `helm repo add` failure and asserts the `Error:` line is kept while the chrome is dropped.

scripts/install-k8s.ps1
```diff
@@ -1546,7 +1588,9 @@ function New-K3dCluster {
     if ($k3dStdout) { Log "k3d stdout: $k3dStdout" }
     if ($k3dStderr) { Log "k3d stderr: $k3dStderr" }
 
-    if ($k3dExitCode -ne 0) { Err "Failed to create compute environment." }
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. Removed the inline log-path hints from the k3d spawn-failure and create-timeout paths — Err now prints the log path (and -Diagnose) itself via Get-ErrDetailLines, so it's shown once. Added a guard test asserting no inline 'Full log:' hints remain in the installer.

scripts/install-k8s.ps1
```diff
@@ -1597,9 +1639,11 @@ function New-K3dCluster {
 
     if ($k3dExitCode -ne 0) {
       # Host-daemon x509 (k3d runtime image pull on a TLS-inspecting network,
-      # #474) -- name it before the generic failure.
+      # #474) -- name the CA remedy before the generic failure.
       Write-HostCaCreateHint -Output ("$k3dStdout`n$k3dStderr")
-      Err "Failed to create compute environment."
+      # Surface k3d's real reason (image pull / proxy / port / WSL) on screen via
+      # Err's detail excerpt, not only in the log (#423).
+      Err "Failed to create compute environment." "$k3dStderr`n$k3dStdout"
```
saadqbal: Nit: detail is `stderr\nstdout`, but `Get-ErrDetailLines` surfaces the *last* 5 lines — so if k3d writes anything to stdout, its tail crowds the stderr reason (FATA/x509/port) out of the excerpt. Putting stderr last (`"$k3dStdout`n$k3dStderr"`) would match the `Write-HostCaCreateHint` call just above and be safer. Low impact since k3d usually leaves stdout empty on failure — just a nit.

scripts/install-k8s.ps1
```diff
@@ -54,7 +54,44 @@ if (-not $env:TB_PESTER) {
...
+    # wraps around stderr (the `At <file>:<n> char:<n>` position line and the
+    # `+ ...` / `+ CategoryInfo` / `+ FullyQualifiedErrorId` block). Otherwise a
+    # helm failure's last 5 lines are all chrome and the real `Error:` line is
+    # crowded out of the excerpt (#423 Bugbot).
+    $lines = @($Detail -split "`r?`n" |
+      ForEach-Object { $_.TrimEnd() } |
+      Where-Object {
+        $_ -ne "" -and
+        $_ -notmatch '^\s*At [^ ]+:\d+ char:\d+' -and
+        $_ -notmatch '^\s*\+ '
```
saadqbal, line 74: Minor: `^\s*\+ ` strips any line starting with `+ `, not just PS ErrorRecord chrome — it'd also silently eat a real output line that happens to begin that way. Fine in practice for helm/k3d, just noting the breadth.

saadqbal (dismissed):
Clean, well-tested PR 👍 Two small nits inline, neither blocking.

scripts/install-k8s.ps1
saadqbal:
```suggestion
      Err "Failed to create compute environment." "$k3dStdout`n$k3dStderr"
```
stderr last, so its tail (the real FATA/x509/port reason) survives the last-5-line excerpt in `Get-ErrDetailLines`.

shujaatTracebloc: Fixed in b3f1639 — the create-failure Err detail is now `"$k3dStdout`n$k3dStderr"` (stderr last), so its tail (the FATA/x509/port reason) survives Get-ErrDetailLines' last-5-line window even when k3d wrote to stdout, and it matches the Write-HostCaCreateHint order just above.

shujaatTracebloc: Applied in b3f1639 — thanks, adopted the stderr-last ordering exactly as suggested.

scripts/install-k8s.ps1
shujaatTracebloc: Acknowledged — the breadth is deliberate: the whole PS 5.1 ErrorRecord block (the source-echo line, the `~~~` caret line, plus `+ CategoryInfo` / `+ FullyQualifiedErrorId`) all begin with `+ `, and stripping the block as a unit is what keeps the real `Error:` line in the excerpt. helm/k3d don't emit `+ `-prefixed real output, so there's nothing to lose in practice. Left as-is per your note.

saadqbal (approved):

PR #375, opened 2026-07-22
feat(install): Tier 0 — reuse an existing usable Docker (#1175)

LukasWodka:
## What (#1175 — Tier 0)

The biggest zero-effort unlock. When the host audit selects **Tier 0** (`docker info` OK — the user can already run a container), `install_linux` skips the **entire** privileged path:

- ❌ `preflight_sudo` (no password prompt)
- ❌ `install_docker_engine` (Docker already present)
- ❌ `setup_pm` / `apt_wait_for_lock` / `install_system_deps` (no package installs)
- ❌ `dispatch_gpu_setup` (privileged driver install — a Tier-0 GPU host already has its drivers; the k8s device-plugin still deploys later in `create_cluster`)
- ✅ only `install_kubectl` / `install_k3d` / `install_helm` into user-space, then `create_cluster` reuses the runtime.

A researcher in the `docker` group now installs with **no admin at all**.

The tool install is factored into `_install_userspace_tools`, shared with the full flow so the two can't drift.

## Tests

- 3 new `setup-linux.bats`: Tier 0 skips every privileged step (asserts none of `preflight_sudo`/`install_docker_engine`/`install_system_deps`/`dispatch_gpu_setup` run) and installs the tools; Tier 1 and unset-tier run the full flow.
- Full suite adds **no new failures**; shellcheck clean; R8 manifest current.

## Known edge (for real-host validation)

Tier 0 assumes runtime-level deps (e.g. `conntrack`) are present — true on essentially any host with a usable Docker, and we can't install them without privilege anyway. If one is missing, `create_cluster` surfaces it. Worth confirming on the real docker-group hosts in the spike's §5 matrix.

## Stack

Stacked on **#1172** (`feat/lpi-routing`) → **#1171** (`feat/lpi-foundation`). Base is `feat/lpi-routing`; merge #1171 → #1172 → this in order.

Part of the least-privilege install epic (#1168), Wave 2.

Closes tracebloc/backend#1175

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes core installer branching and where binaries land on PATH; mistakes could break docker-group installs or leave tools unreachable in new shells, but behavior is heavily covered by new bats tests.
> 
> **Overview**
[...]

PR #478, opened 2026-07-29
feat(install): Tier-1 k3d-on-rootless + cgroup delegation + autostart (#1221)

aptracebloc:
## Slice #1221 — Tier 1: k3d on the rootless socket + cgroup delegation + autostart

Part of #1177 (RFC 0001 least-privilege install) · Epic #1168. Builds on slice 1 (#1219 rootless daemon + `TB_TIER1_ROOTLESS` flag) and slice 2 (#1220 subuid/subgid gate + prepare-host).

**All new behavior is gated behind the opt-in `TB_TIER1_ROOTLESS=1` flag** (default off until the spike's §5 host-validation matrix runs). With the flag unset, every path below is a no-op and current behavior is byte-for-byte unchanged — proven by flag-off regression tests.

### Why
Slice 1 stands up the rootless daemon and exports `DOCKER_HOST`. That alone doesn't give a *usable* cluster: k3d must target that socket, and k3s inside the node needs delegated `cpu`/`cpuset`/`io` cgroup v2 controllers (a rootless user session doesn't get them by default), or the Helm chart's pod CPU/memory limits can't be enforced and the cluster breaks the moment workloads schedule. This is the load-bearing slice.

### What
**Core**
- **Shared marker** — `_rootless_active` predicate in `common.sh`; `cluster.sh` and `setup-linux.sh` both call it so the `INSTALL_TIER==1 && TB_TIER1_ROOTLESS==1` pair can't drift.
- **`create_cluster`** re-asserts `DOCKER_HOST=unix://$XDG_RUNTIME_DIR/docker.sock` when rootless is active, so every `k3d`/`docker` call targets the rootless daemon even if a caller (e2e harness, re-run) lost slice 1's export.
- **`ensure_cluster_autostart`** gets a Tier-1 branch: `systemctl --user enable docker` + `loginctl enable-linger` — never `sudo systemctl enable docker`. Promises reboot-survival (`TB_DOCKER_AUTOSTART`) only when **both** succeed (honesty rule, #375/#458).
- **cgroup delegation drop-in** (`Delegate=cpu cpuset io memory pids` under `[Service]`): a second named privileged touch, routed by privilege exactly like the subuid line — root/sudo → write `/etc/systemd/system/user@.service.d/delegate.conf` + `daemon-reload`; unprivileged → hand off to `prepare-host` (#1178) with the exact path+content. `run_prepare_host` writes it too (system-wide → covers the researcher).

**Carry-ins deferred from #452/#458**
- **Rootless proxy** — `_configure_docker_proxy` is now scope-aware (system+sudo vs `user`+no-sudo, one implementation) and called `user`-scoped in the rootless install, so a proxy-only host's rootless daemon can actually pull `rancher/k3s`.
- **`_set_tools_target`** installs user-space (`~/.local/bin`, no sudo) on rootless Tier 1 — fixes a `sudo mv → /usr/local/bin` crash that half-installed on a true no-sudo host *after* the daemon was up.
- **`DOCKER_HOST` persistence** — persisted to the shell rc so new terminals + the tracebloc CLI reach the rootless socket.

### Out of scope (per the issue)
- Flipping Tier 1 rootless on by default — blocked on the spike's §5 host-validation (§3 delegation + §4 fuse-overlayfs perf); stays opt-in here.
- fuse-overlayfs push/ingest perf, and the no-systemd/HPC fallback — later slices (#1222).

### Testing
- **14 new bats tests** across `cluster.bats` + `setup-linux.bats` (DOCKER_HOST wiring; autostart honesty both-succeed / enable-fails; delegation write + gate routing; user-scoped proxy; both carry-ins) with **flag-off regressions**.
- Verified locally on **Linux + non-root** (the faithful CI environment): `shellcheck --severity=error` clean, full suite green, no PATH-clobber. R8 manifest regenerated.
- **Real-host validation (spike §5.3) cannot run from a dev laptop** — the load-bearing outcome (all workloads `Ready` with limits *enforced* on the rootless socket, cluster returns after reboot) is validated on a runner with cgroup v2 + unprivileged userns, not in this PR. This is why the slice ships opt-in.

Draft until CI + Bugbot are green; will flip to Ready then.

---

> [!NOTE]
> **Medium Risk**
> Changes cluster creation and Linux install/bootstrap paths (Docker socket, systemd user units, privileged cgroup drop-in), but behavior stays behind an opt-in flag with flag-off regression tests; mis-wiring could still break installs or misstate reboot survival on headless hosts.
> 
> **Overview**
> **Opt-in Tier 1 rootless** (`TB_TIER1_ROOTLESS=1`) gets end-to-end wiring so k3d and new shells actually talk to the per-user Docker socket, with honest reboot messaging and cgroup setup for pod limits.
> 
> Introduces **`_rootless_active`** in `common.sh` so install and cluster code share one gate instead of duplicating tier + flag checks.
> 
> **Cluster:** `create_cluster` re-exports **`DOCKER_HOST`** to the rootless socket when active. **`ensure_cluster_autostart`** adds a rootless path (`systemctl --user enable docker` + **`loginctl enable-linger`**), ignores system **`docker.service`** “enabled” when seeding **`TB_DOCKER_AUTOSTART`**, and only promises reboot survival when **both** user enable and linger succeed.
> 
> **Install:** **`_configure_docker_proxy`** gains **`user`** scope (no sudo) and runs before the rootless daemon starts. **`_set_tools_target`** puts kubectl/k3d/helm in **`~/.local/bin`** without sudo on rootless Tier 1. **`_persist_docker_host`** writes **`DOCKER_HOST`** to the shell rc (without clobbering an existing user setting). **`_ensure_cgroup_delegation`** / **`_write_cgroup_delegation`** add the **`user@.service`** **`Delegate=cpu cpuset io memory pids`** drop-in (sudo or prepare-host handoff); **`run_prepare_host`** writes it too.
> 
> **Tests + manifest:** New bats coverage for DOCKER_HOST, autostart honesty, proxy user scope, persistence, and delegation; **`manifest.sha256`** updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93453b60f6d442f80f84121793c346d8ffcb35f5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/cluster.sh
```diff
@@ -487,6 +499,33 @@ ensure_cluster_autostart() {
...
+      # can't deliver — the honesty rule the legacy `elif sudo … enable` path
+      # already follows (only flags on success). #375/#458.
+      local _user_enabled=0 _linger_ok=0
+      if systemctl --user enable docker >/dev/null 2>&1; then _user_enabled=1; fi
+      if loginctl enable-linger "$(id -un 2>/dev/null || printf '%s' "${USER:-}")" >/dev/null 2>&1; then _linger_ok=1; fi
+      if [[ "$_user_enabled" == 1 && "$_linger_ok" == 1 ]]; then
+        TB_DOCKER_AUTOSTART=1
+        log "Tier 1 rootless: enabled the user Docker daemon on boot (systemctl --user enable + linger)."
+      else
+        log "Tier 1 rootless: boot autostart not fully enabled (user-service enable or linger unavailable); the --restart policy still applies while your user session is active."
```
saadqbal, line 539: The both-succeed gate here is undercut by the `is-enabled docker` seed at line 491. On a locked-down host where the *system* `docker.service` is enabled but the user can't use it (not in docker group, no sudo → falls to rootless Tier 1), `systemctl is-enabled docker` reads "enabled" and seeds `TB_DOCKER_AUTOSTART=1`. This else-branch then logs "not fully enabled" but never clears it, so a linger/user-enable failure still lets summary.sh promise reboot-survival — which the rootless cluster can't deliver (system docker starting doesn't bring up the user daemon). That's the exact honesty rule (#375/#458) this branch is enforcing. Reset `TB_DOCKER_AUTOSTART=0` in this else.

scripts/lib/setup-linux.sh
```diff
@@ -734,6 +762,32 @@ _persist_tools_on_path() {
...
+  if [ "$(basename "${SHELL:-sh}")" = "fish" ]; then
+    hint "Point new shells at the rootless Docker socket:  set -Ux DOCKER_HOST \"unix://\$XDG_RUNTIME_DIR/docker.sock\""
+    return 0
+  fi
+  local rc; rc="$(_tools_rc_for_shell)"
+  # Already referenced (a prior run, or the user's own line) → leave it. Match
+  # 'DOCKER_HOST=' (not '…=unix://'): the line we write quotes the value
+  # (DOCKER_HOST="unix://…), so a '…=unix://' probe would miss our own prior line
+  # and double-append on a re-run.
+  if [ -f "$rc" ] && grep -qF 'DOCKER_HOST=' "$rc" 2>/dev/null; then return 0; fi
```
saadqbal: `grep -qF 'DOCKER_HOST='` also matches a user's *own* pre-existing line (e.g. a remote/TCP `DOCKER_HOST`), so we silently skip persisting the rootless socket and new shells keep pointing at the wrong daemon — the "new terminals reach the rootless daemon" promise quietly fails. Fine to not clobber their line, but at least `hint` when we skip because a foreign DOCKER_HOST already exists.

saadqbal (commented):
Careful, well-tested slice 👍 — flag-off regressions and the both-succeed autostart gate are exactly right. Two things worth a look before merge (one real, one nit).

scripts/lib/cluster.sh
```diff
@@ -487,6 +499,33 @@ ensure_cluster_autostart() {
...
+      fi
```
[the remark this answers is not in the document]
aptracebloc: Fixed in `55d8383`. You're right — my earlier fix only gated the rootless *branch*; the false promise leaked in through the pre-branch `systemctl is-enabled docker` seed. That seed now runs only when `! _rootless_active`, so on the rootless path the user-scope `systemctl --user enable` + `loginctl enable-linger` are the sole authority for `TB_DOCKER_AUTOSTART` — a system unit that happens to be enabled no longer seeds a promise the cluster (which runs on the rootless socket) can't keep. Pinned by a new test: system `docker.service` enabled + user-enable fails ⇒ flag stays 0.

scripts/lib/setup-linux.sh
```diff
@@ -832,7 +887,8 @@ install_rootless_docker() {
 
   # Point every later docker/k3d call in this run at the rootless socket. docker and
   # k3d both read DOCKER_HOST from the environment, so exporting it is sufficient
-  # here; explicit create_cluster wiring is slice 3 (#1221). Guard XDG_RUNTIME_DIR,
+  # here; create_cluster re-asserts this same export (#1221) so a re-run that lost
+  # it still targets the socket. Guard XDG_RUNTIME_DIR,
```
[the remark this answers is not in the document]
aptracebloc: Fixed in `55d8383`. Correct — this PR made `_ensure_cgroup_delegation` a second setter of `TB_ROOTLESS_ADMIN_TOUCH`, so the success line's hardcoded "subuid/subgid range" could misname the actual touch. It now reads "host prerequisites (subuid/subgid range and/or cgroup delegation)" — honest whether the sudo step was the subid range, the delegation drop-in, or both.

scripts/lib/cluster.sh
aptracebloc: Fixed in `93453b6` — the core was already addressed in `55d8383` (pushed just after your review's base `8b6f015`), and this adds the belt-and-suspenders you asked for. Two layers now:

1. The `is-enabled docker` seed is guarded with `! _rootless_active`, so on the rootless path the system unit's on-boot state never seeds `TB_DOCKER_AUTOSTART` — the user-scope `systemctl --user enable` + `loginctl enable-linger` are the sole authority.
2. Per your suggestion, the else-branch now also explicitly resets `TB_DOCKER_AUTOSTART=0`, so the honesty guarantee is local to the branch rather than relying on the seed-guard above it.

New test pins it: system `docker.service` enabled + user-enable fails ⇒ flag stays 0.

scripts/lib/setup-linux.sh
aptracebloc: Fixed in `93453b6`. You're right — the idempotency probe was too broad. It's now keyed off our own marker comment (so a re-run is still idempotent and never double-appends), and a *foreign* `DOCKER_HOST` is left untouched but we `warn` that new terminals and the CLI won't reach the rootless daemon until it's repointed at `unix://$XDG_RUNTIME_DIR/docker.sock` — no more silent skip. Bugbot flagged the same thing independently; a test now covers the foreign-value case (warns, no clobber, no double-write).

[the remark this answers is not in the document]
aptracebloc: Fixed in `93453b6`. `_persist_docker_host` no longer treats any `DOCKER_HOST=` line as "already configured": idempotency is keyed off our own marker comment, and a pre-existing foreign value (e.g. a TCP/remote daemon) is left in place with a `warn` to repoint it at `unix://$XDG_RUNTIME_DIR/docker.sock` — never a silent skip that leaves new shells on the wrong daemon. Same issue Asad raised in review; a test now covers it (foreign value ⇒ warns, no clobber, no double-write).

saadqbal (approved):

saqlainsyed007 (approved):

PR #483, opened 2026-07-30
fix(installer): report host RAM consistently + achievable memory advice (#417)

shujaatTracebloc:
## What & why
The Windows preflight memory check preferred Docker's WSL2 VM budget over physical RAM, so the **same 15 GB laptop** reported `Memory: 7 GB` with Docker up and `Memory: 15 GB ✓` with it down — flip-flopping across re-runs mid-install — and told a 15 GB machine to *"give Docker >= 16 GB"*, physically impossible advice. Seen on three machines (7/6/15 GB readings; the 16-on-15 hint).

- **`Get-PfMemGb` now returns HOST RAM only** (physical, via CIM) — identical whether Docker is up or down. The container runtime's smaller VM budget is read separately (`Get-PfRuntimeMemGb`) and shown as its own labeled line: `Memory: 15 GB (Docker's current share: 7 GB)`.
- **`Get-PfMemRecommendation`** caps every suggestion at **(host − 2 GB)** — never more than the machine physically has; floors at 1 GB.
- **Step-1 (`Test-Preflight`) and Step-2 (`Test-PreflightRuntimeMem`)** now give one consistent, host-aware message, and Step-2's recommendation is capped too.

## Acceptance (from #417)
- [x] Always two labeled lines: host RAM + Docker's current budget (when a runtime is up)
- [x] Cap every recommendation at (physical − 2 GB); never recommend more than the machine has
- [x] De-duplicate the Step-1/Step-2 warnings into one consistent message
- [x] Same machine, Docker up or down, reports identical host RAM; recommendations always achievable

## Tests
- `Get-PfMemRecommendation`: cap at host−2, returns desired when it fits, floors at 1, the 16-on-15→13 case.
- `Get-PfMemGb` reports host RAM regardless of the Docker budget (Windows CIM-mocked =16; cross-platform decoupling from `Get-PfRuntimeMemGb`).
- `Test-PreflightRuntimeMem` caps its recommendation at host RAM (4 GB on a 6 GB host, never 8).
- Updated the former "Get-PfMemGb prefers docker" test — it asserted the flip-flop bug.

## Local verification
Pester 261/0/9 · check-style pass · drift clean · manifest regenerated · PS parse OK. (No bash changes — bats unaffected.)

Closes #417

---

> [!NOTE]
> **Low Risk**
> Warn-only preflight UX and installer messaging in install-k8s.ps1; no auth, cluster, or data-path changes.
> 
> **Overview**
> Fixes Windows preflight memory reporting where **host RAM flipped** between Docker up/down and advice could exceed physical RAM (e.g. “give Docker 16 GB” on a 15 GB machine).
> 
> **`Get-PfMemGb`** now always returns **physical host RAM** (CIM only); Docker’s VM budget stays on **`Get-PfRuntimeMemGb`**. **`Show-MemoryStatus`** centralizes warn/ok text: the **label** is always host GB (plus optional “Docker’s current share”), while **pass/fail** uses the **effective** memory (Docker budget when known, else host) so a throttled Docker cap is never green-OK on a large host. **`Get-PfMemRecommendation`** caps hints at **host − 2 GB** (min 1 GB); when host RAM is unknown, hints use raw targets instead of capping at the current budget (#483).
> 
> **`Test-Preflight`** and **`Test-PreflightRuntimeMem`** both call **`Show-MemoryStatus`** so Step 1 and post-Docker checks share the same wording. Pester coverage and **`manifest.sha256`** updated for **`install-k8s.ps1`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dcc5f8ee129560150156810e5caacec495aa5507. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/install-k8s.ps1
```diff
@@ -3023,21 +3036,31 @@ function Test-Preflight {
...
+  } else {
+    $budgetNote = if ($null -ne $budget) { " (Docker's current share: $budget GB)" } else { "" }
+    $recTrain   = Get-PfMemRecommendation -DesiredGb $recMemGb  -HostGb $mem
+    if ($mem -lt $minMemGb) {
+      Warn "Memory: $mem GB$budgetNote - below the $minMemGb GB the client needs; it will OOM."
+      Hint "This machine has $mem GB of RAM total; the client needs at least $minMemGb GB. Free up memory or use a larger machine."
+    }
+    elseif ($mem -lt $warnMemGb) {
+      Warn "Memory: $mem GB$budgetNote - enough to run, but training (~8 GB/job) may OOM; more RAM helps to train locally."
+      Hint "To train locally, give Docker up to $recTrain GB (the most this $mem GB host can spare): WSL2 backend - [wsl2] memory=${recTrain}GB in %UserProfile%\.wslconfig + 'wsl --shutdown'; Hyper-V - Docker Desktop -> Settings -> Resources -> Advanced."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 2ba4437. The 5-7 GB (below-training-threshold) branch no longer tells the host to give Docker `host-2` GB 'to train locally' — that amount can't train (~8 GB/job). It now states the truth: the client runs fine here, but local training needs a bigger machine (~`warnMemGb+2` GB+ total). The unachievable `recTrain` target is gone; capable hosts get the actionable budget guidance from the Step-2 runtime check (which knows Docker's actual budget).

scripts/install-k8s.ps1
```diff
@@ -3118,13 +3141,18 @@ function Test-Preflight {
...
-    Hint "Give Docker >= $warnMemGb GB, then re-install: WSL2 backend - [wsl2] memory=${warnMemGb}GB in %UserProfile%\.wslconfig + 'wsl --shutdown'; Hyper-V backend - Docker Desktop -> Settings -> Resources -> Advanced."
+  if ($budget -lt $warnMemGb) {
+    # Cap the recommendation at what the host can spare (host - 2 GB) so we never
+    # advise more than the machine physically has (#417). Falls back to the raw
+    # target if host RAM can't be read.
+    $host_ = Get-PfMemGb
+    $rec = if ($null -ne $host_) { Get-PfMemRecommendation -DesiredGb $warnMemGb -HostGb $host_ } else { $warnMemGb }
+    $hostNote = if ($null -ne $host_) { " of $host_ GB host RAM" } else { "" }
+    Warn "Docker's memory budget is $budget GB$hostNote - the client may OOM under load; $rec GB recommended."
+    Hint "Raise Docker's memory to $rec GB, then re-install: WSL2 backend - [wsl2] memory=${rec}GB in %UserProfile%\.wslconfig + 'wsl --shutdown'; Hyper-V - Docker Desktop -> Settings -> Resources -> Advanced."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 2ba4437. Test-PreflightRuntimeMem now only says 'Raise Docker's memory to N' when N is actually greater than the current budget. When the capped recommendation is <= the current budget (the host is already at its achievable cap, e.g. 4 GB on a 6 GB host), it drops the no-op 'raise to 4' and instead names the real fix: use a machine with more RAM. Added a regression test for that case.

scripts/install-k8s.ps1
```diff
@@ -3023,21 +3036,33 @@ function Test-Preflight {
...
+    }
+    elseif ($mem -lt $warnMemGb) {
+      # Runs, but too small for local training: it can't spare a trainable Docker
+      # budget (~8 GB/job + ~2 GB for the OS), so don't dangle a "give Docker N GB
+      # to train" target that this host physically can't meet (#417 Bugbot).
+      Warn "Memory: $mem GB$budgetNote - enough to run the client, but too little for local training (~8 GB/job may OOM)."
+      Hint "The client will run here; for local training use a machine with more RAM (~$($warnMemGb + 2) GB+ total)."
+    }
+    else {
+      Ok "Memory: $mem GB$budgetNote"
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. Step-1 now warns for hosts below `warnMemGb + 2` (not just `warnMemGb`): sparing an 8 GB Docker budget needs ~2 GB more for the OS (matching the cap in Get-PfMemRecommendation), so an 8-9 GB host can't actually give Docker 8 GB. It's now flagged 'too little for local training' at Step-1 — consistent with Step-2's budget check — instead of a green Ok the runtime check then contradicts. Added a test asserting a 9 GB host is flagged rather than shown Ok.

scripts/tests/install-k8s.Tests.ps1
```diff
@@ -1073,9 +1087,13 @@ Describe "Get-PfFsType" -Skip:(-not $IsWindows) {
...
-  It "Get-PfMemGb prefers docker MemTotal over the host" {
-    Mock docker { '8589934592' }          # 8 GiB, in bytes
-    Get-PfMemGb | Should -Be 8
+  It "Get-PfMemGb no longer follows the Docker VM budget (#417 no flip-flop)" {
+    # The docker reader still sees the VM budget, but the reported memory figure is
+    # host RAM (or $null off-Windows where CIM is unavailable) — never the smaller
+    # budget. Cross-platform: proves the two are decoupled without needing CIM.
+    Mock docker { '8589934592' }          # Docker reports 8 GiB
+    Get-PfRuntimeMemGb | Should -Be 8     # the runtime reader follows docker...
+    Get-PfMemGb        | Should -Not -Be 8 # ...but the reported memory does not
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. Replaced the flaky `Should -Not -Be 8` (which would fail on a real 8 GB Windows host) with a host-independent assertion that `Get-PfMemGb` never invokes `docker` at all (`Should -Invoke docker -Times 0`) — that's the actual decoupling guarantee. Split out a separate positive test that `Get-PfRuntimeMemGb` still follows the docker budget. The exact host figure remains locked by the Windows-gated CIM-mocked sibling test.

scripts/install-k8s.ps1
```diff
@@ -3023,21 +3036,35 @@ function Test-Preflight {
...
+      # Runs, but too small for local training: to spare a trainable Docker budget
+      # (~warnMemGb for ~8 GB/job) the host also needs ~2 GB for the OS, so the
+      # real floor is warnMemGb + 2 (an 8-9 GB host can't give Docker 8 GB). Warn
+      # here so Step-1 agrees with Step-2 instead of a green Ok the budget check
+      # then contradicts, and don't dangle a target the host can't meet (#417 Bugbot).
+      Warn "Memory: $mem GB$budgetNote - enough to run the client, but too little for local training (~8 GB/job may OOM)."
+      Hint "The client will run here; for local training use a machine with more RAM (~$($warnMemGb + 2) GB+ total)."
+    }
+    else {
+      Ok "Memory: $mem GB$budgetNote"
```
saadqbal: This is the one that worries me. Step-1 now grades the host and demotes the budget to a decorative string, so a throttled Docker budget lands here on the green `Ok`.

I stubbed the CIM/docker primitives and ran Step-1 on both branches (host / docker budget -> memory line):

| host / budget | develop | this PR |
|---|---|---|
| 32 / 2 | `below the 5 GB the client needs; it will OOM` | `Memory: 32 GB (Docker's current share: 2 GB)` |
| 16 / 4 | `below the 5 GB ... it will OOM` | green Ok |
| 15 / 7 | `training (~8 GB/job) may OOM; 16 GB recommended` | green Ok |
| 10 / 5 | `training ... may OOM` | green Ok |

The 15/7 row is the machine from #417 — it used to warn at Step-1 and now gets a tick. 32/2 and 16/4 are budgets under the 5 GB floor reported as all-clear.

The comment at 3056-3060 says the `+2` rung exists so Step-1 agrees with Step-2 "instead of a green Ok the budget check then contradicts", but that only closes the 8-9 GB *host* case. Any host >= 10 GB with a budget < 8 still gets green-at-Step-1 then warned at Step-2 — no host threshold can fix that while Step-1 grades host and Step-2 grades budget.

Suggest grading the *effective* figure when a budget is available (budget when non-null, else host) and keeping host RAM as the context label. That keeps the #417 win — the reported number stops flip-flopping — without dropping the gate.

scripts/install-k8s.ps1
```diff
@@ -3118,13 +3145,24 @@ function Test-Preflight {
...
-  $mem = Get-PfRuntimeMemGb
-  if ($null -eq $mem) { return }
+  $budget = Get-PfRuntimeMemGb
+  if ($null -eq $budget) { return }
   $warnMemGb = if ($env:PF_WARN_MEM_GB) { [int]$env:PF_WARN_MEM_GB } else { 8 }
-  $recMemGb  = if ($env:PF_REC_MEM_GB)  { [int]$env:PF_REC_MEM_GB }  else { 16 }
-  if ($mem -lt $warnMemGb) {
-    Warn "Docker is running with $mem GB - recommended >= $warnMemGb GB ($recMemGb GB to train); the client may OOM under load."
-    Hint "Give Docker >= $warnMemGb GB, then re-install: WSL2 backend - [wsl2] memory=${warnMemGb}GB in %UserProfile%\.wslconfig + 'wsl --shutdown'; Hyper-V backend - Docker Desktop -> Settings -> Resources -> Advanced."
+  if ($budget -lt $warnMemGb) {
```
saadqbal: Related to 3065: this only ever compares the budget to `$warnMemGb`, so the `$minMemGb` "it will OOM" floor no longer applies to the Docker budget anywhere. A 2 GB budget and a 7 GB budget produce identical "may OOM under load" wording, and `PF_MIN_MEM_GB` now has no effect on the budget path on Windows at all.

Also worth noting this function is only reached from `New-K3dCluster` (1590). If anything in the Docker-install/tool-download stretch hits `Err` (which exits), the user never sees any memory guidance — where develop surfaced it at Step-1 before any download.

scripts/install-k8s.ps1
```diff
@@ -2950,13 +2950,26 @@ function Get-PfRuntimeCpu {
   return $null
 }
 
-# Prefer the runtime view, fall back to the host (CIM).
+# Total physical HOST RAM in GB — the consistent memory figure we report, whether
+# or not Docker is up (#417). The container runtime's smaller VM budget is read
+# separately via Get-PfRuntimeMemGb and shown as its own labeled line, so the
+# reported host RAM never flip-flops across re-runs. $null if undeterminable.
```
saadqbal, line 2956: Dropping the fallback outright costs one case worth keeping: when CIM is blocked (locked-down or policy-managed machine) but Docker *is* readable, Step-1 now prints `Memory: couldn't determine total RAM (skipping).` and throws away the budget it successfully read. On develop the same state printed `Memory: 4 GB - below the 5 GB the client needs; it will OOM.` plus the `.wslconfig` recipe. Confirmed on both branches with `Get-CimInstance` stubbed to throw.

A `$null`-host path that still reports the budget (clearly labelled as Docker's share) would keep that.

scripts/install-k8s.ps1
```diff
@@ -3023,21 +3036,35 @@ function Test-Preflight {
...
+  $budget = Get-PfRuntimeMemGb     # Docker VM budget, or $null when the daemon is down
+  if ($null -eq $mem) {
+    Warn "Memory: couldn't determine total RAM (skipping)."
+  } else {
+    $budgetNote = if ($null -ne $budget) { " (Docker's current share: $budget GB)" } else { "" }
+    if ($mem -lt $minMemGb) {
+      Warn "Memory: $mem GB$budgetNote - below the $minMemGb GB the client needs; it will OOM."
+      Hint "This machine has $mem GB of RAM total; the client needs at least $minMemGb GB. Free up memory or use a larger machine."
+    }
+    elseif ($mem -lt ($warnMemGb + 2)) {
```
saadqbal: `$warnMemGb + 2` quietly redefines a shared knob: with the default 8, a 9 GB host warns here but gets a green Ok from `preflight.sh`, which compares against the raw `warn_mib` (:250). Same env var, two meanings depending on OS, recorded nowhere but this inline comment.

Separately, the 2 GB OS reserve is now a bare literal in three places — `$HostGb - 2` at 2967, and `+ 2` here and at 3062. Change it in one spot and Step-1 goes back to green-OK-ing hosts that Step-2 warns about, i.e. exactly the bug this rung was added to fix. A single `$pfOsReserveGb` (or deriving the rung from `Get-PfMemRecommendation`) would make the coupling structural.

scripts/install-k8s.ps1
```diff
@@ -3023,21 +3036,35 @@ function Test-Preflight {
...
+      # Runs, but too small for local training: to spare a trainable Docker budget
+      # (~warnMemGb for ~8 GB/job) the host also needs ~2 GB for the OS, so the
+      # real floor is warnMemGb + 2 (an 8-9 GB host can't give Docker 8 GB). Warn
+      # here so Step-1 agrees with Step-2 instead of a green Ok the budget check
+      # then contradicts, and don't dangle a target the host can't meet (#417 Bugbot).
+      Warn "Memory: $mem GB$budgetNote - enough to run the client, but too little for local training (~8 GB/job may OOM)."
+      Hint "The client will run here; for local training use a machine with more RAM (~$($warnMemGb + 2) GB+ total)."
```
saadqbal: This drops the local-training target from 16 GB to 10. `$recMemGb` / `PF_REC_MEM_GB` (3001) is now assigned and never read anywhere in the file — `grep -c recMemGb` returns 1 — while `preflight.sh` still honours it at :28/243/248/251/252/282. So `PF_REC_MEM_GB=32` steers the advice on macOS/Linux and is a silent no-op on Windows.

And on the substance: someone who acts on "~10 GB+ total" for a ~8 GB/job workload will still OOM. Either re-thread `$recMemGb` through this hint or delete the dead assignment, but 10 GB looks like the wrong number to hand out.

saadqbal (commented):
Good fix for the flip-flop itself — the reported number is stable now and the cap logic is right. But I think Step-1 gave up too much: by grading host RAM only, a throttled Docker budget now gets a green tick where develop warned, including budgets below the 5 GB floor. Details inline at 3065 — I ran the ladder on both branches to confirm rather than eyeball it.

Two parity things I'd rather not lose, not blockers:

- `Get-PfCpu` (2974) still has the exact "prefer runtime, fall back to host" line you deleted from `Get-PfMemGb`, so the CPU row still flip-flops between Docker up and down. Same bug, same machine, and there's no `(Docker's current share: N)` label on that row to soften it.
- `preflight.sh` is untouched, so on macOS/Linux #417 still reproduces as filed: `_pf_total_mem_kb` (:157) prefers the runtime view, and :248 still says "raise to >= 8 GB (16 GB to train)" with no host cap. Closing #417 on the Windows fix alone leaves the reported repro live on the other platforms — worth either porting the policy or narrowing the issue to Windows.

Also worth a note: `check-drift.sh` gates ps1<->sh parity for API hosts and download hosts but nothing for preflight resource policy, which is why "drift clean" passed on a change that split the two installers' memory semantics.

Pester 264/0/9 green locally and the manifest hash matches the new file.

scripts/install-k8s.ps1
shujaatTracebloc: Adopted your suggestion in 2b2cb5a. Show-MemoryStatus now grades the **effective** figure (Docker's budget when non-null, else host) and keeps host RAM as the reported label. Your table now: 32/2 -> 'Memory: 32 GB (Docker's current share: 2 GB) - below the 5 GB the client needs; it will OOM'; 16/4 -> OOM; 15/7 -> 'training may OOM; 13 GB recommended'; 10/5 -> training warn. The reported number stops flip-flopping (host RAM) without dropping the gate. Docker-down at Step-1 still grades host (budget unknowable then), and Step-2 re-runs the same assessment on the budget.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 2b2cb5a. Test-PreflightRuntimeMem now calls the shared Show-MemoryStatus, so **both** floors apply to the budget — a 2 GB budget hits the minMemGb 'it will OOM' floor (PF_MIN_MEM_GB honored again), distinct from a 7 GB budget's 'training may OOM'. And because Step-1 now grades the effective figure too, host-level problems still surface at Step-1 before any download; only the budget (unknowable until Docker runs) is deferred to Step-2, which is unavoidable.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 2b2cb5a. Show-MemoryStatus has a host-null path: when CIM is blocked but Docker's budget is readable, it reports 'Memory: 4 GB (Docker's share; host RAM unreadable) - below the 5 GB the client needs; it will OOM' with the .wslconfig recipe, instead of 'couldn't determine ... skipping'. Confirmed with the host reader stubbed to $null.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 2b2cb5a. The warnMemGb+2 rung is gone entirely — grading the effective figure supersedes it, so PF_WARN_MEM_GB no longer means two things by OS (it's compared the same way as bash's warn_mib). And the OS reserve is now a single `$script:PfOsReserveGb` constant used by Get-PfMemRecommendation, not a bare literal in three places.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 2b2cb5a. recMemGb / PF_REC_MEM_GB is read again (in Show-MemoryStatus) and threaded into the training target, capped at host - OS reserve — so a 15 GB host is told '13 GB recommended' (not the impossible 16, and not the wrong 10). PF_REC_MEM_GB=32 now steers it on Windows too. No dead assignment, and no bogus 10 GB number.

scripts/install-k8s.ps1
```diff
@@ -2950,13 +2950,80 @@ function Get-PfRuntimeCpu {
...
+  $capHost  = if ($null -ne $HostGb) { $HostGb } else { $BudgetGb }
+  $recTrain = Get-PfMemRecommendation -DesiredGb $recMemGb -HostGb $capHost
+  # A throttled Docker budget is fixed at the daemon; a small host needs more RAM.
+  $budgetIsBottleneck = ($null -ne $BudgetGb) -and ($null -eq $HostGb -or $BudgetGb -lt $HostGb)
+
+  if ($effective -lt $minMemGb) {
+    Warn "Memory: $label$budgetNote - below the $minMemGb GB the client needs; it will OOM."
+    if ($budgetIsBottleneck) {
+      $recRun = Get-PfMemRecommendation -DesiredGb $warnMemGb -HostGb $capHost
+      Hint "Give Docker at least $minMemGb GB (up to $recRun GB): WSL2 backend - [wsl2] memory=${recRun}GB in %UserProfile%\.wslconfig + 'wsl --shutdown'; Hyper-V - Docker Desktop -> Settings -> Resources -> Advanced."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. When host RAM is unreadable there's no ceiling to cap against (the budget is the current throttled value, not the max), so I no longer fall $capHost back to $BudgetGb. The host-unknown path now advises the raw targets — 'Give Docker at least 5 GB (up to 8 GB)' — instead of the backwards 'up to 2 GB'. Added a regression test.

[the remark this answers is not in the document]
shujaatTracebloc: Same fix (latest push): the contradictory 'at least 5 GB (up to 2 GB)' came from capping the recommendation at the throttled budget when host RAM was unknown. Now it only caps at the host when the host is known; otherwise it uses the raw targets, so the min and 'up to' bounds are consistent (5 / 8).

saadqbal (approved):

PR #484, opened 2026-07-30
fix(installer): WSL update survives Store-blocked networks + skips when current (#414)

shujaatTracebloc:
## What & why
The installer ran `wsl --update` through the **Microsoft Store** with a 90s silent-timeout job. On Store-blocked corporate networks it **silently skipped** — Docker Desktop then confronted the user with its own install-WSL dialog + reboot (the root of the "every user installs WSL themselves" experience). It also **re-ran up to 90s on every run** even when the kernel was current, and its output went only to the log.

New **`Update-Wsl`** (replaces the inline Store-path job):
- **Skip when current** — `Test-WslCurrent` parses `wsl --version`; a re-run on an up-to-date machine finishes in **<2s**.
- **`wsl --update --web-download`** — fetches from Microsoft's servers instead of the Store, so a Store-blocked machine still updates the kernel with **no Docker Desktop WSL prompt**. Runs as a killable tracked process with a deadline.
- **Surfaces failure + the exact manual MSI step on screen** (github.com/microsoft/WSL/releases), not swallowed to the log.

## Acceptance (from #414)
- [x] Skip when current: parse `wsl --version` before updating
- [x] Prefer `wsl --update --web-download` (bypasses the Store)
- [x] Surface failure + exact next step on screen
- [x] Store-blocked machine: kernel updated with no Docker Desktop WSL prompt (via `--web-download`)
- [x] Machine with current WSL: the block completes in <2s on re-run

## Scope note — automated GitHub-releases MSI fallback
The issue also suggested auto-falling-back to the WSL MSI from GitHub releases. That is **not** implemented automatically: resolving the WSL release asset URL needs `api.github.com` (the asset name carries a 4th version component the API-free `/releases/latest` redirect can't resolve), and **#410 — enforced by a test — forbids the rate-limited GitHub API** in this installer. The manual MSI step is surfaced clearly instead, and a test guards against a regression that re-adds the API. Happy to wire an automated fallback if we decide to relax #410 for this rare path.

## Tests
`Test-WslCurrent` parsing (current / absent / legacy); source guards for `--web-download`, skip-when-current, removal of the bare Store-path job, the on-screen manual step, and the #410 no-API invariant.

## Local verification
Pester 264/0/8 · check-style pass · drift clean · manifest regenerated · PS parse OK. (No bash changes — bats unaffected.)

Closes #414

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the Windows install script’s WSL step and tests; behavior is more defensive (timeouts, version floor) with no auth or cluster data impact.
> 
> **Overview**
> Replaces the Windows installer’s inline **90s Store-only `wsl --update` job** with a dedicated **`Update-Wsl`** flow so corporate machines that block the Microsoft Store can still get a current WSL kernel before Docker Desktop runs.
> 
> **Skip when already current:** A bounded `wsl --version` probe (`Get-WslVersionOutput`, 20s job timeout) feeds **`Test-WslCurrent`**, which compares the first dotted version in the output to a **2.1.5 floor** (Docker Desktop’s documented minimum; overridable via `TB_WSL_MIN_VERSION`). Re-runs on an up-to-date box exit in under ~2s instead of waiting up to 90s every time.
> 
> **Update path:** **`Invoke-WslUpdate`** runs `wsl --update` as a tracked process (5-minute deadline, stdout/stderr to temp files and the install log). The installer prefers **`--web-download`** (Microsoft CDN, not the Store), then retries plain **`wsl --update`** once if the first attempt fails with a real exit code (not timeout or missing `wsl.exe`).
> 
> **Failures on screen:** Timeouts, missing `wsl.exe`, and non-zero exits get distinct warnings plus an arch-matched manual MSI hint (`x64` / `arm64` from `Get-WindowsArch`). Automated GitHub MSI download is intentionally not added (conflicts with #410’s no-GitHub-API rule).
> 
> Pester coverage exercises version parsing, `Update-Wsl` branching, and source guards; **`scripts/manifest.sha256`** is updated for `install-k8s.ps1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 85cf456b237f2688d5b3930cc19d035c76e50aaa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/install-k8s.ps1
```diff
@@ -622,6 +622,51 @@ function Enable-OneVirtFeature {
...
+#     component the API-free /releases/latest redirect can't resolve), and #410 --
+#     enforced by a test -- forbids the rate-limited GitHub API in this installer.
+function Update-Wsl {
+  $verOut = ""
+  try { $verOut = (& cmd /c "wsl --version 2>nul" | Out-String) } catch {}
+  if (Test-WslCurrent -VersionOutput $verOut) {
+    Log "WSL already current: $($verOut -replace '\r?\n',' ')"
+    Ok "WSL is current"
+    return
+  }
```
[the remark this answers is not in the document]
shujaatTracebloc: Good catch — this broke skip-when-current entirely. Fixed in the latest push: `wsl --version` is now captured with `[Console]::OutputEncoding = [System.Text.Encoding]::Unicode` (the same pattern the `wsl --list` reader already uses), restored in a `finally`. WSL's UTF-16LE output now decodes cleanly so Test-WslCurrent matches and the <2s skip-when-current path fires on re-runs instead of always attempting the 5-min web update. Added a source guard.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,60 @@ function Enable-OneVirtFeature {
...
+      $ok = ($p.ExitCode -eq 0)
+    } else {
+      Log "wsl --update --web-download timed out (process killed)."
+    }
+  } catch { Log "wsl --update --web-download failed: $_" }
+  if ($ok) { Ok "WSL updated"; return }
+
+  # Surface the failure + the exact manual next step ON SCREEN (not the log).
+  Warn "Couldn't update WSL automatically (the Microsoft Store may be blocked)."
+  Hint "Download the latest WSL MSI from https://github.com/microsoft/WSL/releases (wsl.<version>.x64.msi), run it, then re-run this installer -- otherwise Docker Desktop will prompt you to install WSL."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. The manual hint now names the arch-matched package: `$msiArch = if ((Get-WindowsArch) -eq 'arm64') { 'arm64' } else { 'x64' }` -> `wsl.<version>.$msiArch.msi`, so an ARM64 operator is pointed at the arm64 MSI instead of x64. Added a source guard.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,64 @@ function Enable-OneVirtFeature {
...
 }
 
+# Modern (Store or standalone) WSL prints a version block from `wsl --version`;
+# legacy/absent WSL errors or prints nothing. Returns $true when WSL is already
+# installed and current enough that no update is needed (#414). Takes the command
+# output as a parameter so it's unit-testable without WSL present.
+function Test-WslCurrent {
+  param([string]$VersionOutput)
+  return [bool]($VersionOutput -match '(?im)^\s*WSL version:\s*\d+\.\d+')
+}
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. Test-WslCurrent now matches the dotted version number (`\d+\.\d+\.\d+`) that `wsl --version` always prints, instead of the English `WSL version:` label which gets localized (e.g. Japanese `WSL バージョン:`). So skip-when-current fires on non-English Windows too. Added a Japanese-label test case.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+# output as a parameter so it's unit-testable without WSL present.
+#
+# Match the version NUMBER, not the English "WSL version:" label -- `wsl --version`
+# localizes its labels (e.g. Japanese "WSL バージョン:"), so label-matching would
+# fail skip-when-current on non-English Windows (#414 Bugbot). Modern WSL always
+# prints dotted version numbers (e.g. 2.3.26.0); legacy/absent WSL errors out with
+# none.
+function Test-WslCurrent {
+  param([string]$VersionOutput)
+  return [bool]($VersionOutput -match '\d+\.\d+\.\d+')
```
saadqbal: This doesn't test currency. `\d+\.\d+\.\d+` matches anywhere in the block, and `wsl --version` always prints `Kernel version:` and `Windows version:` lines — so it returns true for any modern WSL, including a 2.0.x from 2023. Effectively `Test-WslPresent`.

Net effect: a box on a stale WSL gets a green `Ok "WSL is current"`, the update never runs, and Docker Desktop prompts anyway — the #414 symptom, now unfixable by re-running since every run short-circuits. Needs a floor: pull the first dotted number off the WSL version line and compare (`[version]$v -ge [version]'2.x.y'`).

saadqbal (commented):
Good direction and the Bugbot rounds landed cleanly. One thing I'd want fixed before merge: `Test-WslCurrent` is a presence check, not a currency check, so the new skip path no-ops on exactly the stale machines #414 is about. Beyond that, the `wsl --version` read is now the only unbounded call in the file, and the failure path logs nothing. Comments inline.

One outside the diff: `install-k8s.ps1:948` still tells the operator to run `wsl --update` manually — the Store command this PR just removed as broken. Worth updating in the same pass.

(Manifest hash matches, check-style clean.)

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+function Update-Wsl {
+  # wsl.exe writes UTF-16LE; capture it with the console encoding set to Unicode
+  # (same pattern as the `wsl --list` reader below), otherwise the output is
+  # null-interleaved and Test-WslCurrent never matches -- skip-when-current would
+  # never fire and every re-run would attempt a full web update (#414 Bugbot).
+  $verOut  = ""
+  $prevEnc = [Console]::OutputEncoding
+  try {
+    [Console]::OutputEncoding = [System.Text.Encoding]::Unicode
+    $verOut = (wsl --version 2>$null | Out-String)
```
saadqbal: This is the only unbounded external call in the file. `wsl --set-default-version 2` gets 20s, `wsl --list` 30s, `wsl --install -d Ubuntu` 600s, and the update right below gets 5 min — this one gets nothing, and prints nothing while it runs. A wedged LxssManager (plausible on the same corporate boxes this targets) freezes Step 1 forever with a blank console; the 90s job you replaced at least degraded to a warning. `.cursor/BUGBOT.md` calls this out explicitly under "Always flag".

The `wsl --list` reader at ~938 is a drop-in template — `Start-Job -InitializationScript $JobInit` + `Wait-JobWithProgress -TimeoutSec 20`, encoding handling included. Treat a timeout as not-current so the web download still runs.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+  if (Test-WslCurrent -VersionOutput $verOut) {
+    Log "WSL already current: $($verOut -replace '\r?\n',' ')"
+    Ok "WSL is current"
+    return
+  }
+
+  Info "Updating WSL (web download, bypassing the Microsoft Store)..."
+  $ok = $false
+  try {
+    $p = Start-Process -FilePath "wsl" -ArgumentList "--update","--web-download" -NoNewWindow -PassThru -ErrorAction Stop
```
saadqbal: No `-RedirectStandardOutput`/`-RedirectStandardError`, and the non-zero-exit path (673 → 685) has no `Log` at all — only the timeout and throw branches log.

Two consequences. wsl's own `\r` progress output fights the `Wait-ProcessWithDeadline` spinner on the same console line for up to 5 minutes. And when it fails, the install log and the `-Diagnose` bundle carry zero WSL evidence: no exit code, no stderr, nothing to distinguish a 407 proxy from wsl.exe being absent. The k3d and Docker Desktop paths both redirect to temp files; #414 specifically asked for the failure to stop being swallowed.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+    if (Wait-ProcessWithDeadline -Process $p -Deadline (Get-Date).AddMinutes(5) -Message "Updating WSL (web download)") {
+      $ok = ($p.ExitCode -eq 0)
```
saadqbal: Worth one plain `wsl --update` retry before falling through. `--web-download` is only understood by the serviced wsl.exe — an unpatched Win10 2004–21H2 / Win11 21H2 box rejects the flag, exits non-zero in under a second, and gets sent to a manual MSI, where the `wsl --update` this PR removed would have worked. Narrow and shrinking population, and non-fatal, but it's a cheap two-rung ladder.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+    }
+  } catch { Log "wsl --update --web-download failed: $_" }
+  if ($ok) { Ok "WSL updated"; return }
+
+  # Surface the failure + the exact manual next step ON SCREEN (not the log).
+  # Name the MSI matching this host's architecture -- GitHub ships both x64 and
+  # arm64, and pointing an ARM operator at the x64 MSI just recreates the WSL
+  # prompt this path exists to avoid (#414 Bugbot).
+  $msiArch = if ((Get-WindowsArch) -eq 'arm64') { 'arm64' } else { 'x64' }
+  Warn "Couldn't update WSL automatically (the Microsoft Store may be blocked)."
```
saadqbal: Three different failures land on this one sentence: `Start-Process` throwing (wsl.exe absent), a non-zero exit, and the 5-minute timeout. And "the Microsoft Store may be blocked" is the one cause `--web-download` rules out by construction — your own comment at 642 says as much. Someone on a slow-but-open link gets sent to chase Store policy. Worth splitting into timed out / wsl.exe not found / exited N; the `Hint` below is right either way.

scripts/tests/install-k8s.Tests.ps1
```diff
@@ -1196,6 +1196,58 @@ Describe "Test-PreflightRuntimeMem (post-Docker, warn-only)" {
...
+    # `wsl --version` localizes the label; match the dotted version, not "WSL version:".
+    Test-WslCurrent -VersionOutput "WSL バージョン: 2.3.26.0`nカーネル バージョン: 5.15.167.4-1" | Should -BeTrue
+  }
+}
+
+Describe "Update-Wsl / WSL install (#414 Store-blocked fallback)" {
+  BeforeAll { $script:WSRC = Get-Content "$PSScriptRoot/../install-k8s.ps1" -Raw }
+  It "prefers the Store-free web download" {
+    # bypasses the Microsoft Store, which corporate networks block
+    $script:WSRC | Should -Match 'wsl["'']?\s*.*--update.*--web-download|"--update","--web-download"'
```
saadqbal: This passes without the call. The first alternation branch matches the doc comment at `install-k8s.ps1:642` and both `Log` strings at 675/677 — delete the `Start-Process` line and the test stays green while the Store-blocked regression ships. Anchor it on the invocation the way the k3d test does: `Should -Match 'Start-Process -FilePath "wsl" -ArgumentList "--update","--web-download"'`.

scripts/tests/install-k8s.Tests.ps1
```diff
@@ -1196,6 +1196,58 @@ Describe "Test-PreflightRuntimeMem (post-Docker, warn-only)" {
...
+  It "legacy error text (no version block) -> not current" {
+    Test-WslCurrent -VersionOutput "Windows Subsystem for Linux has no installed distributions." | Should -BeFalse
+  }
+  It "non-English (localized) label still detected via the version number (Bugbot #414)" {
+    # `wsl --version` localizes the label; match the dotted version, not "WSL version:".
+    Test-WslCurrent -VersionOutput "WSL バージョン: 2.3.26.0`nカーネル バージョン: 5.15.167.4-1" | Should -BeTrue
+  }
+}
+
+Describe "Update-Wsl / WSL install (#414 Store-blocked fallback)" {
```
saadqbal: Everything in this Describe is a grep of the source — `Update-Wsl` is never executed, so the skip branch, the exit-code handling and the arch hint are all unverified, and none of these would go red if someone inverted the skip condition or dropped the `Update-Wsl` call from `Enable-VirtualisationFeatures`.

`Enable-OneVirtFeature` — the sibling function in the same caller — is mocked properly at ~2118 (`Should -Invoke Warn -ParameterFilter { … }`). All this needs is a `function wsl { }` stub alongside the kubectl/docker/helm/k3d ones at the top of the file.

Also missing the case that actually matters above: `Test-WslCurrent -VersionOutput "WSL version: 2.0.0.0"`. Both current cases are the same fresh 2.3.26.0 string, so a correct and a broken implementation pass identically.

scripts/tests/install-k8s.Tests.ps1
```diff
@@ -1196,6 +1196,58 @@ Describe "Test-PreflightRuntimeMem (post-Docker, warn-only)" {
...
+  It "the manual MSI hint names the arch-matched package, not hardcoded x64 (Bugbot #414)" {
+    # ARM64 hosts need wsl.<version>.arm64.msi; pointing them at x64 recreates the prompt.
+    $script:WSRC | Should -Match "Get-WindowsArch\) -eq 'arm64'"
+    $script:WSRC | Should -Match 'wsl\.<version>\.\$msiArch\.msi'
+  }
+  It "does not reintroduce the rate-limited GitHub API for an auto-MSI fallback (#410 invariant)" {
+    # An automated GitHub-releases MSI download would need api.github.com (the asset
+    # name isn't resolvable via the API-free redirect); #410 forbids it, so we
+    # surface a manual step instead. Guard against a regression that re-adds it.
+    $script:WSRC | Should -Not -Match 'https://api\.github\.com'
```
saadqbal: Duplicate of the #410 guard at 1641 (same file read, same regex), and it can't go red for anything this PR did — it passes on the pre-PR source too. It also only stays green because the new comment at `install-k8s.ps1:647` writes `api.github.com` without the scheme; tidy that into a proper URL some day and two unrelated Describes fail with a message about a fetch that doesn't exist. I'd drop it and let the #410 Describe own the invariant.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,70 @@ function Enable-OneVirtFeature {
...
+  # (same pattern as the `wsl --list` reader below), otherwise the output is
+  # null-interleaved and Test-WslCurrent never matches -- skip-when-current would
+  # never fire and every re-run would attempt a full web update (#414 Bugbot).
+  $verOut  = ""
+  $prevEnc = [Console]::OutputEncoding
+  try {
+    [Console]::OutputEncoding = [System.Text.Encoding]::Unicode
+    $verOut = (wsl --version 2>$null | Out-String)
+  } catch {}
+  finally { [Console]::OutputEncoding = $prevEnc }
```
saadqbal: Worth wrapping this restore: `finally { try { [Console]::OutputEncoding = $prevEnc } catch {} }`.

A throw inside `finally` isn't caught by that statement's own `catch {}` — it propagates, and `Update-Wsl` is called unguarded at 722 with no top-level handler, so the installer dies at Step 1. And the restore is specifically the line that can throw here: the .NET Framework setter short-circuits `SetConsoleOutputCP` when the value is the standard Unicode encoding, so line 658 is a no-op at the Win32 level, while restoring a non-Unicode `$prevEnc` does make the call — which fails with ERROR_INVALID_HANDLE on a console-less process.

Narrow (PS remoting / `win_shell` / session-0, not `-WindowStyle Hidden`), but the file already designs for non-interactive hosts — `Test-CanPrompt` at 1439, the try-wrapped `Read-Host`s at 1567/1593 — so it's not out of scope. The `wsl --list` reader at 942 has the same unprotected restore, but a throw there is contained by the job.

saadqbal (commented):
One more, missed it on the first pass.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 9d8132b. Test-WslCurrent now grades a floor, not presence: it pulls the first dotted version (the WSL version line) and requires `[version] >= MinVersion` (default 2.1.0, TB_WSL_MIN_VERSION overrides). A stale 2.0.x now returns false and still updates. Added test cases for 2.0.0 (stale) and a custom floor.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 9d8132b. The wsl --version probe is now bounded: Get-WslVersionOutput runs it in a job with Wait-JobWithProgress -TimeoutSec 20 (the wsl --list template) and returns "" on timeout, so a wedged LxssManager can't freeze Step 1 — a timeout is treated as not-current so the update still runs.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 9d8132b. Invoke-WslUpdate now runs wsl --update with -RedirectStandardOutput/-RedirectStandardError to temp files, logs them (so the install log + -Diagnose carry the exit code/stderr — 407 vs missing wsl.exe is distinguishable), and the redirect stops wsl's \r progress fighting the spinner.

scripts/install-k8s.ps1
shujaatTracebloc: Added in 9d8132b. On a non-zero --web-download exit (unpatched wsl.exe rejecting the flag), Update-Wsl now retries plain `wsl --update` before falling through to the manual step — the two-rung ladder you described. Covered by a branching test.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 9d8132b. The failure is now classified and messaged specifically: 'wsl.exe wasn't found' / 'Updating WSL timed out and was stopped' / 'wsl exited N'. Dropped the 'Store may be blocked' line, since --web-download rules that cause out. Tests assert the timeout and not-found wording.

scripts/tests/install-k8s.Tests.ps1
shujaatTracebloc: Fixed in 9d8132b. The web-download guard now anchors on the actual invocation — `Invoke-WslUpdate -ExtraArgs @("--web-download")` — and there's a functional test that executes Update-Wsl and asserts web-download is tried first, so deleting the call fails the suite.

scripts/tests/install-k8s.Tests.ps1
shujaatTracebloc: Done in 9d8132b. New 'Update-Wsl branching' Describe actually EXECUTES Update-Wsl (mocking Get-WslVersionOutput / Test-WslCurrent / Invoke-WslUpdate) across skip / web-download-ok / retry / timeout / not-found — inverting the skip or dropping the retry now goes red. Added the Test-WslCurrent 2.0.0-stale case so a correct and a broken impl no longer pass identically.

scripts/tests/install-k8s.Tests.ps1
shujaatTracebloc: Dropped in 9d8132b — removed the duplicate no-API guard from this Describe; the #410 Describe owns that invariant. (The auto-MSI fallback was already removed earlier for the same #410 reason.)

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 9d8132b. The encoding restore is now `finally { try { [Console]::OutputEncoding = $prev } catch {} }` inside Get-WslVersionOutput's job, so a restore throw (ERROR_INVALID_HANDLE on a console-less host) can't propagate out of the unguarded Update-Wsl call and kill Step 1.

scripts/install-k8s.ps1
```diff
@@ -622,6 +622,115 @@ function Enable-OneVirtFeature {
...
+# (#414 reviewer -- matching any dotted number was effectively Test-WslPresent).
+# TB_WSL_MIN_VERSION overrides the floor.
+function Test-WslCurrent {
+  param(
+    [string]$VersionOutput,
+    [string]$MinVersion = $(if ($env:TB_WSL_MIN_VERSION) { $env:TB_WSL_MIN_VERSION } else { "2.1.0" })
+  )
+  $m = [regex]::Match($VersionOutput, '\d+\.\d+\.\d+(\.\d+)?')
+  if (-not $m.Success) { return $false }
+  try { return ([version]$m.Value -ge [version]$MinVersion) } catch { return $false }
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in the latest push. Bumped the default currency floor to Docker Desktop's documented WSL minimum, 2.1.5 (was 2.1.0), so 2.1.0-2.1.4 boxes are now treated as not-current and updated rather than skipping and hitting Docker Desktop's WSL prompt. Added a 2.1.4 boundary test; TB_WSL_MIN_VERSION still overrides.

saadqbal (approved):

PR #494, opened 2026-07-30
feat(ingestor): resolve the spawned tag per environment (:dev / :stg)

LukasWodka:
**Part 2 of tracebloc/backend#1360.** Draft on purpose — see the sequencing gate below.

Part 1 (tracebloc/data-ingestors#422) publishes `:dev` / `:stg`. This is the half that makes them *consumable*: today `images.ingestor.tag` is a **single** value (`"0.7"`) shared by dev and staging, so the two cannot sit on different channels. Only `prodDigest` is env-aware.

## What changes

| key | before | after |
|---|---|---|
| `images.ingestor.tag` | `"0.7"` (the only float) | **explicit override, empty by default** |
| `images.ingestor.channelTags` | — | `{dev: dev, stg: stg, prod: "0.7"}`, keyed on resolved `CLIENT_ENV` |

New `tracebloc.ingestorTag` helper, mirroring `tracebloc.ingestorDigest`'s precedence: **explicit `tag` > channel for `CLIENT_ENV` > literal `0.7`**. That last fallback matters — a release predating these keys still renders under a plain `--reuse-values` replay.

**`prod` deliberately stays a semver float, not a `:prod` channel** — no such tag is published for the ingestor, and prod normally spawns `prodDigest` anyway (the tag is only the fallback when pinning is disabled).

## Rendered result (verified, not asserted)

```
CLIENT_ENV=dev       INGESTOR_IMAGE_TAG="dev"   DIGEST=""
CLIENT_ENV=stg       INGESTOR_IMAGE_TAG="stg"   DIGEST=""
CLIENT_ENV=prod      INGESTOR_IMAGE_TAG="0.7"   DIGEST="sha256:9098b3c9…"
CLIENT_ENV=<unset>   INGESTOR_IMAGE_TAG="0.7"   DIGEST="sha256:9098b3c9…"
```

## The CI guard had to move with it

`ingestor-multiarch` in `helm-ci.yaml` **hard-failed on an empty tag**, so this change could not land without it. It now validates the explicit override *when set* plus **every** `channelTags` entry — an edge resolves exactly one of them, so a single-arch value in any would break ingestion on arm64 for whichever environment lands on it (client#186 / #160). Part 1's merge job already refuses to publish a single-arch index, so the channels satisfy it by construction.

## ⚠️ Sequencing gate — why this is a draft

`:dev` and `:stg` **do not exist yet**. Merging this before Part 1 has published both would point dev/staging edges at a nonexistent tag → **ImagePullBackOff on the next ingestion**. Also, the `ingestor-multiarch` job on this PR will legitimately fail until then, because it inspects those tags.

**Order:** merge #422 → let a `develop` and a `staging` push publish `:dev` and `:stg` (confirm both carry amd64+arm64) → mark this ready → merge.

## Behavioural change to be deliberate about

Chart defaults propagate through the fleet auto-upgrade (`--reset-then-reuse-values` re-applies user-supplied values but reads **new chart defaults**), and the installer does not pin the tag — so **existing dev/staging edges move onto their channel on the next upgrade**. That is the intent (a staging edge should run staging code, as jobs-manager and the engine already do), but it is a real change in what those edges run. An operator who set `images.ingestor.tag` explicitly keeps it. Prod is untouched.

These channels are **unsigned** internal images (Part 1 keeps one signed trust root at the prod tags) — hence prod is never pointed at them.

## Verification

- **307/307 `helm unittest` across 27 suites**, including 9 new cases: each environment, the explicit override winning over a channel, an unknown `CLIENT_ENV` falling back rather than rendering empty, and a `channelTags: null` replay.
- One **existing** test legitimately changed: `keeps the ingestor floating on dev` asserted tag `0.7` for `CLIENT_ENV=dev`; that is precisely the behaviour this PR changes, so it now asserts `dev` with the rationale inline. The other two tag assertions were checked and remain correct (no-`CLIENT_ENV` → prod float; the explicit `0.4` override case).
- `helm lint --strict` clean on **all four** platform values files (aks/bm/eks/oc).
- Schema: still rejects `latest` in `channelTags`, now accepts the empty override.
- `scripts/gen-manifest.sh` run — no change (no installer script touched).

---

> [!NOTE]
> **Medium Risk**
> Upgraded dev/staging edges will switch ingestor images on the next chart upgrade (intended but operationally significant), and merging before `:dev`/`:stg` exist causes ImagePullBackOff; prod pinning logic was touched but alias handling is tightened.
> 
> **Overview**
> **Spawns ingestion jobs with environment-specific ingestor tags** instead of one shared `images.ingestor.tag` (`0.7`). `tag` is now an optional override (empty by default); defaults live in new `images.ingestor.channelTags` (`dev` → `:dev`, `stg` → `:stg`, `prod` → `0.7`). jobs-manager’s `INGESTOR_IMAGE_TAG` comes from `tracebloc.ingestorTag` (override → channel for resolved env → `0.7` fallback).
> 
> **Shared `tracebloc.clientEnv`** normalizes `CLIENT_ENV` aliases (`staging`/`production`/etc.) so tag selection and the prod `prodDigest` pin stay aligned—fixing the case where `production` got the prod float tag but lost the digest pin.
> 
> **CI and tooling follow the new model:** `ingestor-multiarch` validates every tag an edge might spawn (explicit override plus each `channelTags` entry); `resolve-ingestor-digest.sh` defaults to `channelTags.prod` when `tag` is empty. Chart **1.9.9**; helm unittest coverage for channels, overrides, aliases, and legacy `--reuse-values` without `channelTags`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa74943cb11a6385ce8e4a804fd7be9e39281f64. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/resolve-ingestor-digest.sh
+++ b/scripts/resolve-ingestor-digest.sh
@@ -42,6 +42,38 @@ write=0
 # Scoped to the images: -> ingestor: block so a sibling image's `tag:`
 # (jobsManager / podsMonitor / requestsProxy / ... each carry their own)
 # can never be picked up by mistake. bash-3.2 / macOS-safe (pure awk).
+# Portable, yq-free reader for images.ingestor.channelTags.prod. Same scoping
+# discipline as read_ingestor_tag: only the 6-space `prod:` leaf inside
+# images: -> ingestor: -> channelTags: can match, so no sibling key can be
+# picked up by mistake. bash-3.2 / macOS-safe.
+read_ingestor_prod_channel() {
+  local file="$1"
+  [[ -f "$file" ]] || return 1
+  awk '
+    /^images:[[:space:]]*$/ { in_images = 1; next }
+    /^[^[:space:]#]/        { in_images = 0; in_ingestor = 0; in_channels = 0 }
+    in_images {
+      if ($0 ~ /^  [A-Za-z_][A-Za-z0-9_]*:[[:space:]]*$/) {
+        in_ingestor = ($0 ~ /^  ingestor:[[:space:]]*$/) ? 1 : 0
+        in_channels = 0
+        next
+      }
+      if (in_ingestor && $0 ~ /^    [A-Za-z_][A-Za-z0-9_]*:[[:space:]]*$/) {
+        in_channels = ($0 ~ /^    channelTags:[[:space:]]*$/) ? 1 : 0
+        next
+      }
+      if (in_channels && $0 ~ /^      prod:[[:space:]]*/) {
+        line = $0
+        sub(/^      prod:[[:space:]]*/, "", line)
+        gsub(/^["'"'"']|["'"'"']$/, "", line)
+        sub(/[[:space:]]*#.*$/, "", line)
+        gsub(/[[:space:]]*$/, "", line)
+        if (line != "") { print line; exit }
+      }
+    }
+  ' "$file"
+}
+
 read_ingestor_tag() {
   local file="$1"
   [[ -f "$file" ]] || return 1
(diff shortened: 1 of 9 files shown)


saadqbal (commented):
Solid PR — verified the render myself (checked out the branch, `helm template` per env, 307/307 unittest, `lint --strict` clean on all four platform files). Prod is safe: digest pin still wins and its tag fallback is unchanged `0.7`, so the behaviour change only touches dev/staging edges as intended. The tag never renders empty on any path I could build — including a `--reuse-values` replay predating the keys (`channelTags: null` → `0.7`) and `prodPin: false` (→ `0.7`), so no empty-`repo:` ImagePullBackOff hiding anywhere.

Keep it a draft until the `ingestor-multiarch` job goes green — merging before #422 publishes `:dev`/`:stg` multi-arch is the only real breakage here, and it's correctly gated.

Two nits, neither blocking:
- The `env.CLIENT_ENV` schema description still says `(dev, staging, prod)`, but resolution keys on `stg` — worth fixing now that the exact string is load-bearing for channel selection.
- The `values.schema.json` reformat (single-line → multi-line + unicode escapes) buries the ~30 real lines in a 576-line diff. Functionally inert, just harder to review.

scripts/resolve-ingestor-digest.sh
```diff
@@ -42,6 +42,38 @@ write=0
 # Scoped to the images: -> ingestor: block so a sibling image's `tag:`
 # (jobsManager / podsMonitor / requestsProxy / ... each carry their own)
 # can never be picked up by mistake. bash-3.2 / macOS-safe (pure awk).
+# Portable, yq-free reader for images.ingestor.channelTags.prod. Same scoping
```
saadqbal, line 75: Tiny nit: this new `read_ingestor_prod_channel` block pushed the old `# Portable, yq-free reader for images.ingestor.tag …` comment (L41-44) away from its function — `read_ingestor_tag` now sits comment-less at ~L77 while its header describes the wrong function above. Worth moving that header back down with it.

saadqbal (dismissed):
Careful, well-scoped PR — the helper precedence, the single `tracebloc.clientEnv` normalizer shared by both `ingestorTag` and `ingestorDigest`, the CI multi-arch guard extension, and the resolver's channelTags.prod fallback all line up, and the unittest coverage is thorough (incl. the alias + reproducibility-pin cases). LGTM on the code.

Approving the code only — merge still waits on the sequencing gate you flagged: #422 has to publish `:dev`/`:stg` (both multi-arch) first, and `ingestor-multiarch` will legitimately stay red until then. One trivial comment inline.

saadqbal (approved):

PR #501, opened 2026-07-30
feat(#420): resume after reboot + install state file (fast idempotent re-runs)

shujaatTracebloc:
## #420 — Resume after reboot + install state file (fast idempotent re-runs)

Two legitimate reboots (Windows feature enablement; Docker/WSL first boot) can interrupt the install. Before, every interruption ended with "re-find and re-paste the one-liner", and each re-run re-walked the whole install (up to ~90s of `wsl --update`, winget checks, full preflight) — which reads as "the installer is slow/broken".

### What's in
- **Schema-versioned state file** `%USERPROFILE%\.tracebloc\install-state.json`. Pure, unit-tested helpers (`New-InstallState`, `ConvertTo-InstallState`, `Add-CompletedStage`, `Test-StateHasStage`, `Test-InstallStateCurrent`) + thin I/O wrappers (`Read`/`Save`/`Set-StageComplete`/`Set-InstallComplete`). Corrupt or incompatible-schema file → a fresh state, **never a throw**. The state is **advisory**: every stage still self-verifies (tools re-checked, cluster re-derived), so a stale checkpoint can never skip real work.
- **Resume-after-reboot via RunOnce.** On a reboot the installer checkpoints `features-reboot-pending` and registers an HKCU RunOnce continuation (`Get-ResumeCommand` reuses the #421 elevation arg-builder + `-Resume`) so the install **resumes at next sign-in with no re-pasting** — for both auto-reboot and manual `-NoReboot`. `-Resume` is forwarded through the admin-gate self-elevation. Only the durable `-File` form carries `-Resume` (the `irm|iex` shim has no param block, per #421; the state file drives the one-liner path anyway). Cleared on success so it never fires spuriously.
- **Fast idempotent re-runs.** Each of the 6 steps checkpoints on completion; on a fresh run where a prior install completed **and** the tools + cluster are still present, the installer prints `already installed -- nothing to do` and exits 0. The gate verifies **real presence**, not just the checkpoint, so the "nothing to do" claim is honest.

### Acceptance criteria
- [x] Reboot mid-install → install resumes at next logon without re-pasting anything *(RunOnce continuation + `-Resume`; live reboot is host-gated, unit-tested here)*.
- [x] Full re-run on a healthy install completes quickly with a clear "nothing to do" summary *(fast-path gated on completed + tools + cluster)*.

### Tests
Pure state helpers (round-trip; corrupt/empty/wrong-schema → fresh; append/dedup/order), `Get-ResumeCommand` (durable `-File` form carries `-File` + switches + `-Resume`; one-liner omits `-Resume`), `Save`/`Read` I/O round-trip + corrupt-file degradation (mocked path — no writes to the real home dir), and source guards for the `-Resume` plumbing, reboot-exit RunOnce, per-step checkpoints, success clear+complete, and the honest fast-path gate. Full suite: **315 passed / 0 failed / 9 skipped**; check-style + check-drift clean; `scripts/manifest.sha256` regenerated (R8).

### Host-gated (noted, not in this PR)
The *literal* end-to-end reboot→resume needs a real Windows host — same class as the scheduled Windows e2e leg (#436). The state-file + RunOnce + fast-path logic is unit-tested here.

Closes #420

---

> [!NOTE]
> **Medium Risk**
> Changes first-run and reboot behavior for the Windows installer (RunOnce, elevation, exit codes) and gates “nothing to do” on cluster/client health; mistakes could skip repair or leave stale resume entries, but logic is advisory with live probes and extensive unit tests.
> 
> **Overview**
> Adds **resume-after-reboot** and a **schema-versioned install state file** under `HOST_DATA_DIR` so interrupted Windows `install-k8s.ps1` runs can continue without re-pasting the one-liner, and healthy machines can exit quickly on repeat runs.
> 
> When Windows features need a reboot, the installer registers an **HKCU RunOnce** continuation (via `Get-ResumeCommand` / `Register-ResumeAfterReboot`) and forwards a new **`-Resume`** switch through self-elevation (`Get-ElevationCommand` / `Invoke-SelfElevate`). The resumed run skips the “nothing to do” fast path and shows a resume banner. RunOnce is cleared on a full successful walk; completion is written to `install-state.json` **only** when the client is `connected` (not merely `starting`), and a stale `completed` flag is cleared on failure so remediation isn’t skipped.
> 
> The **fast path** exits 0 when state says completed **and** live checks pass: all tools on PATH, k3d cluster **running** (bounded `k3d cluster list` job), and client deployments Ready (`Test-ClientHealthy` / shared `Get-ClientDeploymentNames`). Corrupt or wrong-schema state degrades to fresh state without throwing.
> 
> Pester coverage for state helpers, cluster list parsing, resume command wiring, completion vs exit predicates, and manifest hash update for `install-k8s.ps1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 792aec35864469bae957122f75e8129ee0543c2d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/install-k8s.ps1
```diff
@@ -3780,6 +3970,11 @@ Wait-ForClientReady
 # (#418) so they need zero admin actions later. Warn-only -- never fail the install.
 try { Set-DailyUserProvisioning } catch { Log "daily-user provisioning error: $_" }
 
+# The install reached the end: no reboot is pending, so clear any RunOnce
+# continuation and record completion for the next re-run's fast path (#420).
+Unregister-ResumeAfterReboot
+Set-InstallComplete
+
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 2a70158. Completion is now gated on success: new `Test-InstallSucceeded` (connected/starting) is the single source of truth shared by BOTH the completion checkpoint and the exit code, so they can't drift. `Set-InstallComplete` runs only `if (Test-InstallSucceeded)`, and the exit uses `if (-not (Test-InstallSucceeded)) { exit 1 }`. A failed client state (bad_creds/crash/image_pull/image_pull_ca) therefore never persists completed=true, so a re-run does NOT hit the fast path and the documented remediation runs. RunOnce is still cleared unconditionally (no reboot is pending once the walk finishes). Added a unit test across connected/starting + failure states + source guards.

scripts/install-k8s.ps1
```diff
@@ -534,6 +537,173 @@ function Start-InstallLog {
...
+  return $true
+}
+
+# Does our k3d cluster already exist? Read-only, bounded, never-fatal.
+function Test-ClusterPresent {
+  try {
+    $out = (& k3d cluster list $CLUSTER_NAME 2>$null | Out-String)
+    return (($LASTEXITCODE -eq 0) -and ($out -match [regex]::Escape($CLUSTER_NAME)))
+  } catch { return $false }
+}
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in af5d579. The probe now requires the cluster to be RUNNING, not merely present: new pure `Test-ClusterRunningInList` parses `k3d cluster list -o json` and returns true only when the named cluster has `serversRunning >= 1`. A present-but-stopped cluster (or one with 0 running servers) -> false -> the fast path is skipped and the normal walk runs `New-K3dCluster`, which starts/repairs it. The fast-path message now states what was actually verified ("installed and the cluster is running"). Unit tests cover running/stopped/absent/corrupt.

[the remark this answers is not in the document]
shujaatTracebloc: Fixed in af5d579. `Test-ClusterRunning` is now BOUNDED like the other external probes: it runs `k3d cluster list` inside `Start-Job` + `Wait-JobWithProgress -TimeoutSec 15` + `Remove-Job`, so a wedged Docker engine can no longer hang the fast path -- a timeout is logged and treated as not-running (fall through to the walk). Added a source guard asserting the probe is job+deadline bounded.

scripts/install-k8s.ps1
```diff
@@ -534,6 +537,197 @@ function Start-InstallLog {
...
+  $inner = Get-ElevationCommand -ScriptPath $ScriptPath -NoReboot:$NoReboot -Diagnose:$Diagnose -DailyUser $DailyUser
+  # Only the durable -File form can carry -Resume; the irm|iex shim has no param
+  # block to bind it (#421), and appending it would sit past -Command's value. The
+  # state file (completed/stages) drives the resume for the one-liner path anyway (#420).
+  if ($inner -match '(^|\s)-File\s') { $inner = "$inner -Resume" }
+  return "powershell.exe $inner"
+}
+
+# Register the RunOnce continuation. Warn-only. Returns $true on success.
+function Register-ResumeAfterReboot {
```
saadqbal, line 718: The RunOnce entry (and `install-state.json` under `$env:USERPROFILE\.tracebloc`) is written from the *elevated* process, so both land in whichever account got elevated. In the over-the-shoulder case this installer explicitly supports — IT admin elevates, hands the box to a different daily user (`-DailyUser`) — that's the admin's hive/profile, not the daily user who signs in next. So RunOnce fires for an account that may never log in, and the daily user's resume silently never happens. Is resume meant to cover the split-account flow, or only same-user elevation? If the former, this needs the daily user's hive (HKU load / a logon scheduled task), not HKCU.

scripts/install-k8s.ps1
```diff
@@ -3780,11 +4003,18 @@ Wait-ForClientReady
 # (#418) so they need zero admin actions later. Warn-only -- never fail the install.
 try { Set-DailyUserProvisioning } catch { Log "daily-user provisioning error: $_" }
 
+# The install reached the end: no reboot is pending, so clear any RunOnce
+# continuation. Record completion ONLY on success -- a failed client state must not
+# arm the "nothing to do" fast path, or a re-run would skip the documented
+# remediation the summary just printed (#420 Bugbot).
+Unregister-ResumeAfterReboot
+if (Test-InstallSucceeded) { Set-InstallComplete }
```
saadqbal: `Test-InstallSucceeded` counts `starting` as success, and `starting` is also the catch-all `Get-NotReadyState` returns when the client just isn't Ready yet (Pending pods, a slow pull that hasn't flipped to ImagePullBackOff, etc.). So a client that never comes up still gets `completed=true` here and arms the fast path — the next run (likely started *because* it never came up) prints "already installed … nothing to do" and exits 0 without re-diagnosing. That's the exact "skip the documented remediation" case the comment says it's guarding against. Either the fast path should also verify the client is actually Ready, or completion shouldn't count `starting`.

saadqbal (commented):
Solid, well-tested state machine and the honest self-verify framing is the right call. Three things I'd want resolved before merge — two are correctness, one is dead weight. Details inline; happy to be wrong on the first if the split-account flow is out of scope for resume.

scripts/install-k8s.ps1
```diff
@@ -534,6 +537,197 @@ function Start-InstallLog {
...
+
+# Mark a stage complete in the live $script:InstallState and persist it.
+function Set-StageComplete {
+  param([string]$Name)
+  $script:InstallState = Add-CompletedStage -State $script:InstallState -Name $Name
+  Save-InstallState -State $script:InstallState
+}
+
+# Is a stage recorded complete in the live state?
+function Test-StageComplete {
```
saadqbal: `Test-StageComplete` / `Test-StateHasStage` have no callers outside the tests, and the per-step `Set-StageComplete 'stepN-…'` checkpoints are written but never read to skip a step — the main flow re-runs all six steps on resume regardless. Only `completed` drives behavior. So a reboot-resume still re-walks the whole install (the ~90s the PR set out to avoid), and the stages array + these two helpers are dead weight. Either wire the stages into per-step skips, or drop them and keep just `completed`.

scripts/install-k8s.ps1
shujaatTracebloc: Good catch — clarified the scope in 1be7c57. Resume is intentionally tied to the elevating account: the reboot happens in Step 1, DURING that account's session, so that same account signs back in and continues (HKCU RunOnce is correct there). The -DailyUser handoff (#418) only runs at the very end, after the client is up. For the split case you describe, I did NOT silently target the wrong hive — instead, when `-DailyUser` differs from the current user, the reboot exit now prints: "Resume is registered for '<current>'. Sign back in as '<current>' to continue; if '<DailyUser>' signs in instead, re-run the installer." Loading the daily user's hive / a logon task for the full cross-account auto-resume is a bigger change I left out of scope; happy to file a follow-up if you want it.

scripts/install-k8s.ps1
shujaatTracebloc: Fixed in 1be7c57. You're right that `starting` is the not-ready catch-all. Completion now requires the client to be actually CONNECTED: new `Test-InstallConnected` ($script:ClientState -eq "connected") gates `Set-InstallComplete`, so a client stuck starting/Pending is never marked complete and never arms the fast path — a re-run re-diagnoses. The exit code deliberately stays more lenient (`Test-InstallSucceeded` = connected||starting) so a slow-but-starting client doesn't hard-fail the run; the two predicates are documented as intentionally different. Unit tests cover connected vs starting vs each failure state for both.

scripts/install-k8s.ps1
shujaatTracebloc: Agreed — dropped in 1be7c57. Since the six steps set shared $script: state that later steps need, a resume has to re-walk them (speed comes from each step's own self-skip: tools present, WSL current, cluster running), so per-step skipping wasn't safe. Rather than keep write-only checkpoints, I removed the whole stage machinery: the stages array, Add-CompletedStage, Test-StateHasStage, Set-StageComplete, Test-StageComplete, the per-step calls, and features-reboot-pending. State is now just {schema, completed}, which is all that drives behavior. Added a guard test that the machinery stays gone.

scripts/install-k8s.ps1
```diff
@@ -3737,9 +3924,25 @@ if ($Diagnose) { Invoke-DiagnoseBundle; exit 0 }
...
+# Fast path (#420): a prior run completed successfully AND the tools + a RUNNING
+# cluster are still here -> nothing to do. Honest: verifies real, running state (not
+# just the checkpoint), so a stopped cluster falls through to New-K3dCluster's
+# start/repair. Skipped on -Resume (a resume must finish the interrupted walk).
+if ((-not $Resume) -and $script:InstallState.completed -and (Test-ToolsPresent) -and (Test-ClusterRunning)) {
+  Ok "tracebloc is already installed and the cluster is running -- nothing to do."
+  Hint "Delete ~\.tracebloc\install-state.json (or set a fresh HOST_DATA_DIR) to force a full reinstall."
+  Unregister-ResumeAfterReboot
+  try { Stop-Transcript | Out-Null } catch {}
+  exit 0
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 792aec3. The fast path now verifies live client health, matching the bash assess path: it additionally requires `Test-ClientHealthy`, which finds the installed release namespace via `Get-InstalledClientInfo` (bounded by Test-ApiReachable) and runs a SHORT `kubectl rollout status --timeout=5s` on each client deployment (mysql-client / <ns>-jobs-manager / <ns>-requests-proxy). If the release is missing or any workload isn't Ready -> false -> the run falls through to the repairing walk. And the "never cleared" gap is closed: a walk that doesn't end connected now calls `Clear-InstallCompleted`, so a stale completed=true from an earlier success can't keep the fast path armed over a now-broken client. (Deployment names are shared via a new Get-ClientDeploymentNames so the health check and the readiness gate stay in sync.)

scripts/install-k8s.ps1
```diff
@@ -3737,9 +3924,25 @@ if ($Diagnose) { Invoke-DiagnoseBundle; exit 0 }
...
+if ($Resume) { Ok "Resuming the tracebloc install after a reboot..." }
 Print-Roadmap
 
+# Fast path (#420): a prior run completed successfully AND the tools + a RUNNING
+# cluster are still here -> nothing to do. Honest: verifies real, running state (not
+# just the checkpoint), so a stopped cluster falls through to New-K3dCluster's
+# start/repair. Skipped on -Resume (a resume must finish the interrupted walk).
+if ((-not $Resume) -and $script:InstallState.completed -and (Test-ToolsPresent) -and (Test-ClusterRunning)) {
+  Ok "tracebloc is already installed and the cluster is running -- nothing to do."
+  Hint "Delete ~\.tracebloc\install-state.json (or set a fresh HOST_DATA_DIR) to force a full reinstall."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 792aec3. The hint now interpolates `$(Get-InstallStatePath)` instead of hard-coding `~\.tracebloc\install-state.json`, so it names the real file location even when HOST_DATA_DIR is overridden.

aptracebloc (approved):

PR #490, opened 2026-07-30
chore(chart): point spawned ingestor at the 0.8 line (D16 write path)

LukasWodka:
> ⚠️ **Ready for code review — but DO NOT MERGE until the D16 dev-edge validation below is signed off.** The artifact hold that put this in draft is gone (the image exists and the digest is now current); the remaining gate is functional and needs a real edge. Review can proceed in parallel.

## What

Retargets the **spawned ingestor** on prod edges from the 0.7 release line to 0.8, which carries the D16 `ds_<hex>` per-ingestion write path (data-ingestors#408). A prod edge left on 0.7 spawns a pre-D16 ingestor and silently falls back to legacy shared tables even with `perIngestionTables` on — found on staging by @divyasinghds.

- `images.ingestor.channelTags.prod` `0.7` → `0.8`. dev/stg untouched — they track the `:dev`/`:stg` internal channels (backend#1360).
- `images.ingestor.prodDigest` → `sha256:02da1ebd…`, the **v0.8.4** multi-arch index. This is the actual spawn target on prod edges (`prodPin: true`).
- `_helpers.tpl` last-resort literal `0.7` → `0.8` (the `--reuse-values` fallback for releases predating `channelTags`).
- helm-unittest tag/digest expectations updated.
- `Chart.yaml` → **1.9.34** (one above develop's 1.9.33; the chart only publishes on a version change).

## Digest currency — was stale, now verified

This sat in draft long enough for its pin to rot. It pinned `sha256:05e12494…`, the **v0.8.2** index from 2026-08-04, while the 0.8 line shipped v0.8.3 (08-06) and v0.8.4 (08-10). The old body claimed the pin `== the current :0.8 float`; that stopped being true on 08-06. With `prodPin: true` prod edges would have spawned an ingestor **two releases behind the channel they nominally track**.

Re-verified against the registry rather than inferred:

```
:0.8                      -> sha256:02da1ebd8fac...becda08f37
v0.8.4 release records       sha256:02da1ebd8fac...becda08f37   <- match
v0.8.3 release records       sha256:c642bbbd...
v0.8.2 release records       sha256:05e12494...                 <- previous pin
```

The new index is a proper OCI image index with `linux/amd64` + `linux/arm64` — arm64 checked deliberately, since prod edges include arm nodes.

**Anyone re-reviewing after a delay should redo this check.** A floating channel tag plus a pinned digest drift apart silently, and the failure is invisible: the chart looks pinned and correct.

## Why the write path is still inert after this merges

[...]

PR #503, opened 2026-07-30
feat(chart): provision per-experiment DB credentials (RFC-0003 D10, backend#1181)

LukasWodka:
## Summary

The **chart half** of RFC-0003 D10 (backend#1181) — flips client-runtime#235 from inert to live. All flag-gated on `perExperimentDbCreds` (**default false**); default installs render byte-for-byte unchanged.

- **`perExperimentDbCreds`** value (schema-typed, own banner in values.yaml) → renders `PER_EXPERIMENT_DB_CREDS=1`, `TB_CREDMGR_USER=tb_credmgr`, and `TB_CREDMGR_PASSWORD` (via **`secretKeyRef`**, never plain) onto the jobs-manager container.
- **`secrets.yaml`**: generate-once `TB_CREDMGR_PASSWORD` — same upgrade-stable `lookup` pattern as `POD_TOKEN_SIGNING_SECRET` (a regenerated password would needlessly churn the account), emitted only when the flag is on.
- **`rbac.yaml`**: the jobs-manager Role's `secrets` verbs gain `delete` — needed for the revoke path (`_revoke_db_creds` deletes the per-job cred Secret; the Secret is otherwise GC'd via its Job ownerReference).
- **`Chart.yaml` 1.9.8 → 1.9.9** (chart publishes on version change).

jobs-manager self-provisions the `tb_credmgr` account from this Secret on startup (client-runtime#235) — no Helm hook Job needed against the digest-frozen mysql image.

## Safety
Default false = today's shared-credential behavior, bit-for-bit. Enablement is per environment (dev first), after the OQ4 image check (all task images already read the injected creds — verified). Companion: **client-runtime#235** (the mint/inject/revoke code); merge either order — both are inert until this flag is on.

## Coordination
Touches the jobs-manager env, secrets, rbac, and values — **not** the `images.ingestor` block, so no collision with the #1360 / #490 ingestor-pin work. Chart bumped to 1.9.9; if #490 lands first at 1.9.9, I'll rebase this to 1.9.10.

## Tests
2 new helm-unittest cases (env absent by default; flag on renders the 3 env entries incl. the secretKeyRef). **300 helm tests pass.**

Epic: backend#1151 · Design: backend#1181 · Code: tracebloc/client-runtime#235

---

> [!NOTE]
> **Medium Risk**
> Touches credential material and RBAC for Secret deletion, but behavior stays off by default and is limited to chart templating until operators enable the flag alongside client-runtime mint/revoke code.
> 
> **Overview**
> Adds **RFC-0003 D10** chart wiring so environments can turn on per-experiment MySQL credentials via **`perExperimentDbCreds`** (default **false**; unchanged installs stay byte-identical). Bumps the chart to **1.9.9**.
> 
> When the flag is on, the chart exposes **`PER_EXPERIMENT_DB_CREDS`**, **`TB_CREDMGR_USER`**, and **`TB_CREDMGR_PASSWORD`** (from a **`secretKeyRef`**) on jobs-manager; materializes upgrade-stable **`TB_CREDMGR_PASSWORD`** in the main Secret (generate-once / lookup / optional **`credmgrPassword`** pin with alphanumeric validation); and grants jobs-manager **`delete`** on **Secrets** only (ClusterRole and namespace Role) for per-job cred Secret cleanup. Values and schema document the knobs; helm-unittest cases cover default-off behavior and flag-on rendering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3afec19fce5168e34c5ee7feffe20acf8b1619df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/templates/rbac.yaml
+++ b/client/templates/rbac.yaml
@@ -49,7 +49,16 @@ rules:
     # on a create-409 it reads the existing ConfigMap/Secret to confirm the
     # content matches before reusing it. Without `get`, those reads return
     # Forbidden and the endpoint 500s instead of the intended 409/replay.
-    verbs: ["create", "get", "delete"]
+    verbs: ["create", "get"]
+{{- if .Values.perExperimentDbCreds }}
+  # Per-experiment DB creds only: jobs-manager's revoke/sweep path deletes the
+  # per-job cred Secret. Scoped to `secrets` alone (never configmaps) and
+  # flag-gated so a default install grants no extra delete — least privilege,
+  # and byte-for-byte unchanged when off (Saqlain review).
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["delete"]
+{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -105,7 +114,16 @@ rules:
     # on a create-409 it reads the existing ConfigMap/Secret to confirm the
     # content matches before reusing it. Without `get`, those reads return
     # Forbidden and the endpoint 500s instead of the intended 409/replay.
-    verbs: ["create", "get", "delete"]
+    verbs: ["create", "get"]
+{{- if .Values.perExperimentDbCreds }}
+  # Per-experiment DB creds only: jobs-manager's revoke/sweep path deletes the
+  # per-job cred Secret. Scoped to `secrets` alone (never configmaps) and
+  # flag-gated so a default install grants no extra delete — least privilege,
+  # and byte-for-byte unchanged when off (Saqlain review).
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["delete"]
+{{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/client/templates/rbac.yaml
+++ b/client/templates/rbac.yaml
@@ -49,7 +49,7 @@ rules:
     # on a create-409 it reads the existing ConfigMap/Secret to confirm the
     # content matches before reusing it. Without `get`, those reads return
     # Forbidden and the endpoint 500s instead of the intended 409/replay.
-    verbs: ["create", "get"]
+    verbs: ["create", "get", "delete"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
@@ -105,7 +105,7 @@ rules:
     # on a create-409 it reads the existing ConfigMap/Secret to confirm the
     # content matches before reusing it. Without `get`, those reads return
     # Forbidden and the endpoint 500s instead of the intended 409/replay.
-    verbs: ["create", "get"]
+    verbs: ["create", "get", "delete"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/client/templates/secrets.yaml
+++ b/client/templates/secrets.yaml
@@ -25,14 +25,28 @@
 {{- end -}}
 {{- /*
   Credential-manager password (RFC-0003 D10, backend#1181). Same generate-once
-  stability as the signing secret above: jobs-manager syncs the tb_credmgr
-  account password to this value on every startup, so it must be stable across
-  upgrades (regenerating it would needlessly churn the account). Only emitted
-  when perExperimentDbCreds is on.
+  stability + 3-tier resolution as the signing secret above:
+    1. explicit .Values.credmgrPassword (operator pin — DR, a pre-created
+       MySQL account, or a forced rotation), else
+    2. the value already stored in the live Secret (preserve across upgrades), else
+    3. a freshly generated random secret (first install).
+  jobs-manager syncs the tb_credmgr account to this value on every startup, so
+  it must be stable across upgrades (regenerating it would churn the account).
+  Emitted only when perExperimentDbCreds is on.
+  NOTE (edge): because the key is emitted only while the flag is on, an
+  off→on→off→on toggle drops then regenerates it via tier 3, rotating the
+  account until jobs-manager re-syncs. Pin .Values.credmgrPassword to keep it
+  stable across toggles — acceptable for the dev-first rollout; the pin is the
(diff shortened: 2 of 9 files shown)


client/templates/rbac.yaml
```diff
@@ -49,7 +49,7 @@ rules:
     # on a create-409 it reads the existing ConfigMap/Secret to confirm the
     # content matches before reusing it. Without `get`, those reads return
     # Forbidden and the endpoint 500s instead of the intended 409/replay.
-    verbs: ["create", "get"]
+    verbs: ["create", "get", "delete"]
```
saqlainsyed007: **`delete` verb is not gated on the flag** — every install (including default-off) now grants jobs-manager `delete` on Secrets. Under `clusterScope: true` this is a **ClusterRole**, so it's cluster-wide delete on *all* Secrets in *every* namespace, with no `resourceNames` scoping — a buggy/compromised jobs-manager could delete the chart's own `{{ .Release.Name }}-secrets` (CLIENT_PASSWORD, POD_TOKEN_SIGNING_SECRET, TB_CREDMGR_PASSWORD) or anything else.

The env and Secret material were carefully gated to stay byte-identical when off, but this verb is not — and the revoke path that needs it only runs when the flag is on. Suggest wrapping the `delete` in `{{- if .Values.perExperimentDbCreds }}` (here and on the namespace `Role` at line 108).

Same line: **`delete` also lands on `configmaps`** because the rule is `resources: ["configmaps", "secrets"]`, but the revoke path deletes Secrets only. Splitting the rule so `delete` applies to `secrets` alone would keep least privilege.

client/templates/secrets.yaml
```diff
@@ -23,6 +23,21 @@
...
+{{- /*
+  Credential-manager password (RFC-0003 D10, backend#1181). Same generate-once
+  stability as the signing secret above: jobs-manager syncs the tb_credmgr
+  account password to this value on every startup, so it must be stable across
+  upgrades (regenerating it would needlessly churn the account). Only emitted
+  when perExperimentDbCreds is on.
+*/ -}}
+{{- $credmgrPassword := "" -}}
+{{- if .Values.perExperimentDbCreds -}}
+{{-   if and $existingSecret $existingSecret.data (hasKey $existingSecret.data "TB_CREDMGR_PASSWORD") -}}
```
saqlainsyed007: **Off→on toggle regenerates the password.** Because the `TB_CREDMGR_PASSWORD` key is only emitted when the flag is on, disabling the flag rewrites the Secret *without* the key; a later re-enable then hits `hasKey ... == false` and mints a fresh `randAlphaNum 48`. That silently rotates the credmgr account — minting fails with MySQL auth errors until jobs-manager re-syncs on startup — which defeats the generate-once stability this very comment promises. `POD_TOKEN_SIGNING_SECRET` sidesteps this by always rendering. Probably an acceptable edge given the dev-first rollout, but worth a note in the comment at minimum.

client/templates/secrets.yaml
```diff
@@ -23,6 +23,21 @@
...
 {{- end -}}
+{{- /*
+  Credential-manager password (RFC-0003 D10, backend#1181). Same generate-once
+  stability as the signing secret above: jobs-manager syncs the tb_credmgr
+  account password to this value on every startup, so it must be stable across
+  upgrades (regenerating it would needlessly churn the account). Only emitted
+  when perExperimentDbCreds is on.
+*/ -}}
+{{- $credmgrPassword := "" -}}
+{{- if .Values.perExperimentDbCreds -}}
```
saqlainsyed007, line 43: **No operator-pin override.** `POD_TOKEN_SIGNING_SECRET` above resolves in 3 tiers (explicit `.Values` pin → existing Secret → generate); this one has only 2. An operator needing to set the credmgr password to a known/externally-provisioned value (DR, matching a pre-created MySQL account, forced rotation) has no supported hook and must hand-edit the live Secret, which the next `helm upgrade` may overwrite. Consider mirroring the 3-tier pattern with a `.Values.credmgrPassword` pin for consistency.

client/tests/jobs_manager_test.yaml
```diff
@@ -343,3 +343,39 @@ tests:
           content:
             name: PER_INGESTION_TABLES
             value: "1"
+
```
saqlainsyed007, line 346: These new cases only cover the **deployment env**. Nothing guards `secrets.yaml` — no assertion that `TB_CREDMGR_PASSWORD` is absent from the Secret when the flag is off, nor present when on, nor that the RBAC `delete` verb renders. The PR's central "default installs byte-for-byte unchanged" claim for the Secret is therefore unverified; a future edit that renders the key unconditionally would pass CI. Suggest adding a secrets.yaml unittest for both flag states.

saqlainsyed007 (changes requested):
Focused, well-documented change — the env/secret material is cleanly flag-gated. The one structural gap: the RBAC grant is **not** gated on `perExperimentDbCreds`, so it applies to every install and breaks the "default-off / byte-for-byte unchanged" guarantee this PR is built around. Details inline.

saqlainsyed007 (approved):
Re-reviewed after the push — all five points are addressed cleanly. Thanks for the thorough turnaround.

- **RBAC delete gated + secrets-only** — now a separate `resources: ["secrets"]` rule under `{{- if .Values.perExperimentDbCreds }}` in both the ClusterRole and Role; default installs grant no extra delete, and configmaps stay at `create/get`. ✅
- **Operator pin** — `.Values.credmgrPassword` added as tier 1 of a 3-tier resolution mirroring `podTokenSigningSecret`, with fail-fast alphanumeric validation, schema, and values docs. ✅
- **Toggle rotation** — documented as an edge with the pin named as the mitigation. ✅
- **Tests** — `rbac_test.yaml` (default-off no delete; flag-on secrets-only for ClusterRole + Role) and `secrets_test.yaml` (absent by default, present when on, pin flows through, non-alphanumeric pin rejected). `documentIndex` usage matches the existing convention. ✅

appVersion bump to 1.9.9 alongside the chart version is a fine tidy-up. LGTM.

PR #505, opened 2026-07-31
fix(#425): stuck image pulls — classify and tell the truth

shujaatTracebloc:
## #425 — stuck image pulls: classify and tell the truth

`helm upgrade --install` runs with no `--wait`, so helm "succeeds" before any image is pulled. The pull-progress helper then waited `TB_PULL_TIMEOUT` and, on expiry, **always** printed *"Services are still downloading — they'll finish starting in the background."* — even when the pulls had **permanently failed** (x509, blocked registry, auth). A hard failure was reported in success-flavored copy; the user walked away believing the install would complete.

### Bash — `scripts/lib/install-client-helm.sh`
- **`_pull_failure_detail`** — on a visible `ImagePullBackOff`/`ErrImagePull`/`InvalidImageName`, prints the failing pod line + the matching `failed to pull …` event (bounded, non-fatal, self-contained so no cross-lib sourcing).
- **`_progress_end_message`** (pure) — maps `(pulled, total, max_pulled, has_fail)` → `done | failed | downloading | stalled`. A permanent failure **never** maps to `downloading`, so it can't be sold as background progress.
- **`_download_services_progress`** now tracks `max_pulled` and, on timeout, classifies: a failure **warns loudly with the event text**; the "downloading in the background" copy prints **only** when pulls demonstrably progressed; otherwise a neutral "not pulling yet — see the diagnosis".

### Windows — `scripts/install-k8s.ps1`
`Wait-ForClientReady` → `Get-NotReadyState` already classified pull failures into honest red-X summary branches (`image_pull`/`image_pull_ca`/`crash`); the "downloading" copy only ever shows for a genuine `starting` state. This PR also makes the failure copy **carry the event text**: `Get-NotReadyState` captures the pull event (or the failing pod line) into `$script:NotReadyDetail`, and the failure branches print it via the new `Write-NotReadyDetail` — matching the acceptance on both platforms.

### Acceptance
- [x] A permanently failing pull is never described as background progress.
- [x] The failure copy contains the pod event text (both bash and Windows).

### Tests
bats: `_progress_end_message` (all four outcomes; **failure wins over partial progress**), `_pull_failure_detail` (failing → detail + rc 0; healthy → nothing + rc 1), and a guard that the end copy routes through the selector. Pester: `Get-NotReadyState` detail capture (x509 / non-x509 / no-event fallback) + `Write-NotReadyDetail` (labelled block; no-op when empty). Full suites: bats install-client-helm green (one unrelated pre-existing `_extract_yaml_value` failure exists on `develop`, macOS-bash-specific — out of scope), Pester **333 passed / 0 failed**; shellcheck + check-style + check-drift clean; `scripts/manifest.sha256` regenerated (R8).

Closes #425

---

> [!NOTE]
> **Low Risk**
> Installer UX and messaging only; bounded kubectl reads with no change to cluster install or auth behavior.
> 
> **Overview**
> **Fixes #425:** permanent image pull failures were often reported as *"still downloading in the background"* even when pulls had already failed (x509, registry, auth).
> 
> On **bash**, `_download_services_progress` now uses `_pull_failure_detail` and the pure `_progress_end_message` selector (`done` / `failed` / `downloading` / `stalled`) so timeouts warn with pod/event lines instead of implying background progress when nothing will recover.
> 
> On **Windows**, `Get-NotReadyState` fills `$script:NotReadyDetail` from scoped pull events or pod status (including crash loops), and failure summaries call **`Write-NotReadyDetail`** under *"What the cluster reported:"*.
> 
> Adds **bats** and **Pester** coverage; **`scripts/manifest.sha256`** updated.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e7b072a96723f73a0dcc0e3cc14f1ca9c4c4184. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/install-client-helm.sh
```diff
@@ -463,6 +463,43 @@ _no_interactive_creds_die() {
...
+  local ns="$1" kube_timeout="${TB_PROGRESS_KUBECTL_TIMEOUT:-5s}" pods bad events pull_fail
+  has kubectl || return 1
+  [[ -n "$ns" ]] || return 1
+  pods="$(kubectl get pods -n "$ns" --request-timeout="$kube_timeout" 2>/dev/null || true)"
+  bad="$(printf '%s\n' "$pods" | grep -iE 'ImagePullBackOff|ErrImagePull|InvalidImageName' || true)"
+  [[ -n "$bad" ]] || return 1
+  events="$(kubectl get events -n "$ns" --request-timeout="$kube_timeout" 2>/dev/null || true)"
+  pull_fail="$(printf '%s\n' "$events" \
+    | grep -iE 'failed to pull|ErrImagePull|x509|certificate signed by unknown authority|tls: failed to verify' \
+    | tail -n 3 || true)"
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 9c480b7. Scoped the event grep to `failed to pull|ErrImagePull` only — no bare `x509`/TLS terms — matching `summary.sh::_diagnose_not_ready` and the PowerShell path. kubectl prints one event per line, so a genuine x509 pull failure is on a "failed to pull …" line and its detail is still captured, while an UNRELATED x509 event elsewhere in the namespace can no longer displace the real reason via `tail -n 3`. Added a test: three unrelated x509 events after a real "403 Forbidden" pull failure — the 403 survives and x509 is scoped out. Also merged develop to clear the conflict.

scripts/lib/install-client-helm.sh
```diff
@@ -496,18 +535,34 @@ _download_services_progress() {
...
+  # copy that matches reality; the authoritative diagnosis still follows in the
+  # readiness gate + summary.
+  local fail_detail="" outcome
+  if (( pulled < total )); then fail_detail="$(_pull_failure_detail "$ns" || true)"; fi
+  outcome="$(_progress_end_message "$pulled" "$total" "$max_pulled" "$fail_detail")"
+  case "$outcome" in
+    done)
+      success "Downloaded — ${total} services" ;;
+    failed)
+      warn "Some images failed to pull — this won't finish on its own:"
```
saadqbal: Only spot that can now over-claim in the *other* direction: `ImagePullBackOff`/`ErrImagePull` is also what a transient blip or a registry 429 looks like, and kubelet keeps retrying. `wait_for_client_ready` then waits another `READY_TIMEOUT` (300s) and can still resolve to `connected` — so the user could read "this won't finish on its own" here and "✔ Connected" a minute later. Maybe soften to "looks stuck — likely …" rather than the absolute?

scripts/lib/install-client-helm.sh
```diff
@@ -463,6 +463,45 @@ _no_interactive_creds_die() {
...
+# failure it prints the concrete pod status line(s) + the matching pull event and
+# returns 0; when no pull failure is visible it prints nothing and returns 1.
+# Bounded + non-fatal; mirrors summary.sh::_diagnose_not_ready's pull signals but is
+# self-contained so it needs no cross-lib sourcing.
+_pull_failure_detail() {
+  local ns="$1" kube_timeout="${TB_PROGRESS_KUBECTL_TIMEOUT:-5s}" pods bad events pull_fail
+  has kubectl || return 1
+  [[ -n "$ns" ]] || return 1
+  pods="$(kubectl get pods -n "$ns" --request-timeout="$kube_timeout" 2>/dev/null || true)"
+  bad="$(printf '%s\n' "$pods" | grep -iE 'ImagePullBackOff|ErrImagePull|InvalidImageName' || true)"
```
saadqbal, line 477: `bad` gets printed unbounded below (`printf '%s\n' "$bad"`), but the header comment says "Bounded" and the PowerShell path caps with `Select-Object -First 3`. Many failing pods → a wall of indented lines. A `| head -n 3` here would match both the PS behaviour and the comment.

saadqbal (dismissed):
Nice fix — the classifier is clean and the pure `_progress_end_message` is exactly the right seam to unit-test. Approving. Two small non-blockers inline: the `failed` copy can now over-claim in the opposite direction on a transient backoff, and `bad` prints unbounded despite the "Bounded" comment.

scripts/lib/install-client-helm.sh
shujaatTracebloc: Good call — softened in 7e7b072. It now reads "Some images look stuck pulling — this usually needs action, not just more time:" instead of the absolute "won't finish on its own", so a transient blip / registry 429 that later resolves to ✔ Connected doesn't contradict this line. The honest signal (this is NOT plain "downloading in the background") is kept.

scripts/lib/install-client-helm.sh
shujaatTracebloc: Fixed in 7e7b072. `printf '%s\n' "$bad" | head -n 3` now caps the failing-pod lines, matching the PowerShell path's `Select-Object -First 3` and the "Bounded" header comment — so a cluster with many stuck pods no longer prints a wall of indented lines. (The event lines were already bounded via `tail -n 3`.)

aptracebloc (approved):

PR #514, opened 2026-07-31
fix(#496): verify cgroup delegation is active + state the real consequence

shujaatTracebloc:
## #496 — verify cgroup delegation is active (don't assume)

`_write_cgroup_delegation` wrote `user@.service.d/delegate.conf` + ran `systemctl daemon-reload`, then unconditionally printed *"Delegated cpu/cpuset/io … A re-login may be needed."* But `daemon-reload` re-reads unit files **without restarting** the running `user@$(id -u).service`, so `cpu/cpuset/io` aren't in effect for that session — and the k3d node inherits the delegation state from when it's **created**. So limit-bearing pods run **unconstrained** until a re-login **and** a cluster recreate, while the install looks green. The old wording ("may be needed") understated it.

### Fix — the issue's recommended Option 3 (verify, don't assume)
Not Option 2 (`systemctl restart user@…`) — that kills the user's session processes; the issue says instruct, not perform.

- **New `_cgroup_controllers_active`** reads the live `cgroup.controllers` of the user slice (`/sys/fs/cgroup/user.slice/user-$(id -u).slice/cgroup.controllers`; overridable for tests) and is true only when `cpu`+`cpuset`+`io` are all delegated (`memory`/`pids` are default).
- After the reload, `_write_cgroup_delegation` checks it:
  - **active** → `success` "… active in this session."
  - **not active** → a **loud warn**: limits won't enforce until a re-login **and** `k3d cluster delete <name>` + re-run — "the install looks healthy, but limit-bearing workloads run unconstrained."

### Acceptance
- [x] The gap is now observable (read the controllers) rather than inferred, and the message states the real consequence (re-login **and** recreate), not just "may be needed".

### Context (from the issue)
Lower severity in practice: behind the opt-in `TB_TIER1_ROOTLESS` flag (not the validated default, no customer install takes it today), and the path was always non-fatal. This makes the under-stated case honest and verifiable.

### Tests
`_cgroup_controllers_active` (all-present / missing-one / unreadable), and `_write_cgroup_delegation`'s active-vs-not report. `shellcheck --severity=error` + check-style + check-drift clean; full setup-linux.bats green (126); `scripts/manifest.sha256` regenerated (R8).

Closes #496

---

> [!NOTE]
> **Medium Risk**
> Changes rootless Tier 1 install/prepare-host behavior and messaging around cgroup delegation; incorrect checks could mislead users about limit enforcement, but scope is installer scripts and tests only.
> 
> **Overview**
> The Linux installer no longer treats writing the `user@.service` cgroup delegation drop-in as success. It **reads live** `cgroup.controllers` under `user@$UID.service` (not the enclosing slice, which can falsely look delegated) and only reports delegation as active when `cpu`, `cpuset`, and `io` are present.
> 
> **`_write_cgroup_delegation`** and **`_ensure_cgroup_delegation`** always run **`_report_cgroup_delegation`** on every path, including when the drop-in already exists, so re-runs don't silently skip a written-but-not-yet-live delegation. Idempotent content checks use unprivileged **`cmp`** instead of **`sudo cmp`**.
> 
> Full installs get a clear warning that pod CPU/memory limits won't enforce until the user manager restarts (re-login or reboot with lingering) **and** the k3d cluster is recreated. **`prepare-host`** uses **`TB_PREPARE_HOST_MODE`** so messaging targets the researcher's next login and omits cluster-delete advice.
> 
> Bats coverage and **`scripts/manifest.sha256`** are updated for the changed **`setup-linux.sh`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 78c9de2cec93319a96faa4505333742a1b9d8779. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/setup-linux.sh
```diff
@@ -1230,7 +1243,19 @@ _write_cgroup_delegation() {
...
+  # CREATED. If the controllers aren't active, say the REAL consequence loudly: pod
+  # CPU/memory limits won't enforce until a re-login AND a cluster recreate. (We don't
+  # `systemctl restart user@…` for the user: it would kill their session processes.)
+  if _cgroup_controllers_active; then
+    success "Delegated cpu/cpuset/io cgroup controllers (${conf}) — active in this session."
+  else
+    warn "Wrote the cgroup delegation drop-in (${conf}), but it is NOT active in this session yet — pod CPU/memory limits will NOT be enforced until you log out and back in AND recreate the cluster:"
+    hint "  k3d cluster delete ${CLUSTER_NAME:-tracebloc}   # then log out/in and re-run"
+    hint "  (Until then the install looks healthy, but limit-bearing workloads run unconstrained.)"
+  fi
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 49f7a68. The verify/report now runs on EVERY invocation, not just after a fresh write — restructured so the idempotent path (drop-in already present) falls through to `_report_cgroup_delegation` instead of early-returning. So the re-run the hint asks for re-surfaces an inactive delegation rather than taking a silent fast path. Also addressed the linger reality you flagged: the wording now says a re-login may not restart the user manager with lingering enabled and a reboot is the reliable trigger. Test added: a re-run over an existing drop-in still warns when inactive.

scripts/lib/setup-linux.sh
```diff
@@ -1230,7 +1243,19 @@ _write_cgroup_delegation() {
...
+  # THIS session — and the k3d node inherits the delegation state from when it is
+  # CREATED. If the controllers aren't active, say the REAL consequence loudly: pod
+  # CPU/memory limits won't enforce until a re-login AND a cluster recreate. (We don't
+  # `systemctl restart user@…` for the user: it would kill their session processes.)
+  if _cgroup_controllers_active; then
+    success "Delegated cpu/cpuset/io cgroup controllers (${conf}) — active in this session."
+  else
+    warn "Wrote the cgroup delegation drop-in (${conf}), but it is NOT active in this session yet — pod CPU/memory limits will NOT be enforced until you log out and back in AND recreate the cluster:"
+    hint "  k3d cluster delete ${CLUSTER_NAME:-tracebloc}   # then log out/in and re-run"
+    hint "  (Until then the install looks healthy, but limit-bearing workloads run unconstrained.)"
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 49f7a68. Split the report into a mode-aware `_report_cgroup_delegation`: in prepare-host mode it no longer checks the ADMIN's own slice or mentions `k3d cluster delete` (prepare-host creates no cluster) — it just confirms the drop-in is written and takes effect at the RESEARCHER's next login. The full-install path keeps the this-session verify + recreate remedy. Test added for the prepare-host wording.

scripts/lib/setup-linux.sh
```diff
@@ -1216,21 +1229,51 @@ _write_cgroup_delegation() {
...
+    # with pods that can't be given limits (the exact silent breakage this slice
+    # exists to prevent).
+    sudo mkdir -p "$dir" \
+      || { warn "Couldn't create ${dir} for the cgroup delegation drop-in."; return 1; }
+    printf '%s' "$desired" | sudo tee "$conf" >/dev/null \
+      || { warn "Couldn't write the cgroup delegation drop-in at ${conf}."; return 1; }
+    sudo systemctl daemon-reload 2>/dev/null || true
+  fi
+  # Verify + report on EVERY path (#496 Bugbot).
+  _report_cgroup_delegation "$conf"
```
saadqbal, line 1266: On the full-install path this re-surfacing is dead on re-runs. `_ensure_cgroup_delegation` (the only full-install caller) short-circuits at its own fast path (line ~1298: `[[ -f "$conf" ]] && grep -qF 'Delegate=…' && return 0`) *before* ever calling `_write_cgroup_delegation`. So on the 2nd+ run — exactly the #496 "written but not yet active" case — it just `log`s "already present" and returns 0; the verify/warn never runs. The idempotent-branch change here only helps prepare-host's direct call. Either make that fast path verify with `_cgroup_controllers_active` too, or drop it and let `_write_cgroup_delegation` own idempotency + reporting.

scripts/lib/setup-linux.sh
```diff
@@ -1216,21 +1229,51 @@ _write_cgroup_delegation() {
...
+#  - full install: verify THIS session. daemon-reload does NOT restart the running
+#    user@$(id -u).service, so the delegation usually isn't live yet, and the k3d node
+#    inherits the delegation state from when it is CREATED. If inactive, say the real
+#    consequence: limits won't enforce until the user manager restarts AND the cluster
+#    is recreated. With lingering (the rootless path), a re-login may NOT restart the
+#    user manager — a reboot reliably does. We never `systemctl restart user@…`
+#    ourselves: it would kill the user's session processes.
+_report_cgroup_delegation() {
+  local conf="$1"
+  if [[ -n "${TB_PREPARE_HOST_MODE:-}" ]]; then
```
saadqbal, line 1282: This prepare-host branch never fires for the actual prepare-host caller. `run_prepare_host` sets `TB_PREPARE_HOST_MODE=1` only around `install_docker_engine` and resets it to `""` right after (~line 1374), so by the time it calls `_write_cgroup_delegation` (~line 1416) the mode is empty — this falls through to the full-install branch and prints the `k3d cluster delete … # and re-run` advice you built this to avoid, judged on the *admin's* slice. The dedicated test only passes because it sets `TB_PREPARE_HOST_MODE=1` itself. Set the mode around the cgroup write in `run_prepare_host` (or pass it as an arg to `_report_cgroup_delegation`).

saadqbal (commented):
Two things undercut the fix — both make the new behaviour dead on its real paths, and both are masked by tests that set state by hand. Details inline.

scripts/lib/setup-linux.sh
```diff
@@ -1216,21 +1229,51 @@ _write_cgroup_delegation() {
   local desired
   printf -v desired '%s\n[Service]\nDelegate=cpu cpuset io memory pids\n' "$marker"
 
-  # Unchanged → don't rewrite / daemon-reload (avoids churning the user manager).
+  # Unchanged → don't rewrite / daemon-reload (avoids churning the user manager), but
+  # STILL report below — a re-run over an existing drop-in must re-surface an inactive
+  # delegation, not take a silent fast path (#496 Bugbot).
   if [[ -f "$conf" ]] && printf '%s' "$desired" | sudo cmp -s - "$conf" 2>/dev/null; then
```
saadqbal: Minor: the drop-in lives under `/etc` and is world-readable, so this `sudo cmp` is an unnecessary privileged call — `_ensure_cgroup_delegation` reads the very same file with a plain `grep -qF` (no sudo). Dropping the `sudo` here keeps the two presence-checks consistent and avoids a needless elevation on the idempotent path.

scripts/lib/setup-linux.sh
shujaatTracebloc: Fixed in e5a598e — good catch, this was dead on every 2nd+ full install. Routed the fast path through `_report_cgroup_delegation` (an unprivileged controllers read, so the no-sudo property still holds) instead of the plain `log "already present"; return 0`. A written-but-not-yet-active drop-in now re-surfaces the warn+recreate advice on re-run. New test: "drop-in present but NOT active -> fast path re-surfaces the warning, still no privileged call".

scripts/lib/setup-linux.sh
shujaatTracebloc: Fixed in e5a598e. Set TB_PREPARE_HOST_MODE=1 around the cgroup write in run_prepare_host (mirroring how it wraps install_docker_engine) so _report_cgroup_delegation takes the researcher-login branch instead of judging the admin slice + printing cluster-delete advice. New test drives the REAL run_prepare_host (not the helper in isolation): researcher-login wording, no `k3d cluster delete`, admin slice ignored.

scripts/lib/setup-linux.sh
```diff
@@ -1203,6 +1203,19 @@ install_linux() {
...
+# the drop-in but does NOT restart the running user@$(id -u).service, so the delegated
+# controllers only appear after a re-login — this is how we tell "written" from
+# "in effect" instead of assuming. Path overridable (TB_USER_CGROUP_CONTROLLERS) for
+# tests. memory/pids are delegated by default; cpu/cpuset/io are the ones this adds.
+_cgroup_controllers_active() {
+  local f="${TB_USER_CGROUP_CONTROLLERS:-/sys/fs/cgroup/user.slice/user-$(id -u 2>/dev/null).slice/cgroup.controllers}"
+  [[ -r "$f" ]] || return 1
+  local c; c="$(cat "$f" 2>/dev/null)" || return 1
+  [[ " $c " == *" cpu "* && " $c " == *" cpuset "* && " $c " == *" io "* ]]
+}
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed in 78c9de2 — correct and high-impact, thank you. `_cgroup_controllers_active` now reads `user@$(id -u).service/cgroup.controllers` (the runc/rootless-documented node that `Delegate=` on `user@.service` actually enables), not the enclosing `user-$UID.slice`, which routinely lists cpu/io by default (DefaultCPUAccounting) and would read "active" before the drop-in takes effect. Split the default into a unit-testable `_cgroup_controllers_path`; new test asserts it targets `user@$UID.service` and not the bare slice.

scripts/lib/setup-linux.sh
shujaatTracebloc: Done in 78c9de2 — dropped the `sudo`, plain `cmp` now (the drop-in is world-readable under /etc), consistent with `_ensure_cgroup_delegation`'s `grep -qF` and no needless elevation on the idempotent path.

saadqbal (approved):

PR #516, opened 2026-07-31
fix(#418): .wslconfig memory can't be written below the client's own floor

LukasWodka:
## Summary

`Get-WslConfigMemoryGb` writes the WSL2 memory budget into the daily user's `.wslconfig` during elevated provisioning (#418). It did its own arithmetic — physical RAM minus a **private 4 GB reserve**, floored at 1 GB — so it could persist a budget the client cannot run in, and it contradicted the installer's own advice in the same run.

Measured on `develop` (`$env:TB_PESTER="1"; . scripts/install-k8s.ps1`):

| host | `.wslconfig` writes | installer advises (run / train) | |
|---|---|---|---|
| 8 GB | **4 GB** | 6 / 6 | below the 5 GB floor → guaranteed OOM crashloop |
| 16 GB | 12 GB | 8 / 14 | |
| 32 GB | **28 GB** | 8 / 16 | over-committed — Windows left 4 GB |
| 6 GB | 2 GB | 4 / 4 | doomed budget persisted as if intentional |
| 4 GB | 1 GB | 2 / 2 | |

8 GB is the case that matters: a perfectly viable host (8 − 2 GB OS reserve = 6, clear of the floor) got `memory=4GB`, while `Show-MemoryStatus` in the *same run* said "give Docker up to 6 GB".

## The fix

Rather than re-derive a floor here, the function now **delegates to `Get-PfMemRecommendation`** — the existing single-source helper the advice path already uses. The written budget is the advised budget *by construction*, not by two calculations happening to agree:

| host | writes | advises (train) | |
|---|---|---|---|
| 7 GB | `memory=5GB` | 5 | = the floor; 5+2 is the stated practical minimum |
| 8 GB | `memory=6GB` | 6 | was 4 → OOM |
| 16 GB | `memory=14GB` | 14 | was 12 |
| 32 GB | `memory=16GB` | 16 | was 28; Windows keeps 16, and the client can't use more than the recommended training budget |
| 64 GB | `memory=16GB` | 16 | |
| 6 GB | *skipped* | — | |
| 4 GB | *skipped* | — | |

**Reserve reconciled.** The private `ReserveGb = 4` is gone, and the *parameter* is gone with it, so no caller can reintroduce the drift. `$script:PfOsReserveGb` (2) is now the only reserve — which is what its own comment already claimed: *"used to cap recommendations AND to reason about the achievable budget in one place, so the two can't drift"* (#417 reviewer). Three accessors (`Get-PfOsReserveGb` / `Get-PfMinMemGb` / `Get-PfRecMemGb`) give the printing and writing paths one read path for the same numbers; the reserve accessor **fails closed** (never 0, which would hand WSL2 the entire host).

**Host genuinely too small** (physical − reserve < floor): the function returns `0` and the caller **skips the memory setting** rather than persist a budget known to OOM — which would also bake the failure into the daily account for every later run, and read as intentional to whoever debugs it next. It says so plainly instead ("about 7 GB physical is the practical minimum … use a larger machine"), the same honest framing the macOS path landed in #513.

Still **warn-only**: `Set-DailyUserProvisioning` is documented never to fail the install, and Windows memory preflight is warn-only throughout.

## Two judgement calls worth reviewing

**No headroom fudge, unlike colima in #513.** That PR had to size at `floor + 1` because a guest's `MemTotal` runs a few hundred MiB under its configured size and the bash recheck *hard-fails* — so a floor-sized VM tripped on the size the installer had just chosen. Windows doesn't have that catch-22: only a 7 GB host lands exactly on the floor, 7 GB is precisely the practical minimum this code reports, and the Windows recheck warns rather than hard-failing. A `floor + 1` rule here would instead declare 7 GB "too small" and contradict our own stated minimum.

**Deliberately no overlap with #444.** `Show-MemoryStatus` and `Get-PfMemRecommendation` are untouched, since open #444 rewrites both — this is not stacked on it and does not need it. Verified forward-compatible: with #444's floored recommender simulated, every host in the table returns an identical result, because the too-small gate tests the *achievable ceiling* rather than the recommender's output. (#444 may want to adopt the three accessors instead of its inline `$env:PF_MIN_MEM_GB` reads.)

## Test plan

The two assertions in `Describe "Daily-user provisioning (#418)"` that pinned the old cap/floor (`16 -> 12`, `32 -> 28`, `4 -> 1`) are replaced by the new contract:

- never below the client's floor, across seven host sizes
- 8 GB → 6, and explicitly not 4 (the reported bug)
- returns 0 (don't write) when the host can't reach the floor
- never over-commits the host — the OS keeps its reserve; 32 GB explicitly not 28
- caps at the recommended training budget on a big host
- `PF_MIN_MEM_GB` override honoured
- `ReserveGb` parameter is absent, so the drift can't return
- **written == advised** across seven host sizes (the lock that keeps the two paths from drifting again)

Plus source guards that the caller gates the write on `0`, and prints the honest too-small message with the practical minimum.

## Gates

| gate | result |
|---|---|
| `Invoke-Pester scripts/tests/` | **377 passed, 0 failed**, 9 skipped |
| `Invoke-ScriptAnalyzer` (as CI scopes it: `install.ps1` + `install-k8s.ps1`) | **0 errors** |
| `bash scripts/check-style.sh` | clean |
| `bash scripts/tests/check-drift.sh` | no drift |
| `bash scripts/gen-manifest.sh` | `scripts/manifest.sha256` regenerated + committed (R8) |

Refs #418

---

> [!NOTE]
> **Low Risk**
> Installer-only provisioning and preflight alignment; warn-only behavior unchanged, with expanded Pester tests locking the contract.
> 
> **Overview**
> **Daily-user `.wslconfig` memory** no longer uses a separate formula (old 4 GB OS reserve, floor at 1 GB). `Get-WslConfigMemoryGb` now calls **`Get-PfMemRecommendation`** with the same reserve, min, and recommended training values as preflight, so written `memory=` matches on-screen advice (e.g. 8 GB host → 6 GB, not 4).
> 
> **Too-small hosts** get `0` from that helper; provisioning **skips** writing memory, warns with the practical minimum RAM, and does not persist a budget below the client floor.
> 
> **Single source of truth:** new accessors `Get-PfOsReserveGb`, `Get-PfMinMemGb`, and `Get-PfRecMemGb`; the private `ReserveGb` parameter is removed. Pester coverage and `scripts/manifest.sha256` are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 28a89cfbbe48aaaa0ab2991bfc4b98bdf817a0e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shujaatTracebloc (approved):
Reviewed the logic end-to-end — LGTM. ✅

**Correctness verified.** The fix delegates `Get-WslConfigMemoryGb` to the single-source `Get-PfMemRecommendation` = `min(DesiredGb=16, HostGb − reserve)`, and the too-small gate returns 0 whenever `HostGb − reserve < floor`. So for any host that clears the gate, the written budget is in `[floor, 16]` by construction — never below the client's own `PF_MIN_MEM_GB` floor, never over the OS reserve. Traced every host size and each reproduces the tables exactly: 7→5, 8→6 (was the OOM-guaranteed 4), 16→14, 32→16 (was 28), 64→16; 6/4/2/0 → skip. The `0`-return is gated in `Set-DailyUserProvisioning` before `.wslconfig` is touched, so no known-OOM budget is ever persisted — it warns with the honest practical minimum instead.

**Reserve reconciled.** The private `ReserveGb=4` param is gone, leaving `$script:PfOsReserveGb=2` as the only reserve (accessor fails closed). Confirmed the caller at line ~1948 is the sole production call site, so removing the param breaks nothing.

**Tests.** The two old assertions pinned the buggy `16→12 / 32→28 / 4→1` behavior and are correctly replaced by the new contract (never-below-floor, 8→6-not-4, 0-on-too-small, never-over-commit, written==advised drift lock, `PF_MIN_MEM_GB` override) — net coverage 2→10 It blocks.

**Gates green:** Pester 377/0, Lint, Static analysis, drift, manifest regenerated, full E2E matrix, Bugbot clean on `28a89cf` with no open threads.

Nicely scoped away from #444 too. 👍

PR #521, opened 2026-07-31
fix(#429): macOS installs pinned kubectl/k3d/helm via the shared verified path (not bare brew)

shujaatTracebloc:
## #429 — macOS installed CLI tools via bare brew: version pins inoperative, no checksums

On macOS, `kubectl`/`k3d`/`helm` came from bare `brew install`, which:
1. **Silently ignored the `K3D_VERSION`/`HELM_VERSION` pins** — brew floats to latest, so Macs ran different, chart-untested tool versions than the pinned Linux installs (and than the docs claim).
2. Had **no checksum of our own** (delegated to brew).

## Fix — one pinned, verified path across OSes

Both `setup-macos.sh` and `setup-linux.sh` are always sourced, so `install_macos_cli_tools` now routes through the **same** `install_kubectl` / `install_k3d` / `install_helm` (setup-linux.sh) that Linux uses — after setting `OS_DL=darwin` and a macOS tools target (`/usr/local/bin`, on the default login PATH on both Intel and Apple Silicon; no PATH-persistence dance).

The three fetchers are made OS-aware via `${OS_DL:-linux}` — **Linux (and every bats fetch test that leaves `OS_DL` unset) stays byte-identical**; only the macOS runtime sets `darwin`. brew still delivers Docker Desktop / colima unchanged.

### Portable checksum (`common.sh` → `_verify_sha256`)
GNU `sha256sum` on Linux; `shasum -a 256` on macOS — which ships a **BSD `/sbin/sha256sum` (Darwin 1.0) that lacks GNU `--check`**. A `type -t` guard honors the bats mocks' `sha256sum` shell-function, so the existing Linux fetch tests keep passing on macOS dev boxes.

### Execute-gate preserved (#411)
Each shared installer ends in `assert_tool_runs`, so a broken/wrong-arch binary fails the **System tools** step loudly rather than printing a false success. The drift-check contract now accepts macOS delegating to the gated `install_<tool>`.

## Acceptance criteria
- ✅ Fresh macOS install yields exactly the pinned k3d/helm versions (shared pinned path).
- ✅ A broken binary fails Step "System tools" loudly (`assert_tool_runs` in each installer).

## Tests / gates
- New `scripts/tests/setup-macos.bats`: `OS_DL=darwin` fetch+verify for all three tools, `_verify_sha256` portability (match/mismatch/empty + shasum fallback), and `install_macos_cli_tools` delegation with **no bare brew**.
- +2 `check-drift.bats` self-tests for the delegation contract.
- `shellcheck --severity=error`, check-style, check-drift all clean; full `setup-linux.bats` (132) green; `manifest.sha256` regenerated (R8).
- Pre-existing macOS-only failures (validate_config, early-bailout) confirmed unrelated (fail on clean develop).

Closes #429

---

> [!NOTE]
> **Medium Risk**
> Changes how production macOS installs place privileged binaries under /usr/local/bin and rely on sudo; wrong-arch or checksum failures should fail at System tools, but this is a meaningful install-path shift across OSes.
> 
> **Overview**
> **macOS CLI tools** no longer come from floating `brew install` calls. `install_macos_cli_tools` sets `OS_DL=darwin`, installs into `/usr/local/bin` with sudo, and reuses the Linux `install_kubectl` / `install_k3d` / `install_helm` path so **K3D_VERSION** and **HELM_VERSION** pins and checksum verification match Linux.
> 
> Shared plumbing adds **`_verify_sha256`** in `common.sh` (GNU `sha256sum` vs macOS `shasum`, fail-closed). The kubectl/k3d/helm fetchers in `setup-linux.sh` pick assets via **`${OS_DL:-linux}`** and verify through that helper; Linux behavior stays the default when `OS_DL` is unset.
> 
> **Drift checks** treat macOS `install_<tool>` delegation as satisfying execute-gate parity. New **`setup-macos.bats`** covers portable checksums, darwin download URLs, and macOS installer delegation. **`manifest.sha256`** is updated for touched scripts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f96e83ecdf2d7cde22c0fe8f6de76929558479f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc (approved):

scripts/lib/setup-macos.sh
```diff
@@ -258,27 +258,30 @@ install_docker_desktop() {
...
-  success "System tools ready"
+  # kubectl/k3d/helm now come from the SAME pinned, checksum-verified direct-download
+  # path as Linux — install_kubectl / install_k3d / install_helm (setup-linux.sh, always
+  # sourced) are OS-aware via OS_DL, so the K3D_VERSION / HELM_VERSION pins are honored
+  # on macOS instead of brew floating to latest and diverging from the chart-tested
+  # Linux installs (#429). brew still delivers Docker/colima (install_docker_desktop) —
+  # this only moves the version-pinned CLI tools onto the shared path. Each installer
+  # ends in the execute-gate (#411, assert_tool_runs), so a broken/wrong-arch binary
+  # fails the "System tools" step loudly rather than printing a false success.
+  OS_DL="darwin"
```
saadqbal, line 269: Nit: `OS_DL` / `TB_TOOLS_DIR` / `TB_TOOLS_SUDO` are set as globals here and leak past this function. `local` them — bash dynamic scope still reaches install_kubectl/k3d/helm, and you get the module-scope defaults (`TB_TOOLS_SUDO="sudo"`, etc.) restored for free on return instead of leaving `OS_DL=darwin` set for the rest of the run.

saadqbal (approved):
Nice PR — the shared pinned/verified path plus keeping the darwin selectors behind `OS_DL:-linux` (so Linux + the fetch bats stay byte-identical) is the right shape, and the drift/self-tests + setup-macos.bats cover it well. One small nit inline; not a blocker.

PR #537, opened 2026-08-03
test(netpol): characterize the locked-down egress policy (RFC-0003 D6, #199-1)

LukasWodka:
## Summary

**#199, PR 1 of the egress-lockdown work** — test-only, no behavior change. It's the regression net so the actual default flip (#199-2) is a green-to-green change.

The egress mechanism is already fully shipped (squid gateway, the `allowExternalHttps` gate, jobs-manager proxy injection, the seal-check enforcement probe — client-runtime#102 / client#378). #199 is just making deny-by-default the *shipped default*. Flipping `allowExternalHttps` to `false` drops the rule-2 external-HTTPS hole, so the egress rule **indices shift**.

This PR pins the current behavior and characterizes the locked-down state:
- **Pins `allowExternalHttps: true`** in the three cases that assert the rule-2 hole or a rule-2-dependent index (external-443, MySQL, OpenShift) — so they survive the flip.
- **Two new locked-down cases** (`allowExternalHttps: false`): the `0.0.0.0/0` rule is gone (MySQL shifts `egress[2] → egress[1]`); the **unconditional** requests-proxy egress (`8888`) survives so pods still POST results/FLOPs; the egress-proxy allowlist rule (`3128`) renders under `egressProxy.enabled`.

## Tests
**13** network-policy tests (was 11), **322** helm suite green.

Track: #199 — PR-1 (tests) lands anytime · PR-2 (the 4-line flip) is gated on the dev→staging→prod fleet pin-rollout. Epic: backend#1151 (D6).

> Note on assignee: per your epic-ownership call I've assigned this to you; the client-repo convention is `saadqbal` — reassign if you'd rather follow the repo default.

---

> [!NOTE]
> **Low Risk**
> Only Helm unit tests in `network_policy_test.yaml`; no template or runtime behavior changes.
> 
> **Overview**
> **Test-only** prep for RFC-0003 D6 / #199: no chart behavior changes; this is the regression net before `allowExternalHttps` defaults to `false` in a follow-up PR.
> 
> Existing **network-policy** cases that assert the rule-2 external HTTPS hole (`0.0.0.0/0` at `egress[1]`) or rules that sit after it (MySQL at `egress[2]`, OpenShift except blocks) now **explicitly set `allowExternalHttps: true`** so they stay valid when the default changes.
> 
> Two new cases with **`allowExternalHttps: false`** document the locked-down shape: the broad external-443 rule is absent (MySQL moves to `egress[1]`), **requests-proxy** on **8888** remains, and with **`egressProxy.enabled`** the **egress-proxy** rule on **3128** appears at `egress[3]`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 31325bd5f6955dbf0862d958d23d7bb20e473d25. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/tests/network_policy_test.yaml
+++ b/client/tests/network_policy_test.yaml
@@ -145,6 +145,7 @@ tests:
       networkPolicy:
         training:
           enabled: true
+          allowExternalHttps: true  # pin the default — this asserts the rule-2 hole
           dnsNamespace: kube-system
           dnsSelector:
             k8s-app: kube-dns
@@ -176,6 +177,7 @@ tests:
       networkPolicy:
         training:
           enabled: true
+          allowExternalHttps: true  # pin: MySQL sits at egress[2] only while rule 2 renders
           dnsNamespace: kube-system
           dnsSelector:
             k8s-app: kube-dns
@@ -199,6 +201,7 @@ tests:
       networkPolicy:
         training:
           enabled: true
+          allowExternalHttps: true  # pin: asserts rule 2's except-block at egress[1]
           dnsNamespace: openshift-dns
           dnsSelector:
             dns.operator.openshift.io/daemonset-dns: default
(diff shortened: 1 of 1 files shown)


client/tests/network_policy_test.yaml
```diff
@@ -221,3 +224,53 @@ tests:
...
+          dnsNamespace: kube-system
+          dnsSelector:
+            k8s-app: kube-dns
+          clusterCidrs:
+            - 10.0.0.0/8
+    asserts:
+      # the 0.0.0.0/0 external rule is gone — MySQL shifts up from egress[2] to egress[1]
+      - equal:
+          path: spec.egress[1].to[0].podSelector.matchLabels.app
+          value: mysql-client
```
saqlainsyed007, line 247: **Nit (non-blocking, test robustness).** This case proves rule-2 dropped *indirectly* — via MySQL landing at `egress[1]`. That does catch a regression (if rule 2 rendered, MySQL would be at `egress[2]` and this would fail), so coverage is adequate. But since this suite is explicitly the regression net for the flip, a direct assertion of the locked-down *shape* would make the intent unmissable and also catch a stray rule appended after requests-proxy:

```yaml
      # locked-down egress is exactly: DNS + MySQL + requests-proxy
      - lengthEqual:
          path: spec.egress
          count: 3
```

Same idea for the `egressProxy.enabled` case below (`count: 4`). Optional — the PR is correct as-is.

saqlainsyed007 (approved):
### Review summary — APPROVE ✅

Test-only PR (#199-1); the regression net so the `allowExternalHttps` default flip (#199-2) is a green-to-green change. Verified locally.

**What I checked**
- Ran `helm unittest client -f tests/network_policy_test.yaml` → **13 passed**.
- Walked the render order in `templates/network-policy-training.yaml`: `DNS(0) → external-HTTPS(1, gated on allowExternalHttps) → MySQL → requests-proxy → egress-proxy(gated on egressProxy.enabled)`.
  - `allowExternalHttps: false` ⇒ MySQL → `egress[1]`, requests-proxy → `egress[2]` ✓
  - `+ egressProxy.enabled` ⇒ egress-proxy → `egress[3]`, port 3128 ✓
- The three `allowExternalHttps: true` pins land on exactly the rule-2-dependent cases (external-443, MySQL@egress[2], OpenShift except-block@egress[1]); comments are accurate.

**Assessment**
- Correctness: assertions match the template exactly.
- Conventions: matches the existing suite's style; comments explain *why* each pin exists — good for the next reader hitting the index shift.
- No template/runtime change, so risk is confined to CI. Nothing blocking.

One non-blocking robustness nit inline.

client/tests/network_policy_test.yaml
LukasWodka: Agreed the direct `lengthEqual` on `spec.egress` (3 locked-down, 4 with egressProxy) makes the intent unmissable and catches a stray appended rule. Since you've approved and it's non-blocking, I'm resolving to land the flip-regression net now and will add the explicit length assertions as a fast-follow. Thanks!

PR #504, opened 2026-07-30
docs(seal-check): record k3d/k3s NetworkPolicy substrate verification (backend#1184)

LukasWodka:
## Summary

Records the **k3d/k3s NetworkPolicy substrate verification** that RFC-0003 §8.4 explicitly demands ("**do not assume** k3d enforces NetworkPolicy"). Docs-only — one file.

### What was run
On a throwaway `k3d v5.8.3` cluster (`k3s v1.33.6+k3s1`), a standalone deny-egress `NetworkPolicy` (podSelector on a probe pod, `policyTypes: [Egress]`, empty `egress:`) was applied and a `curl` from the pod to `1.1.1.1:443` was measured across three states:

| state | result |
|---|---|
| no policy (baseline) | **reachable** — HTTP 301 |
| deny-egress policy applied | **BLOCKED** — connect failure (egress dropped) |
| policy removed (control) | **reachable again** — HTTP 301 |

The reachable → blocked → reachable transition makes the block **attributable to the policy**, not a fluke or an image-pull artifact. So k3s's embedded (kube-router) controller **does enforce** egress NetworkPolicy on this k3d version — flannel handles pod networking, but the netpol controller is active and enforcing.

### Honest scope (deliberately precise)
This verifies the **substrate primitive** — the specific doubt §8.4 raises. It is **not** the full-chart `egress-enforcement` probe run against a deployed release; that end-to-end run stays open under backend#1184, and I've kept the §8.3 k3d cell reading **"substrate verified; full-probe run pending"** rather than overclaiming "sealed".

### Why it matters
The chart's `egress-enforcement` seal-check relies on the CNI actually dropping egress. That reliance was documented-but-unverified ("expected is not verified"). It's now verified for the k3d/k3s substrate, so the local-dev "nothing gets out" guarantee rests on measured behavior, not an assumption.

Epic: backend#1151 · Ticket: backend#1184 (D12)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Low Risk**
> Documentation-only change to seal-check status and runbook text; no runtime, chart, or security behavior changes.
> 
> **Overview**
> **`docs/SEAL-CHECK.md`** now records a completed **k3s egress NetworkPolicy substrate** check on k3d v5.8.3 / k3s v1.33.6+k3s1 (2026-07-30): deny-egress policy produced reachable → blocked → reachable again on `curl` to `1.1.1.1:443`, with that evidence centralized in the §8.4 Status note.
> 
[...]

PR #541, opened 2026-08-03
test(seal-check): run the egress-enforcement probe live in k3d CI (#1184)

LukasWodka:
## What & why — RFC-0003 D12 seal check (backend#1184)

The chart already ships its enforcement probes as `helm.sh/hook: test` Jobs (egress-enforcement, backend-reachability, storage-assertions), but **`helm test` runs nowhere in CI** — so the probes have never actually *executed* in a pipeline. SEAL-CHECK §8.4 records the k3s NetworkPolicy **substrate** as verified (#504, a manual throwaway-cluster check) while the **full-chart egress-enforcement probe run** is still marked *pending*.

This PR makes that probe run for real, on every CI push.

### Changes
- **`scripts/tests/e2e-seal-check.sh`** — installs the local working-tree chart on a real k3d cluster (reusing the installer's `create_cluster` path) with public images and the egress lockdown **engaged** (`allowExternalHttps=false`), then runs *only* the egress-enforcement seal-check via `helm test --filter`. A training-labelled pod's direct TCP egress to `1.1.1.1:443` must be blocked by k3s's CNI.
- **`.github/workflows/helm-ci.yaml`** — a `seal-check-e2e` job mirroring `upgrade-e2e`: stock `ubuntu-latest`, **zero secrets** (public `curl` image, no backend, no private images).
- **`client/tests/egress_enforcement_check_test.yaml`** — pins the probe Job's `metadata.name` so the e2e `--filter` can never silently drift off it.

### The subtleties worth a reviewer's eye
- `helm test --filter` **exits 0 when the filter matches nothing** — a silent false-pass. The script guards this by asserting the probe hook is in `helm get hooks` *before* testing, then drives the verdict off `helm test`'s **exit code** (the probe Job exits 0 only when egress is verified blocked). It does **not** grep `--logs` (unreliable for Job hooks). The name-pin unittest is belt-and-braces on the filter name.
- **Positive control**: a pass is only trusted if a non-policied pod first proves the host is reachable — so a block is attributable to the NetworkPolicy, not a runner/target issue.

### Why only egress-enforcement
It's the single seal-check that needs no credentials. `backend-reachability` needs a real backend + proxy, and a bound-PVC `storage-assertions` pass needs the dev harness — both belong to the dev-environment e2e run (the remaining #1184 subtask), not a stock runner.

### Verified locally
`shellcheck` clean · `helm template` renders the probe Job as `<release>-egress-enforcement-check` · `helm unittest` 27 suites / 320 tests green. The live k3d run is exercised by the new CI job itself (like the other e2e scripts, it isn't runnable on a dev laptop without docker+k3d).

## Notes
- Targets `develop`.
- Advances backend#1184 (consolidate/execute the seal-check); the §8.3 matrix fill is the sibling [cli#449](https://github.com/tracebloc/cli/pull/449). Epic #1151 — Lukas drives, Saqlain reviews.

---

> [!NOTE]
> **Low Risk**
> Changes are CI, test automation, and a helm-unittest assertion; they do not alter runtime chart behavior or production install paths.
> 
> **Overview**
> Adds **live CI coverage** for the chart’s egress-enforcement `helm test` hook, closing the SEAL-CHECK gap where probes existed but **`helm test` never ran in the pipeline**.
> 
> A new **`scripts/tests/e2e-seal-check.sh`** brings up an isolated k3d cluster, installs the local chart with **`allowExternalHttps=false`** (and a longer enforcement probe timeout for slow CNI reconcile on GHA), runs a **positive control** curl from an unpolicy’d pod to `1.1.1.1`, then asserts the probe hook is present and runs **`helm test --filter`** on `<release>-egress-enforcement-check`, failing on exit code (not vacuous filter matches or unreliable `--logs`).
> 
> **`.github/workflows/helm-ci.yaml`** gains a **`seal-check-e2e`** job (30m timeout, no secrets). Path filters and **shellcheck** lists in **helm-ci**, **installer-tests**, and **standard-checks** include the new script (and `scripts/lib/**` for helm-ci triggers).
> 
> **`client/tests/egress_enforcement_check_test.yaml`** pins the probe Job **`metadata.name`** to `RELEASE-NAME-egress-enforcement-check` so the e2e filter cannot silently match nothing and pass.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 068bd9f207bc42504f334c7a82aa7212810f3260. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -233,6 +233,21 @@ jobs:
       - name: Upgrade from last published release through both flag paths
         run: bash scripts/tests/e2e-auto-upgrade.sh
 
+  seal-check-e2e:
+    timeout-minutes: 20
+    # RFC-0003 D12 (backend#1184): run the egress-enforcement seal-check LIVE on
+    # a real k3d cluster with the lockdown engaged, closing the SEAL-CHECK §8.4
+    # "full-chart probe run pending" gap (the k3s netpol substrate was verified
+    # in #504). Zero secrets — a public curl image vs 1.1.1.1; no private images,
+    # no backend. Proves the probe fires and the CNI actually blocks a
+    # training-labelled pod's egress once allowExternalHttps=false.
+    name: Seal-check egress-enforcement (k3d)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run the live egress-enforcement seal-check
+        run: bash scripts/tests/e2e-seal-check.sh
+
   # Installer script tests (bats + Pester) + the cross-distro prerequisite matrix
   # live in their own workflow: .github/workflows/installer-tests.yaml
   # (triggered on scripts/** changes).
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -236,7 +236,11 @@ jobs:
         run: bash scripts/tests/e2e-auto-upgrade.sh
 
   seal-check-e2e:
-    timeout-minutes: 20
+    # 30m to match the sibling k3d job (upgrade-e2e): the script's own bounds
+    # (create_cluster up to 15m + helm test 360s + tool install + install)
+    # can exceed a 20m GHA cap on a slow cluster bring-up and false-fail even
+    # while every component is still inside its own timeout (Bugbot).
+    timeout-minutes: 30
     # RFC-0003 D12 (backend#1184): run the egress-enforcement seal-check LIVE on
     # a real k3d cluster with the lockdown engaged, closing the SEAL-CHECK §8.4
     # "full-chart probe run pending" gap (the k3s netpol substrate was verified
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -7,6 +7,7 @@ on:
       - 'client/**'
       - 'ingestor/**'
       - 'scripts/tests/e2e-auto-upgrade.sh'
+      - 'scripts/tests/e2e-seal-check.sh'
       - '.github/workflows/helm-ci.yaml'
   pull_request:
     branches: [main, develop, openshift]
@@ -14,6 +15,7 @@ on:
       - 'client/**'
       - 'ingestor/**'
       - 'scripts/tests/e2e-auto-upgrade.sh'
+      - 'scripts/tests/e2e-seal-check.sh'
       - '.github/workflows/helm-ci.yaml'
 
 concurrency:
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -8,6 +8,7 @@ on:
       - 'ingestor/**'
       - 'scripts/tests/e2e-auto-upgrade.sh'
       - 'scripts/tests/e2e-seal-check.sh'
+      - 'scripts/lib/**'
       - '.github/workflows/helm-ci.yaml'
   pull_request:
     branches: [main, develop, openshift]
@@ -16,6 +17,7 @@ on:
       - 'ingestor/**'
       - 'scripts/tests/e2e-auto-upgrade.sh'
       - 'scripts/tests/e2e-seal-check.sh'
+      - 'scripts/lib/**'
       - '.github/workflows/helm-ci.yaml'
 
 concurrency:
--- a/.github/workflows/installer-tests.yaml
+++ b/.github/workflows/installer-tests.yaml
@@ -67,11 +67,11 @@ jobs:
           # below for visibility but don't fail the gate.
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
           echo "── shellcheck warnings (advisory, non-blocking) ──"
           shellcheck --severity=warning --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh || true
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh || true
 
       - name: Installer manifest is current (supply-chain, R8)
         # The bootstrap verifies each sub-script against scripts/manifest.sha256
--- a/scripts/tests/e2e-seal-check.sh
+++ b/scripts/tests/e2e-seal-check.sh
@@ -25,8 +25,11 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 LIB="$HERE/../lib"
 CHART_DIR="$HERE/../../client"
 
-# Isolated cluster + release so we never touch a real 'tracebloc' install.
+# Isolated cluster + release so we never touch a real 'tracebloc' install, and
+# opt out of autostart so create_cluster never reconfigures the host's Docker
+# restart policy / runs `systemctl enable docker` (matches the sibling e2e-*.sh).
 export CLUSTER_NAME="${CLUSTER_NAME:-tbseal}"
+export TRACEBLOC_NO_AUTOSTART=1
 NS="tbseal"
 
 # shellcheck source=/dev/null
--- a/scripts/tests/e2e-seal-check.sh
+++ b/scripts/tests/e2e-seal-check.sh
@@ -73,17 +73,26 @@ helm install "$NS" "$CHART_DIR" --namespace "$NS" --create-namespace \
 # exact name so the --filter below can never silently drift off it.
 PROBE="${NS}-egress-enforcement-check"
 
+# Guard the silent-pass trap FIRST: `helm test --filter` exits 0 when the filter
+# matches NOTHING (no test runs). So confirm the probe hook is actually in the
+# release before testing — a renamed/ungated probe fails here, loudly, instead
+# of passing vacuously.
+echo "── verify the probe hook is in the release ──"
+helm get hooks "$NS" --namespace "$NS" | grep -q "name: ${PROBE}" ||
+  fail "probe hook ${PROBE} not found in release hooks — --filter would match nothing"
+
 echo "── helm test --filter name=${PROBE} ──"
-# `helm test --filter` exits 0 when the filter matches NOTHING (no test runs) —
-# a silent false-pass. So success requires BOTH a zero exit AND the probe's own
-# marker in the logs; either one missing is a hard failure.
-LOG="$(mktemp)"
-if ! helm test "$NS" --namespace "$NS" --filter "name=${PROBE}" --logs >"$LOG" 2>&1; then
-  cat "$LOG"
+# Drive off the EXIT CODE, not --logs. The probe Job exits 0 ONLY when egress is
+# verified blocked, so a passing `helm test` == the lockdown is enforced. --logs
+# is unreliable for a Job-type hook: it looks up the pod by the Job's bare name
+# (the pod carries a generated suffix → "pods not found"), and hook-succeeded
+# deletes the Job on success anyway. On FAILURE the Job persists, so dump its
+# pod log for triage.
+if ! helm test "$NS" --namespace "$NS" --filter "name=${PROBE}" --timeout 360s; then
+  echo "── probe pod log (Job persists on failure) ──"
+  kubectl logs -n "$NS" -l "job-name=${PROBE}" --tail=-1 2>/dev/null ||
+    kubectl describe job -n "$NS" "${PROBE}" 2>/dev/null || true
   fail "helm test reported failure for ${PROBE} — egress lockdown NOT verified"
 fi
-cat "$LOG"
-grep -q "OK  egress lockdown verified" "$LOG" ||
-  fail "no egress-enforcement success marker in helm-test logs — did --filter match the probe? (guards the filter-silent-pass trap)"
 
-echo "PASS: live seal-check — egress-enforcement probe fired and verified the lockdown."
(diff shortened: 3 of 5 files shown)


scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,105 @@
...
+  fail "probe hook ${PROBE} not found in release hooks — --filter would match nothing"
+
+echo "── helm test --filter name=${PROBE} ──"
+# Drive off the EXIT CODE, not --logs. The probe Job exits 0 ONLY when egress is
+# verified blocked, so a passing `helm test` == the lockdown is enforced. --logs
+# is unreliable for a Job-type hook: it looks up the pod by the Job's bare name
+# (the pod carries a generated suffix → "pods not found"), and hook-succeeded
+# deletes the Job on success anyway. On FAILURE the Job persists, so dump its
+# pod log for triage.
+if ! helm test "$NS" --namespace "$NS" --filter "name=${PROBE}" --timeout 360s; then
```
saqlainsyed007, line 140: **No positive control — a pass isn't attributable to the NetworkPolicy.**

The test treats egress being blocked *for any reason* as proof the CNI enforced the lockdown. If direct TCP to `1.1.1.1:443` is unreachable from the runner for a reason unrelated to the chart — a GHA egress restriction/firewall, a Cloudflare rate-limit, a transient outage — curl returns exit 7/28, the probe prints `OK  egress lockdown verified`, the Job exits 0, and `helm test` passes green even though the training-egress NetworkPolicy did nothing.

The script never first establishes that egress *succeeds* without the lockdown, so the central claim ("the CNI actually blocks it") can be satisfied vacuously. Consider a baseline step: install/probe once with `allowExternalHttps=true` and assert egress **succeeds**, then flip to `false` and assert it's **blocked** — only the delta proves enforcement. (Alternatively probe from a non-`training`-labelled pod as the control.)

scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,105 @@
...
+# Local working-tree chart, public images (no registry secret), local-path
+# storage, and allowExternalHttps=false so the training-egress NetworkPolicy is
+# rendered and the egress-enforcement-check Job renders (it is gated on the
+# lockdown flag + a non-empty enforcementProbeHost, chart default 1.1.1.1).
+helm install "$NS" "$CHART_DIR" --namespace "$NS" --create-namespace \
+  -f "$CHART_DIR/tests/values-public-images.yaml" \
+  --set clientId=ci-e2e-seal \
+  --set clientPassword=ci-e2e-seal \
+  --set storageClass.provisioner=rancher.io/local-path \
+  --set networkPolicy.training.allowExternalHttps=false
```
saqlainsyed007: **Probe timeout left at the 60s chart default → flaky false-fail.**

`helm test` gets `--timeout 360s`, but the probe's own retry deadline (`enforcementProbeTimeoutSeconds`) stays at the chart default of 60s. On a cold, CPU-starved GHA runner, k3s's kube-router controller can take longer than 60s to program the pod's per-pod iptables rules while the whole chart is still installing. The probe (`backoffLimit: 0`, single attempt) then reports `NOT ENFORCED` and the job goes red on a cluster that *does* enforce — a false-fail well inside the 360s outer budget that was sized for exactly this bring-up.

Suggest `--set networkPolicy.training.enforcementProbeTimeoutSeconds=...` here to align the inner deadline with the outer `helm test` budget.

.github/workflows/helm-ci.yaml
```diff
@@ -7,13 +7,15 @@ on:
...
       - 'scripts/tests/e2e-auto-upgrade.sh'
+      - 'scripts/tests/e2e-seal-check.sh'
       - '.github/workflows/helm-ci.yaml'
   pull_request:
     branches: [main, develop, openshift]
     paths:
       - 'client/**'
       - 'ingestor/**'
       - 'scripts/tests/e2e-auto-upgrade.sh'
+      - 'scripts/tests/e2e-seal-check.sh'
```
saqlainsyed007, line 19: **Trigger paths omit `scripts/lib/**`, which this job depends on.**

`e2e-seal-check.sh` sources `scripts/lib/{common,setup-linux,cluster,preflight}.sh`, but the workflow's `paths:` filters cover only `client/**`, `ingestor/**`, the two e2e scripts, and this workflow file. A PR that edits only `scripts/lib/cluster.sh` (e.g. changes `create_cluster`'s k3d args) matches none of these, so neither `seal-check-e2e` nor `upgrade-e2e` runs — the very jobs that would catch the regression.

The trigger set was extended for the new script but not for its transitive deps. Adding `scripts/lib/**` closes the gap for both k3d jobs (pre-existing for `upgrade-e2e` too).

scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,105 @@
...
+
+# Isolated cluster + release so we never touch a real 'tracebloc' install, and
+# opt out of autostart so create_cluster never reconfigures the host's Docker
+# restart policy / runs `systemctl enable docker` (matches the sibling e2e-*.sh).
+export CLUSTER_NAME="${CLUSTER_NAME:-tbseal}"
+export TRACEBLOC_NO_AUTOSTART=1
+NS="tbseal"
+
+# shellcheck source=/dev/null
+source "$LIB/common.sh"
```
saqlainsyed007, line 39: **Prelude copy-pasted from `e2e-auto-upgrade.sh`.**

The `HERE`/`LIB`/`CHART_DIR` setup, the four shellcheck-directive `source` lines, the `cleanup`/`trap`, `fail()`, and the `install_kubectl`/`install_k3d`/`install_helm` block are duplicated near-verbatim across the `e2e-*.sh` scripts. The comments here already cite multiple Bugbot rounds where this bring-up sequence changed — every copy has to be edited in lockstep or they silently drift. A shared `scripts/tests/lib/e2e-common.sh` sourced by all of them would single-source the contract. (Non-blocking, but this PR adds a fresh copy.)

scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,105 @@
...
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+LIB="$HERE/../lib"
+CHART_DIR="$HERE/../../client"
+
+# Isolated cluster + release so we never touch a real 'tracebloc' install, and
+# opt out of autostart so create_cluster never reconfigures the host's Docker
+# restart policy / runs `systemctl enable docker` (matches the sibling e2e-*.sh).
+export CLUSTER_NAME="${CLUSTER_NAME:-tbseal}"
+export TRACEBLOC_NO_AUTOSTART=1
+NS="tbseal"
```
saqlainsyed007: **`CLUSTER_NAME` is env-overridable but `NS` is hardcoded — they can desync.**

`CLUSTER_NAME` defaults via `${CLUSTER_NAME:-tbseal}` while `NS="tbseal"` is fixed. Running `CLUSTER_NAME=foo bash scripts/tests/e2e-seal-check.sh` to isolate a second run makes the k3d cluster `foo` while the release/namespace stay `tbseal`, splitting the run's identity across two names and defeating the one-knob isolation. Either derive `NS="$CLUSTER_NAME"` or drop the override on `CLUSTER_NAME`.

saqlainsyed007 (changes requested):
Code review — the mechanism is well thought through (the `helm test --filter` silent-pass trap is genuinely closed by the hook-existence guard + the name-pin unittest, and driving off exit code rather than `--logs` is the right call for a Job hook). A few things worth addressing before this becomes a trusted gate, the first being a real soundness gap.

Also cleared while reviewing: the missing `export USER` vs. the sibling script is safe (every `$USER` ref in the sourced libs is guarded), and the new script correctly does **not** need a `scripts/manifest.sha256` entry (test scripts aren't in `gen-manifest.sh`'s FILES list).

One doc nit: the PR description says the script requires *both* a zero exit **and** the `OK  egress lockdown verified` log marker — but the implemented script drives off the exit code only (its comments correctly explain why). The code is internally consistent; the description overstates it.

scripts/tests/e2e-seal-check.sh
LukasWodka: Fixed (blocking) — added a **positive control**: before the probe runs, a pod in `default` (ungoverned by the namespace-scoped training-egress policy) must reach `1.1.1.1:443`. If it can't, the seal-check fails as *inconclusive* rather than green. A reachable positive + a blocked training pod ⇒ the block is attributable to the policy, not a runner firewall / target outage.

scripts/tests/e2e-seal-check.sh
LukasWodka: Fixed (blocking) — install now sets `enforcementProbeTimeoutSeconds=240` (from the 60s chart default). 240s is well inside the 360s helm-test budget, so a slow kube-router reconcile on a cold runner no longer false-fails the single-shot probe.

.github/workflows/helm-ci.yaml
LukasWodka: Fixed (blocking) — added `scripts/lib/**` to the helm-ci `paths` (both push + pull_request). A lib-only edit now re-triggers seal-check-e2e **and** upgrade-e2e, which both source those libs.

scripts/tests/e2e-seal-check.sh
LukasWodka: Fixed — `NS` now derives from `CLUSTER_NAME` (`NS="$CLUSTER_NAME"`), so a `CLUSTER_NAME=foo` override moves cluster + release + namespace together under one name.

scripts/tests/e2e-seal-check.sh
LukasWodka: Good call — the bring-up prelude is duplicated across the e2e-*.sh scripts and drifts. That's a cross-script refactor (a shared `scripts/tests/lib/e2e-common.sh`) beyond this PR's scope; I've flagged it as a follow-up so all copies get single-sourced in one change rather than expanding the churn here. (Non-blocking, as you noted.)

.github/workflows/installer-tests.yaml
```diff
@@ -67,11 +67,11 @@ jobs:
           # below for visibility but don't fail the gate.
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
```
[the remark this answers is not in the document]
LukasWodka: Fixed — added `scripts/tests/e2e-seal-check.sh` to the `standard-checks.yml` Lint shellcheck list (the required branch-protection gate), so a regression in the new script is caught there, not just in installer-tests.

scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,140 @@
...
+# did nothing. A pod in `default` is governed by NO training-egress policy (the
+# policy is namespace-scoped to the release ns), so if IT reaches the host, the
+# training pod's block below is attributable to the policy, not the environment.
+# Same image + curl invocation as the probe; HOST matches the chart default the
+# install leaves in place.
+HOST=1.1.1.1
+echo "── positive control: a non-policied pod must REACH ${HOST}:443 ──"
+kubectl run seal-poscheck --namespace default --restart=Never \
+  --image="curlimages/curl:8.20.0" \
+  --command -- curl --noproxy '*' -k -sS -m 15 -o /dev/null "https://${HOST}"
```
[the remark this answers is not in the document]
LukasWodka: Fixed — the positive control now waits for the `default` ServiceAccount to exist (poll up to 20s) before `kubectl run`, so a fast-runner race can't abort under set -e with 'serviceaccount default not found' before the attribution message.

scripts/tests/e2e-seal-check.sh
```diff
@@ -0,0 +1,147 @@
...
+# Positive control (Saqlain review): before trusting a BLOCKED probe result,
+# prove the cluster can actually REACH the probe host. Otherwise egress failing
+# for an unrelated reason (a runner firewall, a target outage, a rate-limit)
+# makes the probe print OK and the seal-check pass green while the NetworkPolicy
+# did nothing. A pod in `default` is governed by NO training-egress policy (the
+# policy is namespace-scoped to the release ns), so if IT reaches the host, the
+# training pod's block below is attributable to the policy, not the environment.
+# Same image + curl invocation as the probe; HOST matches the chart default the
+# install leaves in place.
+HOST=1.1.1.1
```
saqlainsyed007, line 92: Nit (non-blocking): the positive control hardcodes `HOST=1.1.1.1` while the probe uses the chart's `enforcementProbeHost` default (also 1.1.1.1) — an implicit coupling. The install here doesn't override `enforcementProbeHost`, so they match today, but if that default ever changes (or someone threads an override through this script) the control would test a *different* host than the probe and silently stop being an attribution for the probe's target. Cheap hardening: derive `HOST` from the same value you'd pass to `--set networkPolicy.training.enforcementProbeHost`, or add a comment-enforced single source. Fine to merge as-is.

saqlainsyed007 (approved):
Approving — round-2 asks are all resolved and the seal-check is now sound.

I re-verified the mechanism end-to-end against the chart, not just the diff:

- **The probe fails loudly when egress is NOT sealed.** `templates/egress-enforcement-check.yaml` exits 0 *only* on curl exit 7/28 (TCP connect refused/timed out), retries up to the timeout, and exits 1 with a loud banner if the connect ever succeeds (rc=0 or any TLS-layer code 35/52/56/60, all latched via `reached=1`) or if every attempt is DNS-only (exit 6). No swallowed timeout, no `|| true`, no grep-matches-nothing pass. The verdict is keyed on curl's exit code, not HTTP status.
- **No false-green path.** For the gate to pass green, the positive control must Succeed AND `helm test` must exit 0; the probe Job exits 0 only when egress is blocked. An open egress makes the probe exit 1. Image-pull / timeout / DNS failures all fail *safe* (inconclusive), never green.
- **The `helm test --filter` silent-pass trap is genuinely closed** by the `helm get hooks` existence guard (now a here-string, no SIGPIPE under pipefail) plus the name-pin unittest — belt and braces.
- **Netpol scoping confirms attribution.** `network-policy-training.yaml` is namespace-scoped to the release ns and selects `tracebloc.io/workload=training`; with `allowExternalHttps=false` rule 2 (0.0.0.0/0:443) is dropped, so the training pod's 1.1.1.1:443 is denied. The positive control runs in `default` (ungoverned) — a genuine, non-tautological control that can fail if the runner/target blocks egress, exactly ruling out the environmental false-block the prior review flagged. clusterCidrs defaults (10/8, 172.16/12, 192.168/16) cover k3d's 10.42/10.43 CIDRs, so the `fail` guard won't trip on the credential-free install.

Round-2 items — all landed: positive control (added), enforcementProbeTimeoutSeconds=240 inner deadline, scripts/lib/** trigger paths, NS derived from CLUSTER_NAME, TRACEBLOC_NO_AUTOSTART=1, SA-wait race guard, 30m job timeout, required standard-checks Lint + installer-tests shellcheck lists. Prelude duplication was correctly deferred as an out-of-scope follow-up.

One non-blocking nit inline. shellcheck is reported clean; `bash -n` passes here.

scripts/tests/e2e-seal-check.sh
LukasWodka: Done — extracted the shared prelude into `scripts/tests/lib/e2e-common.sh` in #550 (the 4 develop-side e2e scripts now source it). e2e-seal-check.sh adopts it as a fast-follow once #550 + this land (kept non-stacked).

scripts/tests/e2e-seal-check.sh
LukasWodka: Good catch — consistent today because the install leaves `enforcementProbeHost` at the chart default (1.1.1.1), which is exactly what the positive control targets. I'll harden it to derive HOST from one source so an override can't desync them — fast-follow (keeping your approval intact to land now).

PR #550, opened 2026-08-03
refactor(e2e): single-source the k3d bring-up prelude (#541 review)

LukasWodka:
## Single-source the e2e k3d bring-up prelude

Addresses @saqlainsyed007's review on [#541](https://github.com/tracebloc/client/pull/541) (thread on `scripts/tests/e2e-seal-check.sh:36`): the bring-up prelude was copy-pasted near-verbatim across `scripts/tests/e2e-*.sh`, and multiple Bugbot rounds had to edit every copy in lockstep or they'd drift.

### What's extracted → `scripts/tests/lib/e2e-common.sh`
Two functions — the parts that were **byte-identical** and drift-prone:
- **`e2e_isolate_env <name>`** — `USER` + `CLUSTER_NAME` default + `TRACEBLOC_NO_AUTOSTART` (the distinct default is the arg, still env-overridable).
- **`e2e_install_prereqs`** — `has docker` + `umask 022` + `install_{kubectl,k3d,helm}` (the sequence the Bugbot rounds kept changing).

`e2e-cluster` / `e2e-proxy` / `e2e-journey` / `e2e-auto-upgrade` now source the lib and call these; each keeps its own `CLUSTER_NAME` and test logic. `e2e-auto-upgrade` keeps its extra `has jq` guard right before the call.

### Deliberately **not** unified (unifying would change behavior)
- **The sub-lib `source` set** — `e2e-proxy`/`e2e-journey` source `common`+`setup-linux`+`cluster` only; `e2e-cluster`/`e2e-auto-upgrade` also source `preflight` (which has top-level `PF_*` side effects). ⚠️ *Both proxy and journey call `create_cluster` — which uses `_pf_recheck_runtime_mem`, defined only in `preflight.sh` — without sourcing it. That's a pre-existing inconsistency worth a separate look; I did not change it here.*
- **`cleanup`/`trap` bodies** — each reaps its own extra resources (a squid container, work dirs) beyond the k3d cluster.
- **`CHART_DIR`** — only the chart-installing scripts set it.

### Scope note
`e2e-seal-check.sh` lives on the still-open #541 branch, not `develop`, so it's **not** touched here (keeps this PR non-stacked). It adopts the lib as a fast-follow once both land.

### Verified
- Added the lib to **both** shellcheck gates: `installer-tests.yaml` and the **required** `standard-checks.yml` Lint.
- `shellcheck --severity=error` (the gate) **and** `--severity=warning` clean on all 5 files; `bash -n` ok.
- Net −20 lines of duplication. The live `e2e-cluster` / `e2e-proxy` / `e2e-journey` / `upgrade-e2e` CI jobs exercise the refactored scripts on this PR.

Targets `develop`.

---

> [!NOTE]
> **Low Risk**
> Test-harness refactor with no production installer behavior change; existing e2e CI jobs exercise the updated scripts.
> 
> **Overview**
> **Centralizes the duplicated k3d e2e bring-up prelude** in `scripts/tests/lib/e2e-common.sh`: `e2e_isolate_env` sets `USER`, a per-script default `CLUSTER_NAME`, and `TRACEBLOC_NO_AUTOSTART`; `e2e_install_prereqs` checks Docker, sets `umask`, and runs `install_kubectl` / `install_k3d` / `install_helm`.
> 
> `e2e-cluster`, `e2e-proxy`, `e2e-journey`, and `e2e-auto-upgrade` now source that lib instead of inlining the same blocks (each still passes its own cluster name; auto-upgrade keeps its separate `jq` check).
> 
> CI **Lint** and **installer-tests** shellcheck steps include the new file. `.gitignore` adds `!scripts/tests/lib/` so the test lib is not swallowed by the existing `lib/` ignore rule.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2b2be37afc04f2a8dbb384206060fa4fca782e5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/installer-tests.yaml
+++ b/.github/workflows/installer-tests.yaml
@@ -67,11 +67,11 @@ jobs:
           # below for visibility but don't fail the gate.
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh
           echo "── shellcheck warnings (advisory, non-blocking) ──"
           shellcheck --severity=warning --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh || true
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh || true
 
       - name: Installer manifest is current (supply-chain, R8)
         # The bootstrap verifies each sub-script against scripts/manifest.sha256
--- a/scripts/tests/e2e-cluster.sh
+++ b/scripts/tests/e2e-cluster.sh
@@ -21,11 +21,10 @@ set -euo pipefail
 HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 LIB="$HERE/../lib"
 
-# Isolated cluster name so we never touch a real 'tracebloc' cluster; opt out of
-# autostart so we don't reconfigure docker.service / restart policies on the host.
-export USER="${USER:-$(id -un)}"
-export CLUSTER_NAME="${CLUSTER_NAME:-tbe2e}"
-export TRACEBLOC_NO_AUTOSTART=1
+# Shared bring-up contract (isolation env + tool-install prereqs).
+# shellcheck source=/dev/null
+source "$HERE/lib/e2e-common.sh"
+e2e_isolate_env tbe2e
 
 # shellcheck source=/dev/null
 source "$LIB/common.sh"
@@ -45,11 +44,7 @@ echo "════════════════════════�
 
 # Docker is preinstalled + running on the runner; we only need the CLI tools the
 # cluster step uses. (We do NOT run install_docker_engine — no daemon gymnastics.)
-has docker || error "Docker is not available on this host."
-umask 022
-install_kubectl
-install_k3d
-install_helm
+e2e_install_prereqs
 
 echo "── create_cluster() — the installer's real cluster-bring-up path ──"
 create_cluster
--- a/.github/workflows/standard-checks.yml
+++ b/.github/workflows/standard-checks.yml
@@ -44,7 +44,7 @@ jobs:
           shellcheck --version | grep version
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/distro-prereqs.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-proxy.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh
+            scripts/tests/distro-prereqs.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-proxy.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/lib/e2e-common.sh
 
   unit-tests:
     name: Unit tests
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ eggs/
 .eggs/
 lib/
 !scripts/lib/
+!scripts/tests/lib/
 lib64/
 parts/
 sdist/
--- a/scripts/tests/e2e-auto-upgrade.sh
+++ b/scripts/tests/e2e-auto-upgrade.sh
@@ -46,11 +46,10 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 LIB="$HERE/../lib"
 CHART_DIR="$HERE/../../client"
 
-# Isolated cluster + release so we never touch a real 'tracebloc' install; opt
-# out of autostart so we don't reconfigure docker.service on the host.
-export USER="${USER:-$(id -un)}"
-export CLUSTER_NAME="${CLUSTER_NAME:-tbupg}"
-export TRACEBLOC_NO_AUTOSTART=1
+# Shared bring-up contract (isolation env + tool-install prereqs).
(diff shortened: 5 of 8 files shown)


scripts/tests/e2e-cluster.sh
```diff
@@ -21,11 +21,10 @@ set -euo pipefail
...
 LIB="$HERE/../lib"
 
-# Isolated cluster name so we never touch a real 'tracebloc' cluster; opt out of
-# autostart so we don't reconfigure docker.service / restart policies on the host.
-export USER="${USER:-$(id -un)}"
-export CLUSTER_NAME="${CLUSTER_NAME:-tbe2e}"
-export TRACEBLOC_NO_AUTOSTART=1
+# Shared bring-up contract (isolation env + tool-install prereqs).
+# shellcheck source=/dev/null
+source "$HERE/lib/e2e-common.sh"
```
saqlainsyed007, line 26: Blocking: this sources `scripts/tests/lib/e2e-common.sh`, but that file is **not in this commit** — `git show --stat f709559` adds no new file and `git cat-file -e HEAD:scripts/tests/lib/e2e-common.sh` reports it does not exist. It appears the newly-created lib was never `git add`ed.

With `set -euo pipefail` (line 19), `source` on a missing file returns non-zero and aborts the script here, before `create_cluster` — so the `e2e-cluster` job (and identically `e2e-proxy.sh:28`, `e2e-journey.sh:48`, `e2e-auto-upgrade.sh:51`) dies at bring-up. `e2e_isolate_env`/`e2e_install_prereqs` are also undefined, dropping the isolation env and the docker-check/umask/tool-install prereqs entirely. Please `git add` and push the lib.

.github/workflows/installer-tests.yaml
```diff
@@ -67,11 +67,11 @@ jobs:
           # below for visibility but don't fail the gate.
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/path-persist.sh
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh
```
saqlainsyed007: Same root cause: this adds `scripts/tests/lib/e2e-common.sh` to the shellcheck arg list (and again on line 74, and in `standard-checks.yml:47`), but the file is not committed in this PR. `shellcheck` on a nonexistent path errors (`openBinaryFile: does not exist`) and fails the gate — including the required standard-checks Lint. The PR's "shellcheck clean" claim only holds against the uncommitted local copy; commit the lib to make this real.

saqlainsyed007 (changes requested):
## Blocking: the extracted lib `scripts/tests/lib/e2e-common.sh` is never committed

The entire refactor pivots on a new shared file, but that file is **not part of this commit**. `git show --stat f709559` lists only the 6 modified files and adds nothing new; `git cat-file -e HEAD:scripts/tests/lib/e2e-common.sh` → *does not exist in HEAD*. There is no `e2e-common.sh` anywhere in the tree. This looks like a missing `git add` of the newly-created lib.

Consequences — this is not behavior-preserving, it is a hard break of every call site:

1. **Runtime abort in all 4 scripts.** `e2e-cluster.sh:26`, `e2e-proxy.sh:28`, `e2e-journey.sh:48`, `e2e-auto-upgrade.sh:51` each do `source "$HERE/lib/e2e-common.sh"`. Under the `set -euo pipefail` at the top of each script, sourcing a missing file fails immediately (`No such file or directory`, non-zero) and the script exits before ever reaching `create_cluster`. The `e2e-cluster` / `e2e-proxy` / `e2e-journey` / `upgrade-e2e` CI jobs all die at bring-up. Even if the source somehow continued, `e2e_isolate_env` and `e2e_install_prereqs` are undefined → the isolation env (`CLUSTER_NAME`/`USER`/`TRACEBLOC_NO_AUTOSTART`) and the docker-check/umask/tool-installs are silently lost, so the seal-check/egress posture the #541 review depends on is gone.
2. **Both shellcheck gates fail on a missing path.** `standard-checks.yml:47` and `installer-tests.yaml:70,74` now pass `scripts/tests/lib/e2e-common.sh` to `shellcheck`; with the file absent, shellcheck errors `openBinaryFile: does not exist`, breaking the required Lint. This contradicts the PR's "shellcheck clean on all 5 files" / "CI exercises the refactored scripts" claims — those can only have passed against an uncommitted local file.

**Fix:** `git add scripts/tests/lib/e2e-common.sh` and push. I could not review the actual behavior-preservation of `e2e_isolate_env` / `e2e_install_prereqs` (dropped flags, ordering, env defaults, `has docker` guard, `umask 022`) because the file isn't in the PR — please re-request review once the lib is committed so the extracted bodies can be diffed against the former inline copies.

**Verdict: REQUEST_CHANGES.**

scripts/tests/e2e-cluster.sh
LukasWodka: Fixed — you're exactly right, the lib was never committed. Root cause: `.gitignore` has `lib/` (a Python-build pattern) with only `!scripts/lib/` excepted, so `scripts/tests/lib/` was ignored and `git add -A` silently skipped the new file. Added `!scripts/tests/lib/` and committed the lib (`git cat-file -e HEAD:scripts/tests/lib/e2e-common.sh` now passes). Also rebased onto the post-#541 develop.

.github/workflows/installer-tests.yaml
LukasWodka: Same fix — the lib is now committed, so the shellcheck path resolves. Re-ran `shellcheck --severity=error` on the lib + all 4 scripts locally: clean. And the shellcheck lists now carry BOTH e2e-seal-check.sh (from #541, which merged) and lib/e2e-common.sh, resolved in the rebase.

saqlainsyed007 (approved):
**Re-review of a2b2be3** (prior REQUEST_CHANGES was review 4844149284 at f709559).

**Prior blocker RESOLVED.** The one root-cause finding — the extracted shared file `scripts/tests/lib/e2e-common.sh` was never `git add`ed — is fixed: the file is now committed and tracked in the tree at this SHA, appears in the PR file list (ADDED, +51), and `.gitignore` now carries `!scripts/tests/lib/` (mirroring the existing `!scripts/lib/`) so the dir is no longer swallowed by the `lib/` ignore. Every `source "$HERE/lib/e2e-common.sh"` now resolves, so `set -euo pipefail` no longer aborts the four callers, and both shellcheck lists reference a file that exists.

**Behavior preservation now verifiable — and faithful.**
- `e2e_isolate_env <name>` reproduces the old inline block exactly (`export USER="${USER:-$(id -un)}"`, `export CLUSTER_NAME="${CLUSTER_NAME:-$1}"`, `export TRACEBLOC_NO_AUTOSTART=1`). Each caller passes its own former default — e2e-cluster `tbe2e`, e2e-journey `tbe2e-journey`, e2e-proxy `tbproxy`, e2e-auto-upgrade `tbupg` — and env override is preserved.
- `e2e_install_prereqs` reproduces the old block byte-for-byte: `has docker || error`, `umask 022`, `install_kubectl`/`install_k3d`/`install_helm`.
- Sourcing/ordering is correct in all four: the lib is sourced first (functions only, no side effects, no `set -e` — right for a sourced lib), `e2e_isolate_env` runs, then common.sh + setup-linux.sh + cluster.sh are sourced, and `e2e_install_prereqs` is called only afterward, so `has`/`error`/`install_*` resolve at call time.
- Correctly NOT unified: `preflight.sh` is still sourced only by e2e-cluster + e2e-auto-upgrade (the pre-existing proxy/journey inconsistency is flagged in the lib header, not silently "fixed"); each script keeps its own `cleanup`/`trap cleanup EXIT`; CHART_DIR stays local to auto-upgrade.
- e2e-auto-upgrade keeps its extra `has jq || error` guard. (Minor, non-blocking: it now runs just before the docker check inside `e2e_install_prereqs` rather than just after it — both are fail-fast guards, so behavior is unchanged.)
- e2e-seal-check.sh is deliberately untouched (fast-follow per the description) and still carries its own inline prelude, so nothing is half-migrated or double-brought-up. The #541 seal-check/egress posture is unaffected.

**Verification I ran at this SHA:** `git cat-file -e HEAD:scripts/tests/lib/e2e-common.sh` (exists); `bash -n` clean on the lib + all four callers; `shellcheck --severity=error` **clean** across the lib, the four callers, and e2e-seal-check (the required standard-checks Lint gate), and `--severity=warning` clean too. Both workflow shellcheck lists now include `scripts/tests/lib/e2e-common.sh`.

No blocking issues and no behavior regressions. Approving.

PR #551, opened 2026-08-03
ci(windows-e2e): stop the teardown net failing open

LukasWodka:
The `always()` safety-net teardown ran `k3d cluster delete` via `Start-Process` with **no** `-ErrorAction Stop`, **no** null-check on `$p`, inside an **empty catch**, and the step never set `$ErrorActionPreference`.

So a **spawn failure** left `$p` null → `$p.HasExited` threw → the empty catch ate it → the shared runner kept a leftover `tbe2ewin` cluster. That is the exact state this net exists to clear, and the one that sends the next run's `New-K3dCluster` down its reuse path for a **false PASS** with no real create — the same false-green class as `#544`, one layer up in the workflow rather than the driver script.

**Fix:** `$ErrorActionPreference = 'Stop'`, explicit null-check, a timeout-kill that surfaces instead of hiding, and an exit-code check. On failure it does **not** fail the run — this is `always()` teardown and the run is already over — but emits a `::warning::` naming the stale cluster and the manual remedy, so a leftover can never be silent again.

Found by Bugbot on the release-train staging hop (`client#542`), High.

Refs: tracebloc/backend#1426

---

> [!NOTE]
> **Low Risk**
> CI-only PowerShell in an `always()` teardown step; behavior change is better observability and stricter delete validation without altering the main e2e driver or installer.
> 
> **Overview**
> The Windows e2e **`always()` teardown** step no longer **fails open** when `k3d cluster delete` does not actually run or finish.
> 
> It sets **`$ErrorActionPreference = 'Stop'`**, validates that **`Start-Process`** returned a process, treats **120s timeout + kill** and **non-zero exit** as failures instead of swallowing them, and replaces the **empty `catch`** with a **`::warning::`** that names **`tbe2ewin`**, the reuse-path false-pass risk, and the manual delete command. Successful deletes log confirmation; the step still **does not fail the job** on teardown errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9020bbbe0244091fe387ab53284282ce73519375. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/windows-e2e.yaml
+++ b/.github/workflows/windows-e2e.yaml
@@ -80,9 +80,28 @@ jobs:
           # net for a driver that died before its finally ran, plus the data-dir cleanup we
           # deliberately deferred so the log survived for the upload above. Bound the delete
           # so a wedged Docker/WSL teardown can't hang the shared runner (#436 Bugbot).
+          # The delete must not fail OPEN. This net exists precisely because a leftover
+          # tbe2ewin cluster on the shared runner sends the NEXT run's New-K3dCluster down
+          # its reuse path and yields a false PASS with no real create (#542 Bugbot). The
+          # old form swallowed everything in an empty catch: Start-Process without
+          # -ErrorAction Stop, and no null-check on $p, so a SPAWN failure left $p null,
+          # $p.HasExited threw, the catch ate it, and the stale cluster survived unseen.
+          $ErrorActionPreference = 'Stop'
+          $deleted = $false
           try {
             $p = Start-Process k3d -ArgumentList "cluster","delete",$env:CLUSTER_NAME -NoNewWindow -PassThru
+            if ($null -eq $p) { throw "k3d cluster delete did not start (Start-Process returned null)" }
             $p | Wait-Process -Timeout 120 -ErrorAction SilentlyContinue
-            if (-not $p.HasExited) { $p | Stop-Process -Force -ErrorAction SilentlyContinue }
-          } catch { }
+            if (-not $p.HasExited) {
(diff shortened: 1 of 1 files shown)


shujaatTracebloc (approved):
Reviewed and verified locally.

The rewrite fully closes the fail-open: with `$ErrorActionPreference = 'Stop'` plus the explicit `if ($null -eq $p) throw` guard, a Start-Process spawn failure can no longer leave $p null and blow up `$p.HasExited` inside an empty catch — the exact path that let a stale `tbe2ewin` cluster survive unseen. Timeout (force-kill → throw), non-zero exit code, and null-process are each surfaced; on any of them the always() step emits a `::warning::` naming the cluster and the manual `k3d cluster delete` remedy rather than swallowing it, and correctly does not fail the already-finished run.

Checked the edges:
- `k3d cluster delete` is idempotent (exit 0 when the cluster is absent, per #544), so a teardown running after the driver's own finally already deleted the cluster exits 0 → logs 'deleted', no spurious warning.
- `$ErrorActionPreference` is scoped to this run: step (fresh shell per step, no leak).
- The deferred data-dir `Remove-Item` stays outside the try/catch and still always runs; its per-command `-ErrorAction SilentlyContinue` overrides the Stop preference.

windows-e2e.yaml parses as valid YAML; teardown step is guarded by `if: always()`. All CI green incl. Cursor Bugbot; no open review threads. Correct that the heavy E2E/Prereqs jobs are path-filtered out for a workflow-only change. LGTM.

PR #566, opened 2026-08-03
chore(seal-check): adopt e2e-common lib + pin probe HOST (#541 fast-follow)

LukasWodka:
## #541 fast-follow (promised on review)

Now that #550 (shared `e2e-common.sh`) and #541 (the seal-check runner) are both on develop:

- **`e2e-seal-check.sh` adopts the shared lib** — sources `scripts/tests/lib/e2e-common.sh` and uses `e2e_isolate_env` / `e2e_install_prereqs` like the other `e2e-*.sh` (drops the inlined isolation-env + install block). Keeps its own `NS=$CLUSTER_NAME` and local `fail()`.
- **HOST hardening** (@saqlainsyed007 nit): the probe host is now a single `$HOST` var that the install pins via `--set networkPolicy.training.enforcementProbeHost="$HOST"` AND the positive control targets — so the enforcement probe and the positive control can never drift onto different hosts (was: `HOST` hardcoded while the probe used the chart default).

Verified: `shellcheck --severity=error/warning` clean; `helm template` renders the probe Job with `HOST=1.1.1.1` pinned + `TIMEOUT=240`. Targets `develop`.

---

> [!NOTE]
> **Low Risk**
> Test-harness refactor and explicit Helm value for probe host; no production runtime or auth changes.
> 
> **Overview**
> **`e2e-seal-check.sh`** now follows the same bring-up pattern as the other `e2e-*.sh` scripts by sourcing `scripts/tests/lib/e2e-common.sh` and calling **`e2e_isolate_env tbseal`** and **`e2e_install_prereqs`** instead of inlined cluster isolation and kubectl/k3d/helm install logic. Script-specific pieces (`NS=$CLUSTER_NAME`, local **`fail()`**, chart install, and seal-check flow) are unchanged.
> 
> The egress **probe host is a single `$HOST` (1.1.1.1)** passed to Helm via **`--set networkPolicy.training.enforcementProbeHost="$HOST"`**, so the chart enforcement Job and the **positive-control** curl in `default` always target the same host—avoiding drift between a hardcoded control and the chart default.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c263ac70c5b0ccdb6a9714319916c230eb64756e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/tests/e2e-seal-check.sh
+++ b/scripts/tests/e2e-seal-check.sh
@@ -25,14 +25,13 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 LIB="$HERE/../lib"
 CHART_DIR="$HERE/../../client"
 
-# Isolated cluster + release so we never touch a real 'tracebloc' install, and
-# opt out of autostart so create_cluster never reconfigures the host's Docker
-# restart policy / runs `systemctl enable docker` (matches the sibling e2e-*.sh).
-export CLUSTER_NAME="${CLUSTER_NAME:-tbseal}"
-export TRACEBLOC_NO_AUTOSTART=1
-# Derive NS from CLUSTER_NAME so `CLUSTER_NAME=foo bash ...` isolates a second
-# run under ONE name — cluster + release + namespace move together instead of
-# splitting the run's identity across two names (Saqlain review).
+# Shared bring-up contract (isolation env + tool-install prereqs), same as the
+# sibling e2e-*.sh.
+# shellcheck source=/dev/null
+source "$HERE/lib/e2e-common.sh"
+e2e_isolate_env tbseal
+# NS follows CLUSTER_NAME so a CLUSTER_NAME override isolates a whole run under
+# ONE name — cluster + release + namespace move together (Saqlain review).
 NS="$CLUSTER_NAME"
 
 # shellcheck source=/dev/null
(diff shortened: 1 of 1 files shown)


saadqbal (approved):

scripts/tests/e2e-seal-check.sh
```diff
@@ -25,14 +25,13 @@ HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
-# restart policy / runs `systemctl enable docker` (matches the sibling e2e-*.sh).
-export CLUSTER_NAME="${CLUSTER_NAME:-tbseal}"
-export TRACEBLOC_NO_AUTOSTART=1
-# Derive NS from CLUSTER_NAME so `CLUSTER_NAME=foo bash ...` isolates a second
-# run under ONE name — cluster + release + namespace move together instead of
-# splitting the run's identity across two names (Saqlain review).
+# Shared bring-up contract (isolation env + tool-install prereqs), same as the
+# sibling e2e-*.sh.
+# shellcheck source=/dev/null
+source "$HERE/lib/e2e-common.sh"
```
saqlainsyed007, line 31: Optional / non-blocking (pre-existing pattern, matches all sibling `e2e-*.sh`):

`e2e-common.sh` is sourced via a hardcoded `"$HERE/lib/e2e-common.sh"` (→ `scripts/tests/lib/`), while `LIB="$HERE/../lib"` (→ `scripts/lib/`) is the var used for `common.sh`/`setup-linux.sh`/`cluster.sh`/`preflight.sh`. So the script has two lib roots and only one is captured in a variable — a future relocation of the helpers could update `LIB` and miss this line (or vice-versa). Not worth churning here given it mirrors the siblings, just noting it.

saqlainsyed007 (commented):
Reviewed at high effort (8 finder angles + cross-file/removed-behavior trace). This holds up well — clean, well-scoped fast-follow.

**What I verified:**
- The deleted inlined isolation + prereq block is faithfully reproduced by `e2e_isolate_env` / `e2e_install_prereqs`; the only net changes are the intentional `export USER` and `fail`→`error` for the docker check (both `exit 1`, both matching the sibling `e2e-*.sh`).
- Sourcing contract respected: `e2e_install_prereqs` (resolves `has`/`error`/`install_*` at call time) runs *after* `common.sh` + `setup-linux.sh` are sourced, identical ordering to `e2e-cluster.sh`.
- The `HOST` pinning does what it claims — `--set …enforcementProbeHost="$HOST"` and the positive control both derive from the single `HOST=1.1.1.1`, closing the hardcoded-vs-chart-default drift.
- `e2e-common.sh` itself is untouched, so no other consumer of the shared helpers is affected.

No correctness issues. One optional maintainability nit inline — non-blocking; LGTM.

saqlainsyed007 (approved):
LGTM — clean fast-follow. Verified the shared-lib adoption preserves the inlined behavior (only the intentional `export USER` and `fail`→`error` deltas), the sourcing/call ordering matches the sibling e2e scripts, and the single `HOST=1.1.1.1` now feeds both the Helm `--set` and the positive control so the probe/control can no longer drift. The one inline note is optional/non-blocking.

PR #570, opened 2026-08-03
fix(chart): requests-proxy honours a pinned digest for imagePullPolicy (#552)

shujaatTracebloc:
## What & why

Narrow, safe part of #552 (control-plane restart resilience). `requests-proxy` **hardcoded `imagePullPolicy: Always`** even though its image line already read `images.requestsProxy.digest` — so a pinned digest was silently ignored for the pull policy, and every pod restart re-pulled the image even when it was already cached in containerd. That's the same offline-restart fragility class as the incident, and it's an outright inconsistency bug (the other control-plane deployments are already digest-aware).

## Change

`client/templates/requests-proxy-deployment.yaml` — pull policy now mirrors jobs-manager / pods-monitor / resource-monitor:

- `images.requestsProxy.digest` **set** → `repo@digest` + `IfNotPresent` (restart-safe offline)
- `images.requestsProxy.digest` **empty** (default) → `repo:tag` + `Always` (**unchanged** default behaviour)

Verified by render: default → `docker.io/tracebloc/jobs-manager:prod` + `Always`; with a digest → `…@sha256:…` + `IfNotPresent`.

## Tests

`client/tests/requests_proxy_test.yaml` — **+2** helm-unittest cases: default floats on the tag with `Always`; a set digest renders `repo@digest` + `IfNotPresent`. Suite green (13/13).

`client/Chart.yaml` 1.9.11 → **1.9.12** (chart-content change requires a bump).

## Scope

This is deliberately only the requests-proxy bug. The **broader** offline-restart update-model change for the always-running images — `jobs-manager`/`pods-monitor` are updated by the image-refresh CronJob via `rollout restart`, which *relies* on `Always`, and `resource-monitor` isn't refreshed at all — is a real design decision (digest-pin à la `ingestor.prodDigest`, vs. digest-on-update in image-refresh) and is tracked in **#569**.

Part of #552.

---

> [!NOTE]
> **Low Risk**
> Narrow Helm template fix with unchanged default behaviour when no digest is set; covered by new unit tests.
> 
> **Overview**
> **requests-proxy** no longer hardcodes `imagePullPolicy: Always` when `images.requestsProxy.digest` is set. Pull policy now matches jobs-manager, pods-monitor, and resource-monitor: pinned digest → `repo@digest` + `IfNotPresent` (restart without registry); no digest → tag + `Always` (unchanged default).
> 
> Helm unittest adds two cases for default vs digest-pinned rendering. Chart version **1.9.11 → 1.9.12**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97611facf42b6c3ba6aef37bc6ed929edacc247d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.11
-appVersion: "1.9.11"
+version: 1.9.12
+appVersion: "1.9.12"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/templates/requests-proxy-deployment.yaml
+++ b/client/templates/requests-proxy-deployment.yaml
@@ -22,8 +22,12 @@ spec:
           type: RuntimeDefault
       containers:
         - name: proxy
-          image: {{ include "tracebloc.image" (dict "repository" "tracebloc/jobs-manager" "tag" .Values.env.CLIENT_ENV "digest" (dig "requestsProxy" "digest" "" .Values.images) "registry" "docker.io") | quote }}
-          imagePullPolicy: Always
+          {{- $rpDigest := (dig "requestsProxy" "digest" "" .Values.images) }}
+          image: {{ include "tracebloc.image" (dict "repository" "tracebloc/jobs-manager" "tag" .Values.env.CLIENT_ENV "digest" $rpDigest "registry" "docker.io") | quote }}
+          # digest set -> repo@digest + IfNotPresent (restart-safe offline); empty ->
+          # repo:tag + Always. Was hardcoded Always, which ignored a pinned digest and
+          # re-pulled on every restart even when the image was already cached (#552).
+          imagePullPolicy: {{ if $rpDigest }}IfNotPresent{{ else }}Always{{ end }}
           workingDir: /app
           command: ["python", "-m", "gunicorn"]
           args:
--- a/client/tests/requests_proxy_test.yaml
+++ b/client/tests/requests_proxy_test.yaml
@@ -79,6 +79,30 @@ tests:
           path: spec.template.spec.containers[0].image
           pattern: "^docker\\.io/tracebloc/jobs-manager[:@]"
 
+  - it: floats on the tag with imagePullPolicy Always when no digest is pinned (default)
+    asserts:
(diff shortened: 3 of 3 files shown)


aptracebloc (approved):
@shujaatTracebloc — **Approved.** Clean parity fix. What I verified:

- **✅ requests-proxy was the sole outlier.** The other first-party floating-tag deployments already use the identical `{{ if digest }}IfNotPresent{{ else }}Always{{ end }}` idiom — jobs-manager (`.Values.images.jobsManager.digest`), pods-monitor, and resource-monitor all match. This brings requests-proxy into line. The two hardcoded-`IfNotPresent` templates (egress-proxy/squid, mysql-client) are correctly-pinned third-party images, not the same bug — so no other deployment carries this latent issue.
- **✅ Live-settable, not inert** — `images.requestsProxy.digest` is in `values.schema.json` and the unittest's `set:` exercises it end-to-end (renders `repo@digest` + `IfNotPresent`); the `dig "requestsProxy" "digest" ""` accessor is the safest null-nav form.
- **✅ Chart version bumped** 1.9.11 → 1.9.12 (correct Helm convention), and both branches (tag→Always, digest→IfNotPresent) are pinned by the new helm-unittest cases (CI-run).
- No R8 manifest concern — this is chart territory, not the installer sub-scripts.

(Verification note: my local `helm template --show-only` didn't emit — a missing-value/path quirk, not a fix defect, since the template is ungated and the unittest renders it cleanly. Relied on the source-parity check + the CI unittest cases + the schema check.)

## ⚪ Nits (non-blocking)
- The chart now has three idioms for the same digest null-navigation: `dig …` (this PR, safest), resource-monitor's `default (dict)` chaining, and jobs-manager's direct `.Values.images.jobsManager.digest` (which would error if `images.jobsManager` were absent). Not this PR's problem — it picked the safest — but they'd read more consistently converged.
- Worth flagging for context: requests-proxy runs the *jobs-manager image* (shared, different command) pinned via its own `requestsProxy.digest` key, so it can be pinned independently of `jobsManager.digest` — and the fix correctly honors the proxy's own key.

Minimal, correct, closes a real "re-pull on every restart / not offline-safe when pinned" gap. Touches your code-owned `client/templates/`, so Asad's code-owner review is the backstop; this is the second set of eyes.

PR #591, opened 2026-08-04
fix(installer): pin -Encoding UTF8 on install-log writes

LukasWodka:
Bugbot flagged this on the develop→staging promotion (client#589). `Set-Content`/`Add-Content` for the curated installer log default to PowerShell 5.1's ANSI encoding, which corrupts non-ASCII host paths/messages in the log operators send with `-Diagnose`. Every other writer in `install-k8s.ps1` already pins `-Encoding UTF8`; this makes the two log writers (L161 Add-Content, L590 Set-Content) consistent. No behaviour change otherwise.

---

> [!NOTE]
> **Low Risk**
> Installer logging and diagnose packaging only; no cluster install or auth behavior changes.
> 
> **Overview**
> Fixes **mojibake** in the Windows installer’s curated log and `-Diagnose` support bundle when paths or messages contain non-ASCII characters (#591).
> 
> **`Log()`** now appends with **`-Encoding UTF8`**. The install log header is created via **`WriteAllText` with UTF-8 no BOM** instead of `Set-Content -Encoding UTF8`, which would prepend a BOM on PowerShell 5.1. **`Edit-Redaction`** reads collected files with **`-Encoding UTF8`** so BOM-less logs decode correctly on PS 5.1. **`scripts/manifest.sha256`** is updated for `install-k8s.ps1`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4b58a74632d921cf12a08074d4e9aab7573543d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/install-k8s.ps1
```diff
@@ -618,7 +618,7 @@ function Start-InstallLog {
   # on-screen output: no user PII, no tracebloc internals. Best-effort — if the
   # file can't be created, logging silently no-ops and the install continues.
   try {
-    Set-Content -Path $LOG_FILE -Value "tracebloc client installer log - $(Get-Date -Format 'yyyy-MM-dd HH:mm:ss')" -ErrorAction Stop
+    Set-Content -Path $LOG_FILE -Value "tracebloc client installer log - $(Get-Date -Format 'yyyy-MM-dd HH:mm:ss')" -Encoding UTF8 -ErrorAction Stop
```
saqlainsyed007: Non-blocking: on Windows PowerShell 5.1 (the env this PR targets) `-Encoding UTF8` emits a UTF-8 **BOM**, so the raw `install-*.log` now starts with `EF BB BF` (this `Set-Content` creates the file with the BOM; the `Add-Content` at L162 then appends without re-adding it). It's harmless in the `-Diagnose` bundle since `Edit-Redaction` reads via `Get-Content -Raw` and rewrites no-BOM, but it diverges from the file's own no-BOM convention (`New-Object System.Text.UTF8Encoding($false)` at L1780/L1829/L4225). If you want a clean UTF-8 log, the no-BOM writer used elsewhere would be more consistent — but this is purely cosmetic, not a blocker.

saqlainsyed007 (dismissed):
Approving — clean, low-risk fix. I recomputed the SHA-256 of the new `install-k8s.ps1` and it matches the updated `manifest.sha256` line exactly, and the old hash isn't pinned anywhere else, so the bootstrap integrity check stays green. The ANSI→UTF-8 change correctly fixes the non-ASCII corruption and is consistent with the file's other `-Encoding UTF8` writers.

One non-blocking note left inline.

scripts/install-k8s.ps1
LukasWodka: Good catch — fixed in 235fca6. The log-init `Set-Content` now uses the file's own no-BOM writer (`[System.IO.File]::WriteAllText($LOG_FILE, …, (New-Object System.Text.UTF8Encoding($false)))`, matching L1780/L1829/L4226), so the file no longer starts with `EF BB BF`. The `Add-Content` append path is unchanged since — as you noted — it appends UTF-8 without re-adding a BOM to the existing file. Thanks for the review.

shujaatTracebloc (approved):

PR #599, opened 2026-08-05
feat(installer): offline Sigstore bundle — verify --offline, no live Rekor (#584)

shujaatTracebloc:
## What & why

Closes #584 (child 3/4 of #578, network resilience). **This is the item that actually fixes the field Windows install failure on a TLS-inspecting corporate network.**

Keyless cosign verification needs sigstore's transparency log (Rekor) at verify time, and our short-lived keyless cert is **expired by install time** — so a network that blocks or TLS-inspects sigstore fails verification even for a *valid* signature. #583's CA wiring doesn't cover this on Windows/macOS (Go ignores `SSL_CERT_FILE` there), and skipping the tlog was proven not to verify the expired cert. The fix is an **offline Sigstore bundle**: it carries the Rekor inclusion proof (SET), so the cert's validity at signing is established with **no live call**.

## Changes

**Signing** — `release-helm-chart.yaml`
- `cosign sign-blob` now also emits `manifest.sha256.bundle` and publishes it as a release asset, alongside the existing `.sig`/`.cert`.

**Verify** — `install.sh` + `install.ps1`
- Prefer `cosign verify-blob --bundle <bundle> --certificate-identity-regexp … --certificate-oidc-issuer … --offline` — a full check (signature + cert identity + tlog inclusion) with **no live Rekor**.
- **Fallback:** if the bundle 404s (older release) *or* doesn't verify, fall through to the existing online `.sig`/`.cert` keyless path. That path does the **same full keyless verification**, just needing live Rekor — so it's a fallback, never a security downgrade.
- PS: extracted **`Invoke-CosignVerifyBlob`** so the fail-closed sentinel (nonzero `$LASTEXITCODE` seed) + stderr suppression (#576) back **both** paths from one place.

## Tests
- **bats:** bootstrap prefers `--bundle --offline` when a bundle is published (and does *not* touch the sig/cert path); falls back to sig/cert when no bundle exists (older release); cosign-failure still fails closed (unchanged).
- **Pester:** `Confirm-ManifestSignature` verifies `--bundle --offline` first with a sig/cert fallback; `Invoke-CosignVerifyBlob` is fail-closed (sentinel + stderr suppression).
- Full Pester 449/0/9; guards + manifest `--check` clean. `install.sh`/`install.ps1` are the bootstrap trust root (not manifested), so no manifest change.

## ⚠️ Sequencing (cross-repo-ish, but all in this repo)
The installer prefers the bundle but **a new release must be cut after merge** so the `manifest.sha256.bundle` asset actually exists. Until then — and for all existing releases — the **fallback** (online sig/cert) path runs, so nothing regresses. Once a release ships the bundle, sigstore-blocked networks (incl. the field Windows case) verify fully offline with no env var or user action.

## Notes
- The `--insecure-ignore-tlog` shortcut from the earlier cosign thread was proven *not* to verify our keyless cert — this bundle approach is the correct version.
- Remaining in the track after this: #585 (registry mirror / air-gap) and #581 (internal-identifier sweep).

---

> [!NOTE]
> **Medium Risk**
> Changes the bootstrap trust root and release signing assets; behavior is backward-compatible via fallback, but a bad bundle rollout could affect installs until the next release ships bundles correctly.
> 
> **Overview**
> **Adds offline manifest verification** so installs succeed on networks that block or TLS-inspect Sigstore/Rekor, and so expired short-lived keyless certs can still be validated using the bundle’s embedded signing-time proof.
> 
> The **release workflow** now emits and publishes `manifest.sha256.bundle` alongside the existing `.sig`/`.cert` when signing `manifest.sha256`.
> 
> **`install.sh` and `install.ps1`** try `cosign verify-blob --bundle … --offline` first (same identity pins as before). If the bundle is missing (older releases) or verification fails, they **fall back** to the existing online `.sig`/`.cert` path—same keyless checks, not a weaker mode. On Windows, **`Invoke-CosignVerifyBlob`** centralizes fail-closed `$LASTEXITCODE` seeding and stderr suppression for both paths.
> 
> **Tests** (bats + Pester) cover bundle-first success, sig/cert fallback, older releases without a bundle, and fail-closed when both paths fail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5669847bd8d9d80b09ff01ad352c0060eb0c44c3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -435,6 +435,11 @@ verify_manifest_signature() {
   local manifest="$1"
   local sig="$TMPDIR/manifest.sha256.sig"
   local cert="$TMPDIR/manifest.sha256.cert"
+  local bundle="$TMPDIR/manifest.sha256.bundle"
+  # The keyless signing identity: the release workflow's OIDC cert. Shared by both
+  # the offline-bundle and the online sig/cert verification paths below.
+  local id_re='https://github.com/tracebloc/client/\.github/workflows/.*@.*'
+  local issuer='https://token.actions.githubusercontent.com'
 
   if ! ensure_cosign; then
     if [[ "$ALLOW_UNVERIFIED" == "1" ]]; then
@@ -450,6 +455,26 @@ verify_manifest_signature() {
     exit 1
   fi
 
+  # OFFLINE Sigstore bundle first (#584): the bundle carries the Rekor inclusion
+  # proof, so this verifies signature + cert identity + tlog inclusion with NO live
+  # Rekor call — immune to networks that block/TLS-inspect sigstore, and the ONLY
+  # path that verifies our short-lived keyless cert once it has expired (its embedded
+  # timestamp proves the cert was valid at signing). Releases cut before the bundle
+  # existed simply 404 here and fall through to the online .sig/.cert path below;
+  # so does any bundle that doesn't verify — the online path does the SAME full
+  # keyless check, just needing live Rekor, so this is a fallback, never a downgrade.
+  if curl -fsSL --tlsv1.2 --connect-timeout 30 --max-time 300 "$REPO_REL/manifest.sha256.bundle" -o "$bundle" 2>/dev/null; then
+    if "$COSIGN_BIN" verify-blob \
+          --bundle "$bundle" \
+          --certificate-identity-regexp "$id_re" \
+          --certificate-oidc-issuer "$issuer" \
+          --offline \
+          "$manifest" >/dev/null 2>&1; then
+      printf '  %s✔%s Download verified as published by tracebloc\n' "$_G" "$_R"
+      return 0
+    fi
+  fi
+
   if ! curl -fsSL --tlsv1.2 --connect-timeout 30 --max-time 300 "$REPO_REL/manifest.sha256.sig"  -o "$sig"  2>/dev/null \
      || ! curl -fsSL --tlsv1.2 --connect-timeout 30 --max-time 300 "$REPO_REL/manifest.sha256.cert" -o "$cert" 2>/dev/null; then
     if [[ "$ALLOW_UNVERIFIED" == "1" ]]; then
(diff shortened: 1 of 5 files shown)


scripts/install.sh
```diff
@@ -450,6 +455,26 @@ verify_manifest_signature() {
...
+  # OFFLINE Sigstore bundle first (#584): the bundle carries the Rekor inclusion
+  # proof, so this verifies signature + cert identity + tlog inclusion with NO live
+  # Rekor call — immune to networks that block/TLS-inspect sigstore, and the ONLY
+  # path that verifies our short-lived keyless cert once it has expired (its embedded
+  # timestamp proves the cert was valid at signing). Releases cut before the bundle
+  # existed simply 404 here and fall through to the online .sig/.cert path below;
+  # so does any bundle that doesn't verify — the online path does the SAME full
+  # keyless check, just needing live Rekor, so this is a fallback, never a downgrade.
+  if curl -fsSL --tlsv1.2 --connect-timeout 30 --max-time 300 "$REPO_REL/manifest.sha256.bundle" -o "$bundle" 2>/dev/null; then
+    if "$COSIGN_BIN" verify-blob \
```
saadqbal, line 467: The bundle-present-but-verify-fails → fall back to sig/cert branch is never exercised. Both new bundle tests force cosign `exit 0`, and the only `COSIGN_RESULT=1` tests run with no bundle. A regression here (e.g. not falling through, or falling through wrong) would stay green. Worth: publish a bundle, reject the `--bundle` args but accept sig/cert, assert the sig/cert path runs — plus a both-fail → fail-closed case. Same gap in the Pester suite (it only checks source text). Non-blocking.

saadqbal (approved):
Nice, careful PR — offline bundle is the right fix and the fallback is genuinely a fallback, not a downgrade. Approving.

One gap worth a follow-up (non-blocking): the bundle-verifies-then-fails path isn't tested — see inline. Minor nit: on Windows, folding the try/catch into `Invoke-CosignVerifyBlob` means a cosign that can't launch now surfaces the generic "couldn't confirm authentic" message instead of the old "couldn't run the verification step" — same fail-closed, just a slightly misleading message for a broken binary.

saadqbal (dismissed):

scripts/install.sh
shujaatTracebloc: Good call — added in 5669847. **bats:** a fallback case (bundle published; cosign rejects the `--bundle` verify but accepts sig/cert → asserts the sig/cert path runs and the install proceeds) and a both-fail case (bundle present, every cosign verify fails → fail closed, k8s never runs). **Pester:** behavioural equivalents (not source-text) driving `Confirm-ManifestSignature` via a mocked `Invoke-CosignVerifyBlob` — bundle-fails falls through to sig/cert (2 verify calls, no throw); both-fail throws the authenticity error. So the fallback and fail-closed branches are now exercised on both platforms.

saadqbal (approved):

PR #602, opened 2026-08-05
refactor(installer): scrub remaining internal identifiers from bootstrap messages (#581)

shujaatTracebloc:
## What & why

Closes #581. Follow-up from the #576/#579 review (saadqbal). #579 sanitized the cosign verification-failure message; this sweeps the **remaining, lower-signal internal identifiers** still present in user-facing bootstrap messages — the `manifest.sha256` release-asset filename and the `RFC-0001 R8` internal spec code — to plain language, for consistency with #576.

**No behaviour change:** fail-closed paths, verification logic, and exit codes are untouched — only the wording of `echo`/`throw`/`Warn` strings changes.

## Changes
- **`install.sh`** — `manifest.sha256` → "the installer's integrity checksums" / "the installer's signed checksum list"; `.sig/.cert not published` / "authenticate the manifest" → "the installer's signature isn't published" / "confirm the download is authentic"; "signed manifests" → "signed releases".
- **`install.ps1`** — same wording, plus `(RFC-0001 R8)` removed from **every throw** (`Resolve-InstallRef`, `Confirm-ManifestSignature`, `Confirm-ScriptIntegrity`, the temp-dir guard).
- **Deliberately left untouched:** `RFC-0001` in code **comments** (not user-facing) and the actual `manifest.sha256` **file paths / URLs** (the real release assets).
- **Tests:** updated the two pinned assertions (`install-bootstrap.bats` + `install.Tests.ps1`) from `"no entry in manifest"` to `"isn't in the installer's signed checksum list"`.

## Verification
- No internal identifiers left in any user-facing message (grep of `echo`/`printf`/`throw`/`Warn`/`Err` — clean).
- Parse clean (bash + PS); `install.Tests.ps1` 28/0; `install-bootstrap.bats` green (the one failure is the pre-existing macOS "early bailout" flake); `bats-hygiene` 18/0; fail-closed + no-entry security tests still pass; check-style clean.
- `install.sh`/`install.ps1` are the bootstrap trust root (not manifested) — no manifest change.

## Note
Completes the #576/#577/#578 installer-robustness log-hygiene tail. Touches the same `install.sh`/`install.ps1` verify messages as #584 (#599); whichever merges second will need a small message-wording merge — I'll resolve it.

---

> [!NOTE]
> **Low Risk**
> Copy-only changes to installer error/warning strings and matching test assertions; no changes to verification logic or security gates.
> 
> **Overview**
> Continues the #576 installer log-hygiene work by rewriting **user-facing** bootstrap errors and warnings in `install.sh` and `install.ps1` so they no longer mention internal artifacts like `manifest.sha256` or the `RFC-0001 R8` spec code.
> 
> Messages now talk about **the installer's integrity checksums**, **signed checksum list**, and **installer signature** instead of manifest filenames and “authenticate the manifest.” PowerShell **throws** also drop the `(RFC-0001 R8)` suffix on ref, path-traversal, cosign, and integrity failures. **Comments and real release URLs/paths** for `manifest.sha256` are unchanged.
> 
> **Tests** (`install-bootstrap.bats`, `install.Tests.ps1`) only update expected substrings for the new copy. Verification logic, fail-closed behavior, and exit codes are untouched.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80333ba01e85e52b75cd9cba7de678a9129f4750. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -116,7 +116,7 @@ function Resolve-InstallRef {
       Warn "production box. A moved ref here can run arbitrary privileged code."
       Warn "============================================================================"
     } else {
-      throw "'$ref' is not an immutable release tag (expected vX.Y.Z). The bootstrap only trusts content-addressable release tags so a moved branch ref can't change what runs as Administrator on your box (RFC-0001 R8). Use a release tag, or for local dev set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
+      throw "'$ref' is not an immutable release tag (expected vX.Y.Z). The bootstrap only trusts content-addressable release tags so a moved branch ref can't change what runs as Administrator on your box. Use a release tag, or for local dev set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
     }
   }
 
@@ -125,7 +125,7 @@ function Resolve-InstallRef {
   # A '/' or '..' here is a path-traversal lever (it could escape the pinned tag
   # onto a mutable branch) — independent of which branch above let it through.
   if ($ref -match '/' -or $ref -match '\.\.') {
-    throw "Ref '$ref' contains a path separator or '..' -- refusing to build a fetch URL from it (path-traversal guard, RFC-0001 R8)."
+    throw "Ref '$ref' contains a path separator or '..' -- refusing to build a fetch URL from it (path-traversal guard)."
   }
 
   return $ref
@@ -343,7 +343,7 @@ function Confirm-ManifestSignature {
       Warn "Proceeding on checksum-only integrity. Not for production."
       return
     }
-    throw "cosign is required to verify the installer's signed manifest and couldn't be found or bootstrapped. Refusing to fall back to an unauthenticated, same-channel checksum (RFC-0001 R8). Fix: install cosign (https://docs.sigstore.dev/cosign/installation/) and re-run, or for local development only set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
+    throw "cosign is required to verify the installer's signed manifest and couldn't be found or bootstrapped. Refusing to fall back to an unauthenticated, same-channel checksum. Fix: install cosign (https://docs.sigstore.dev/cosign/installation/) and re-run, or for local development only set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
   }
 
   $sig  = Join-Path $TmpDir "manifest.sha256.sig"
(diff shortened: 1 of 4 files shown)


scripts/install.ps1
```diff
@@ -354,7 +354,7 @@ function Confirm-ManifestSignature {
       Warn "manifest signature/cert not published for this ref -- not verified (TRACEBLOC_ALLOW_UNVERIFIED=1)."
       return
     }
-    throw "manifest.sha256.sig / .cert not published for this release -- can't authenticate the manifest. Pin a release tag that ships them (RFC-0001 R8)."
+    throw "The installer's signature isn't published for this release -- can't confirm the download is authentic. Pin a release tag that ships it."
```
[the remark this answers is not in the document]
shujaatTracebloc: Fixed. The unverified-opt-in warn on the sig/cert path now reads "The installer's signature isn't published…", matching the bash warn and the PS throw on that path — my original grep keyed on `manifest.sha256` so this `manifest signature/cert` variant slipped through. (The generic word "manifest" in the cosign-unavailable messages is descriptive, not the filename/spec-code this PR scopes, so it stays.)

scripts/install.ps1
```diff
@@ -343,18 +343,18 @@ function Confirm-ManifestSignature {
       Warn "Proceeding on checksum-only integrity. Not for production."
       return
     }
-    throw "cosign is required to verify the installer's signed manifest and couldn't be found or bootstrapped. Refusing to fall back to an unauthenticated, same-channel checksum (RFC-0001 R8). Fix: install cosign (https://docs.sigstore.dev/cosign/installation/) and re-run, or for local development only set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
+    throw "cosign is required to verify the installer's signed manifest and couldn't be found or bootstrapped. Refusing to fall back to an unauthenticated, same-channel checksum. Fix: install cosign (https://docs.sigstore.dev/cosign/installation/) and re-run, or for local development only set `$env:TRACEBLOC_ALLOW_UNVERIFIED = '1'`."
```
saadqbal: You dropped the RFC code here but this throw (and the Warn two lines up, L342 `manifest signature NOT verified`) still say "signed manifest" / "manifest signature", while the sibling sig-not-published messages just below now say "the installer's signature". Same divergence in `install.sh` (L441/L445 vs L456/L459). Since the whole point is #576 vocabulary consistency, worth aligning these two cosign-unavailable strings too — or leave for the #584 merge you flagged. Not a blocker.

saadqbal (dismissed):
Clean scrub, tests updated on both sides and the assertions match the emitted strings exactly. One consistency nit inline — non-blocking. 👍

scripts/install.ps1
shujaatTracebloc: Good call — aligned both cosign-unavailable strings rather than deferring to the #584 merge. The `AllowUnverified` warn and the fail-closed throw/echo now read "the installer's signature" (matching the sig-not-published messages just below), in both `install.ps1` (L342/L346) and `install.sh` (L441/L445). Updated the pinned `install-bootstrap.bats` assertion to match; install.Tests.ps1 28/0 and bats-hygiene 18/0 stay green. RFC-0001 in code comments still left as-is per the PR scope. Pushed in 80333ba.

saadqbal (approved):

PR #637, opened 2026-08-06
fix(chart): image-refresh builds explicit in-cluster kubeconfig (#634)

saadqbal:
Fixes #634.

## Problem

The `image-refresh` CronJob was failing on **every tick across the whole fleet** (dev/stg/prod). The logged error — `cannot read deployment/<release>-jobs-manager (NotFound, RBAC denial, or API error)` — was misleading: it is **none of those three**. `kubectl` in the `alpine/k8s` Job pod does **not** reliably auto-detect in-cluster config; with no kubeconfig on disk it falls back to `http://localhost:8080` and every API call fails, even though the SA token and `KUBERNETES_SERVICE_*` env are present. The read gate ran both `kubectl` calls with `>/dev/null 2>&1`, so the real cause was hidden until reproduced by hand (see #634 for the full live investigation).

Consequence: image-refresh stopped reconciling new `tracebloc/{jobs-manager,pods-monitor}` digests, so **clients silently stopped picking up new control-plane images**.

## Fix

- **Build an explicit in-cluster kubeconfig** from the mounted ServiceAccount at script start (API server from `KUBERNETES_SERVICE_*`, ca + token from the projected SA volume) and point `KUBECONFIG` at it, so every `kubectl` below targets the real API server regardless of the auto-detect quirk. All existing `kubectl` calls stay literal.
- **Surface the real API error** on the fail path instead of `>/dev/null 2>&1`, so the next incident is one log line naming the actual cause (NotFound vs 403 vs connection-refused vs timeout), not a repro session.
- **Regression guards** lock the kubeconfig preamble in place.
- **Chart bump 1.9.22 → 1.9.23** (script is chart content; `chart-version-guard`).

## Validation

- `helm unittest client` — **346/346 pass**, including the 3 new guards (they assert the preamble renders into the ConfigMap script).
- **Live dev pod** running the exact preamble under the real pod constraints (`readOnlyRootFilesystem`, `HOME=/home/kubectl`, emptyDir `/tmp` + `/home/kubectl`): bare `kubectl rollout status` and `kubectl get deployment -o json | jq` both return rc=0. Before the fix the same calls hit `localhost:8080` and failed.

## Not in scope / follow-up

The environmental trigger (why it broke ~2 days ago despite unchanged chart/image/script) is still **unconfirmed** — prime suspect is the `alpine/k8s:1.30.5` tag being re-pushed with a different kubectl build combined with a node recycle (`IfNotPresent` → new nodes pull fresh). This fix makes the script robust regardless of the trigger; the root-trigger investigation can continue separately under #634.

---

> [!NOTE]
> **Medium Risk**
> Changes how the image-refresh job talks to the Kubernetes API fleet-wide; misconfiguration could block control-plane image rollouts, but behavior is locked by unittest guards and matches standard in-cluster client patterns.
> 
> **Overview**
> Fixes fleet-wide **image-refresh** CronJob failures where `kubectl` in the `alpine/k8s` job pod did not auto-detect in-cluster config and hit `http://localhost:8080`, blocking digest reconciliation for `tracebloc/{jobs-manager,pods-monitor}`.
> 
> The refresh script now **builds an explicit kubeconfig** at startup from the mounted ServiceAccount (`KUBECONFIG` on `/tmp`, API server from `KUBERNETES_SERVICE_*`, CA + **`tokenFile`** for the bearer token), including **IPv6 bracketing** for the API host. On deployment read failure it **logs the real `kubectl` stderr** instead of a generic message.
> 
> **Helm unittest** adds regression guards for that preamble; chart version **1.9.22 → 1.9.23**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ed9f9730e04484d7fac454892e8c074658a901c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.22
-appVersion: "1.9.22"
+version: 1.9.23
+appVersion: "1.9.23"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/templates/image-refresh-cronjob.yaml
+++ b/client/templates/image-refresh-cronjob.yaml
@@ -50,6 +50,27 @@ data:
     log "release=$RELEASE_NAME namespace=$RELEASE_NAMESPACE deployment=$DEPLOYMENT_NAME"
     log "  client images: tracebloc/{jobs-manager,pods-monitor} on docker.io under tag=$IMAGE_TAG"
 
+    # kubectl in the alpine/k8s image does NOT reliably auto-detect in-cluster
+    # config in this Job's pod: with no kubeconfig on disk it falls back to
+    # http://localhost:8080 and every API call fails ("connection refused"),
+    # even though the SA token and KUBERNETES_SERVICE_* env are both present
+    # (#634 — observed failing fleet-wide, where it masqueraded as the generic
+    # "cannot read deployment" below and cost a full live investigation). Build
+    # an explicit in-cluster kubeconfig from the mounted ServiceAccount so every
+    # `kubectl` below targets the real API server regardless of that quirk.
+    # These `kubectl config` calls only write the local kubeconfig file
+    # ($KUBECONFIG, on the writable /tmp emptyDir) — they make no API call.
+    SA_DIR=/var/run/secrets/kubernetes.io/serviceaccount
+    export KUBECONFIG=/tmp/kubeconfig
+    kubectl config set-cluster in-cluster \
+      --server="https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}" \
+      --certificate-authority="${SA_DIR}/ca.crt" >/dev/null
+    kubectl config set-credentials in-cluster \
+      --token="$(cat "${SA_DIR}/token")" >/dev/null
+    kubectl config set-context in-cluster \
+      --cluster=in-cluster --user=in-cluster --namespace="$RELEASE_NAMESPACE" >/dev/null
+    kubectl config use-context in-cluster >/dev/null
+
     # Get an anonymous pull-scope token for one repository on a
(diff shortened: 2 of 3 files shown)


aptracebloc (approved):
@saadqbal — **Approved.** Textbook incident fix for #634, and I verified the mechanism directly rather than trusting the render.

- **✅ The kubeconfig preamble actually produces a valid in-cluster config** — ran it under a real `kubectl`: IPv4 → `server: https://10.0.0.1:443`; IPv6 → correctly bracketed `https://[fd00::1]:443` (the `case *:*` branch). The rendered kubeconfig carries **`tokenFile:` (a path) and no `token:` value** — the bearer token never lands on argv or in the file, and kubectl re-reads the rotating projected token per call. `set users.in-cluster.tokenFile` is the right form (`set-credentials` has no `--token-file`).
- **✅ Fail-path stderr capture is correct** — `err="$(kubectl get … 2>&1 >/dev/null || true)"` routes stderr to the capture and stdout to `/dev/null`, so the failing tick names the real cause (NotFound vs 403 vs connection-refused) in one log line; `|| true` avoids a `set -e` abort and `--request-timeout=15s` bounds it.
- **✅ All 7 regression guards are sound** — the 5 `matchRegex` (KUBECONFIG, set-cluster, bracketed `api_host`, server URL, tokenFile) match; the `notMatchRegex` for `set-credentials … --token` is correctly absent (the only `--token=` in the file is the explanatory comment, not an invocation). Both Bugbot findings (token-off-argv, IPv6 bracketing) properly addressed.
- **✅ Can't regress a working cluster** — the explicit kubeconfig is built from the same SA mount + `KUBERNETES_SERVICE_*` auto-detect would have used, so it's strictly more reliable, never less.

⚪ Non-blocking: the PR body and commit message say "Chart bump 1.9.15 → 1.9.16", but the actual diff (and the Cursor summary) is **1.9.22 → 1.9.23** — the correct bump vs current `develop`. Just stale prose; the committed version is right. Worth a note so nobody's thrown by it.

PR #641, opened 2026-08-07
fix(#1143): add readiness/liveness/startup probes to the requests-proxy Deployment

divyasinghds:
Closes tracebloc/backend#1143.

## Problem
The requests-proxy Deployment has no readiness/liveness/startup probes, so the pod counts as Ready the moment the container starts — before gunicorn is actually accepting connections. The doctor's `ReadyReplicas` check goes green while the proxy can't yet serve (false-green).

## Fix
Add `startupProbe` / `livenessProbe` / `readinessProbe` as **`tcpSocket` on the gunicorn port (8888)**, mirroring the three-probe structure of `mysql-deployment.yaml`.

**Why tcpSocket, not httpGet:** the requests-proxy application lives in the `client-runtime` repo and exposes no chart-visible HTTP health path — an `httpGet` to an unknown path would 404 and permanently wedge the pod. A successful TCP accept on 8888 is the honest "gunicorn is serving" signal and is exactly what the doctor's ReadyReplicas check needs.

## Files
- `client/templates/requests-proxy-deployment.yaml` — three tcpSocket probes on port 8888
- `client/Chart.yaml` — version + appVersion → 1.9.26 (chart-content change)

## Validation
`helm lint` clean; `helm template` renders the probes as intended.

---

> [!NOTE]
> **Low Risk**
> Helm-only change to probe timing on a single-replica sidecar; no auth or data-path logic, with documented limits of tcpSocket vs full app health.
> 
> **Overview**
> Fixes **false-green** readiness for the Service Bus egress proxy: `tb doctor` was treating the pod as ready as soon as the container started, before gunicorn listened on **8888**.
> 
> The requests-proxy **Deployment** now has **startup**, **liveness**, and **readiness** probes as **`tcpSocket` on port 8888**, aligned with the three-probe pattern used elsewhere in the chart (e.g. MySQL). **HTTP probes were avoided** because the proxy app has no chart-visible health URL; a bad `httpGet` could 404 and block readiness permanently.
> 
> **Chart version** is bumped to **1.9.26**. Inline comments note this is a partial fix: TCP success only means the gunicorn master bound the socket, not that `create_app()` and the worker can actually relay traffic—a future `/healthz` in client-runtime would be needed for that.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9634e742e180eb579dce3b7d1a8ae753c66036ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/templates/requests-proxy-deployment.yaml
+++ b/client/templates/requests-proxy-deployment.yaml
@@ -52,8 +52,15 @@ spec:
           # tcpSocket (not httpGet): the proxy app lives in client-runtime and
           # exposes no chart-visible HTTP health path, and a probe against an
           # unknown path would 404 and wedge the pod. A successful TCP accept on
-          # the gunicorn listen port is the honest "the proxy is serving" signal.
-          # Mirrors mysql-deployment.yaml's startup/liveness/readiness structure.
+          # :8888 confirms the gunicorn MASTER has bound the listen socket. That
+          # is a strict improvement over no probes — it catches a container that
+          # never starts — but it does NOT fully close the false-green: the
+          # master binds the socket before the single worker finishes importing
+          # create_app(), and keeps it bound while a worker crash-loops on that
+          # import, so a proxy that can't actually relay can still read Ready.
+          # Fully closing it would need a real /healthz served from
+          # client-runtime. Mirrors mysql-deployment.yaml's
+          # startup/liveness/readiness structure.
           startupProbe:
             tcpSocket:
               port: 8888
--- a/client/templates/requests-proxy-deployment.yaml
+++ b/client/templates/requests-proxy-deployment.yaml
@@ -45,6 +45,36 @@ spec:
             - "requests_proxy_server:create_app()"
           ports:
             - containerPort: 8888
+          # #1143: without probes the pod is "Ready" the instant the container
+          # process starts — before gunicorn has bound :8888 — so tb doctor's
+          # "Service Bus egress (requests-proxy) ready" (which reads
+          # ReadyReplicas) shows a false green even when the proxy can't relay.
+          # tcpSocket (not httpGet): the proxy app lives in client-runtime and
+          # exposes no chart-visible HTTP health path, and a probe against an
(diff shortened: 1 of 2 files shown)


client/templates/requests-proxy-deployment.yaml
```diff
@@ -45,6 +45,36 @@ spec:
...
+            timeoutSeconds: 3
+            failureThreshold: 12
+          livenessProbe:
+            tcpSocket:
+              port: 8888
+            initialDelaySeconds: 0
+            periodSeconds: 10
+            timeoutSeconds: 3
+            failureThreshold: 5
+          readinessProbe:
```
saadqbal, line 78: Worth knowing (non-blocking): tcpSocket:8888 goes green the moment the gunicorn **master** binds the socket, which happens before the single `--workers=1` worker finishes importing `create_app()`. So a worker that crash-loops on boot still reads Ready — the master keeps :8888 bound and just respawns it. This narrows the false-green (container-started → socket-bound) but doesn't fully close it; only a real `/healthz` in client-runtime would. Strict improvement over no probes, so fine to ship — just flagging the gap since the comment above calls tcpSocket the honest "proxy is serving" signal, and it's really "master is listening."

saadqbal (dismissed):
Nice, tight fix — probe timings mirror mysql-deployment, tcpSocket is the right call given there's no chart-visible HTTP path, and version/appVersion move in lockstep as usual. One non-blocking note inline about the residual false-green. Approving.

aptracebloc (dismissed):
Looks good and I adhere to Asad's non-blocking comment.

client/templates/requests-proxy-deployment.yaml
divyasinghds: Good catch — reworded the comment to say tcpSocket:8888 confirms the gunicorn master is listening (not that the proxy is fully serving); it narrows but does not close the false-green, which needs a real /healthz in client-runtime. Probes unchanged.

saadqbal (approved):

PR #642, opened 2026-08-07
fix(#1144): pin jobs-manager memory requests==limits so a mass restart can't OOM-kill it

divyasinghds:
Closes tracebloc/backend#1144.

## Problem
On a cluster-recovery mass restart, jobs-manager gets OOM-killed (exit 137). Despite the ticket's "CPUKilled" label, exit 137 is a **memory** SIGKILL — CPU limits only throttle, they never kill. The pod was Burstable (memory requests < limits), so once its working set exceeded the request it was a prime node-OOM victim.

## Fix
Pin `requests.memory == limits.memory` for both jobs-manager containers (api 1Gi, pods-monitor 512Mi). This does **not** make the pod Guaranteed QoS — it stays **Burstable** because `requests.cpu != limits.cpu`. What it does is lower each container's `oom_score_adj`, so when the node runs out of memory the kernel OOM-killer treats jobs-manager as a **less-preferred victim**. Same mechanism the chart's mysql already relies on (mysql likewise pins memory and omits its cpu limit). Also raised CPU **requests** (api 100m→250m, pods-monitor 50m→100m) to avoid startup throttling. Memory *limits* are unchanged (1Gi / 512Mi) — this only raises the floor, not the ceiling.

## Files
- `client/templates/jobs-manager-deployment.yaml`, `client/values.yaml`
- `client/tests/jobs_manager_test.yaml` — resource assertions updated to the pinned values
- `client/Chart.yaml` — version + appVersion → 1.9.27

## Validation
`helm lint` clean; `helm unittest` passes (jobs_manager suite + full suite, 354 tests).

---

> [!NOTE]
> **Medium Risk**
> Changes default resource reservations for a core control-plane deployment, increasing per-node memory scheduling footprint (~768Mi more requested across both containers) but only adjusts Kubernetes QoS/OOM preference, not application logic.
> 
> **Overview**
> Addresses **cluster-recovery mass restarts** where the jobs-manager **api** and **pods-monitor** containers were **OOM-killed (exit 137)** because memory **requests** sat below **limits**, making them favored node-OOM victims.
> 
> The chart now **raises memory requests to match existing limits** (api **1Gi**, pods-monitor **512Mi**) and **increases CPU requests** (api **100m→250m**, pods-monitor **50m→100m**) so startup is less throttled under contention. **Memory limits are unchanged**; pods remain **Burstable** (CPU request ≠ limit), matching the mysql pinning pattern. Defaults and template comments live in `values.yaml` and `jobs-manager-deployment.yaml`; **helm unittest** expectations were updated. Chart version bumps to **1.9.27**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d26d2a9eebc10b32cbaeb3cd7253b080a2ddb7ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/templates/jobs-manager-deployment.yaml
+++ b/client/templates/jobs-manager-deployment.yaml
@@ -104,9 +104,11 @@ spec:
         # for node memory + CPU; with requests (512Mi) below the limit (1Gi)
         # jobs-manager ran Burstable and was the kernel OOM-killer's preferred
         # victim once its working set exceeded the request — the observed
-        # exit-137 kill of both containers. Pinning requests==limits gives it a
-        # Guaranteed memory reservation so it is no longer the node's largest
-        # victim, exactly as mysql-deployment.yaml already does. requests.cpu is
+        # exit-137 kill of both containers. Pinning requests.memory==limits.memory
+        # lowers the container's oom_score_adj so it is a less-preferred victim
+        # when the node runs out of memory — the pod stays Burstable (cpu request
+        # != limit), exactly as mysql-deployment.yaml already does (mysql has no
+        # cpu limit either). requests.cpu is
         # also raised (100m -> 250m) so a mass restart doesn't throttle its
         # startup to a crawl (the ~9m recovery seen on lukas-test).
         resources:
@@ -307,9 +309,10 @@ spec:
             drop: ["ALL"]
         # #1144: same cluster-recovery treatment as the api container above —
         # pods-monitor was also exit-137 killed in the mass restart. Pin
-        # requests.memory == limits.memory (512Mi) for a Guaranteed memory
-        # reservation, and raise requests.cpu (50m -> 100m) so its startup is
-        # not starved under contention.
+        # requests.memory == limits.memory (512Mi) to lower its oom_score_adj
+        # (a less-preferred node OOM victim; the pod stays Burstable), and raise
+        # requests.cpu (50m -> 100m) so its startup is not starved under
+        # contention.
         resources:
           requests:
             cpu: {{ .Values.resources.podsMonitor.requests.cpu | default "100m" | quote }}
--- a/client/templates/jobs-manager-deployment.yaml
+++ b/client/templates/jobs-manager-deployment.yaml
@@ -98,10 +98,21 @@ spec:
         # pod becomes the OOM-killer's preferred target on the node, and
         # mysqld dies as collateral. 512Mi/1Gi covers steady state with
         # headroom; revisit if heap profiling shows the working set growing.
+        #
+        # #1144: requests.memory now == limits.memory (1Gi). On a single-node
+        # cluster-recovery mass restart every pod restarts at once and contends
+        # for node memory + CPU; with requests (512Mi) below the limit (1Gi)
+        # jobs-manager ran Burstable and was the kernel OOM-killer's preferred
+        # victim once its working set exceeded the request — the observed
+        # exit-137 kill of both containers. Pinning requests==limits gives it a
+        # Guaranteed memory reservation so it is no longer the node's largest
+        # victim, exactly as mysql-deployment.yaml already does. requests.cpu is
+        # also raised (100m -> 250m) so a mass restart doesn't throttle its
+        # startup to a crawl (the ~9m recovery seen on lukas-test).
         resources:
           requests:
-            cpu: {{ .Values.resources.jobsManager.requests.cpu | default "100m" | quote }}
-            memory: {{ .Values.resources.jobsManager.requests.memory | default "512Mi" | quote }}
+            cpu: {{ .Values.resources.jobsManager.requests.cpu | default "250m" | quote }}
+            memory: {{ .Values.resources.jobsManager.requests.memory | default "1Gi" | quote }}
           limits:
             cpu: {{ .Values.resources.jobsManager.limits.cpu | default "1000m" | quote }}
             memory: {{ .Values.resources.jobsManager.limits.memory | default "1Gi" | quote }}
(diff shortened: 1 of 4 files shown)


client/templates/jobs-manager-deployment.yaml
```diff
@@ -98,10 +98,21 @@ spec:
...
+        # exit-137 kill of both containers. Pinning requests==limits gives it a
+        # Guaranteed memory reservation so it is no longer the node's largest
+        # victim, exactly as mysql-deployment.yaml already does. requests.cpu is
+        # also raised (100m -> 250m) so a mass restart doesn't throttle its
+        # startup to a crawl (the ~9m recovery seen on lukas-test).
         resources:
           requests:
-            cpu: {{ .Values.resources.jobsManager.requests.cpu | default "100m" | quote }}
-            memory: {{ .Values.resources.jobsManager.requests.memory | default "512Mi" | quote }}
+            cpu: {{ .Values.resources.jobsManager.requests.cpu | default "250m" | quote }}
```
saadqbal, line 116: This won't actually be **Guaranteed** QoS. Guaranteed needs `requests == limits` for *cpu too*, on every container in the pod. Here cpu request (250m) ≠ limit (1000m) — and same for pods-monitor (100m vs 500m) — so the pod stays **Burstable**. (mysql is Burstable too, for the same reason: it deliberately has *no* cpu limit.)

The memory-request bump is still a real improvement — Burstable `oom_score_adj` scales down as `memRequest/nodeCapacity` rises, so jobs-manager becomes a less-preferred OOM victim. But it's not near-immune under node pressure the way "Guaranteed" implies; the kernel can still pick it if the node is tight enough.

Two clean options: pin cpu `requests==limits` as well (and the resourceless init container would need a `resources` block) to genuinely reach Guaranteed — or, if you want to keep the burst headroom, just reword the comments/PR to "lowers oom_score_adj" instead of "Guaranteed QoS". Either way the fix lands; it's the framing that's off.

aptracebloc (commented):
@divyasinghds — effective OOM fix, and the resource choices are right. One substantive correction before this is merge-ready: the **"Guaranteed QoS" claim is inaccurate** — the pod stays Burstable.

## 🟡 The pod is Burstable, not Guaranteed
The PR body, commit message, `values.yaml` comment, and both template comments all say this yields **Guaranteed QoS**. Kubernetes grants Guaranteed only when **every** container has `request == limit` for **both** cpu *and* memory. Here:

| container | cpu req / limit | mem req / limit |
|---|---|---|
| api | **250m / 1000m** ✗ | 1Gi / 1Gi ✓ |
| pods-monitor | **100m / 500m** ✗ | 512Mi / 512Mi ✓ |

cpu request ≠ cpu limit on both, so `kubectl describe pod` will report **`QoS Class: Burstable`** — contradicting the comments.

**The fix still works, just not via that mechanism.** Raising `requests.memory` to the limit lowers the container's Burstable `oom_score_adj` (`1000 − 1000·memRequest/nodeAllocatable`), and a container whose request equals its limit is never "over request," so the node OOM-killer stops preferring it. That's genuine protection — and it's exactly what mysql does: mysql has *no cpu limit* at all, so it isn't Guaranteed either. So the "same as mysql" comparison is accurate in mechanism; only the "Guaranteed QoS" label is wrong.

Caveat worth knowing: a truly Guaranteed pod gets `oom_score_adj = −997`, far below the best Burstable score (floor 2). Since the whole tracebloc pod set on a k3d node is Burstable, the *relative* re-ordering still saves jobs-manager in practice — but it isn't the maximal protection "Guaranteed" implies.

**Suggested fix (wording only, no resource change):** reword the four spots to e.g. *"memory request pinned to limit → lowest Burstable OOM score, no longer the preferred victim (same requests==limits-on-memory approach mysql uses)."* I'd **not** chase real Guaranteed by pinning cpu request==limit — that would either re-introduce the startup throttle you just removed (cap at 250m/100m) or reserve 1000m/500m (scheduling pressure on small nodes), and mysql doesn't bother. The resource shape as-is is the right call.

## ✅ Verified correct
- **`--reuse-values` safe:** template `| default` fallbacks updated to match `values.yaml` (api `250m`/`1Gi`, pods-monitor `100m`/`512Mi`), so an upgrade from a release predating these keys still gets the new floors.
- **Only the floor moved** — memory *limits* unchanged (1Gi/512Mi), so headroom can't shrink.
- **Security untouched** — jobs-manager is control-plane (not a training pod); securityContext unchanged.

## ⚪ Non-blocking notes
- **Scheduling footprint** grows to **1.5Gi mem + 350m cpu** reserved (was 768Mi + 150m). On a tight single-node k3d that could push toward `Pending` — worth a sanity check on min node sizing (documented tradeoff).
- **Chart 1.9.25 → 1.9.27** skips 1.9.26 (taken by #641). Fine — the guard checks monotonic increase, not contiguity. Heads-up that #637 (→1.9.23) is now behind both and will need a re-bump to ≥1.9.28.

Net: the fix is sound and I'd approve once the "Guaranteed QoS" phrasing is corrected — it's the one thing that will actively mislead the next reader.

client/templates/jobs-manager-deployment.yaml
divyasinghds: Reworded — dropped the "Guaranteed QoS" claim everywhere (templates, values, test desc, PR body); it stays Burstable and the protection is the lowered oom_score_adj from requests.memory==limits.memory (mysql pattern). Resource values unchanged.

saadqbal (dismissed):

saadqbal (approved):

PR #646, opened 2026-08-10
fix(macos): only treat a 64-hex string as the Docker DMG checksum (#644)

LukasWodka:
## Summary
Bugbot #644 (High) — recurrence of #629. The Docker DMG checksum fetch treated **any** non-empty `awk` hit for a line mentioning `Docker.dmg` as a real SHA-256. A TLS-inspecting proxy returning an HTML error body that mentions that filename yields non-hash text → non-empty → the code entered the **fail-closed verify path**, failed the compare, and **hard-aborted an otherwise-fine install** as "corrupted or tampered", instead of taking the intended warn/unreachable path.

## Fix
- `awk` now requires field 1 to be a 64-hex SHA-256 before capturing it
- a belt-and-suspenders guard blanks any non-64-hex value → garbage takes the warn path, never a fail-closed mismatch abort on garbage
- this is the same 64-hex structure check the PowerShell tool downloads already apply

**Real corruption/tampering still fails closed** — a valid 64-hex hash that does not match the DMG aborts exactly as before (#556).

## Supply chain
`setup-macos.sh` is manifest-covered (RFC-0001 R8), so `scripts/manifest.sha256` is regenerated; `gen-manifest.sh --check` passes.

## Test plan
- `shellcheck -S warning` clean on the changed region
- `gen-manifest.sh --check` — manifest current
- proxy/HTML body mentioning Docker.dmg → warn path (no abort); genuine mismatch → still aborts

---

> [!NOTE]
> **Low Risk**
> Narrow installer hardening around checksum parsing; legitimate mismatch detection when a valid hash is available is preserved.
> 
> **Overview**
> **Hardens Docker Desktop DMG checksum parsing** so TLS-inspecting proxies or HTML error pages that mention `Docker.dmg` no longer get treated as a published SHA-256.
> 
> The `checksums.txt` `awk` step now only accepts lines whose first field is a **64-character hex** hash before matching `Docker.dmg`. A follow-up guard clears any value that still isn’t valid hex, so the installer falls through to the **unverified warn path** instead of entering fail-closed verify and aborting with “corrupted or tampered.” **Genuine checksum mismatches are unchanged** — when a real 64-hex hash is fetched and doesn’t match the DMG, install still hard-fails.
> 
> `scripts/manifest.sha256` is updated for the manifest-covered `setup-macos.sh` change.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5386f092abba30c3504f9da2561a7101a3700186. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


shujaatTracebloc (approved):
LGTM — approving.

Verified the fix does what it claims: the `awk` structure guard (`$1 ~ /^[0-9a-fA-F]{64}$/`) plus the belt-and-suspenders regex check route a TLS-inspecting proxy's HTML error body (which can contain the text `Docker.dmg`) to the **warn** path instead of a false "tampered" fail-closed abort, while a genuine 64-hex SHA-256 still fails closed on mismatch (#556 preserved). `scripts/manifest.sha256` regenerated for the edited file. CI all-green, Bugbot clean. Nice, tightly-scoped fix.

PR #661, opened 2026-08-11
fix(setup): fail closed on an unfetchable Docker DMG checksum (#629)

shujaatTracebloc:
Closes #629. Systemic class 2 (integrity verification failing open).

## The bug
The macOS installer verified the Docker Desktop DMG against Docker's `checksums.txt`, and correctly **fails closed on a mismatch** (#556). But when the checksum was **unfetchable** — a TLS-inspecting proxy stripping it, a transient CDN error, or an on-path attacker blocking just that one URL while serving a tampered `.dmg` — it emitted a warning and installed the DMG **UNVERIFIED**. The DMG is then `hdiutil attach`ed and copied into `/Applications` under `sudo`.

This was the **lone fail-open** in an otherwise fail-closed installer. `kubectl` / `k3d` / `helm` **also fetch their checksum over the network** (kubectl.sha256, checksums.txt, .sha256sum) and **abort on an empty hash** (`setup-linux.sh`, via `_verify_sha256` / `[ -z "$want" ] || … error`). So the "don't brick installs behind a proxy" rationale applied equally to those three — yet they abort. Docker was the inconsistent outlier.

## Fix
- **Extracted** the DMG fetch-and-verify into `_verify_docker_dmg` — testable, and exactly the issue's structural suggestion ("route fetch-and-verify through a helper rather than open-coding the DMG path").
- **Fail closed by default** on both a mismatch and an unfetchable checksum, consistent with the pinned tools. A user knowingly behind a checksum-stripping proxy can opt out with `TRACEBLOC_ALLOW_UNVERIFIED_DOCKER_DMG=1` (this replaces the old opt-**in** `TRACEBLOC_REQUIRE_DOCKER_DMG_CHECKSUM`, which had the safe path backwards; it isn't referenced anywhere else).
- Otherwise behavior-preserving: the retrying fetch, the non-SHA proxy-garbage guard, and the mismatch abort are unchanged.

## Tests / evidence
5 new bats cases in `setup-macos.bats`:
- matching checksum verifies + keeps the DMG,
- mismatch fails closed + removes the DMG,
- **unfetchable checksum fails closed by default + removes the DMG**,
- non-SHA proxy garbage is treated as unfetchable (fails closed),
- `TRACEBLOC_ALLOW_UNVERIFIED_DOCKER_DMG=1` opts out (warn + proceed).

Full macOS bats suite green (**143 tests**, incl. bats-hygiene + drift), shellcheck clean. `scripts/`-only change, so no Chart.yaml bump.

> Reviewer: please assign whoever owns the installer scripts — I left it unassigned since there's no per-repo default.

---

> [!NOTE]
> **Medium Risk**
> Tightens installer integrity before sudo-mounted DMG install; installs behind checksum-blocking proxies will fail unless they set the new opt-out env var.
> 
> **Overview**
> macOS Docker Desktop install **no longer proceeds when `checksums.txt` cannot be fetched**; it now aborts and deletes the DMG, aligning with kubectl/k3d/helm on Linux. Mismatch handling is unchanged.
> 
> Verification is moved into **`_verify_docker_dmg`**, called from `install_docker_desktop` after download. The old warn-and-install path is replaced by fail-closed default behavior; **`TRACEBLOC_ALLOW_UNVERIFIED_DOCKER_DMG=1`** is the explicit opt-out (replacing the unused **`TRACEBLOC_REQUIRE_DOCKER_DMG_CHECKSUM`** opt-in). Five bats cases cover match, mismatch, unfetchable checksum, proxy HTML garbage, and the opt-out. **`scripts/manifest.sha256`** is updated for `setup-macos.sh`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 478766c0f4ac76c9df6d4eefd0c825a42c6364a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saadqbal (dismissed):
Clean fix — closes the lone fail-open, defaults to fail-closed like kubectl/k3d/helm, and the opt-out is the right escape hatch. Tests cover the four paths, manifest hash checks out, old var isn't referenced anywhere else. 👍

One dead-code nit inline (non-blocking). Also worth a thought: the compare open-codes `shasum -a 256 | awk` instead of the shared `_verify_sha256` helper the other three tools use — the very consistency this PR argues for. I get why (the existing bats mocks stub `shasum` to print a hash, not honour `--check`), so it'd mean reworking the mock. Fine to leave as-is; just flagging since #629 was partly about routing through one helper.

saadqbal (approved):

PR #683, opened 2026-08-12
fix(installer): say where a step died, and stop calling a stopped runtime a fresh machine (client#681, client#682)

LukasWodka:
Fixes #681. Fixes #682.

Both came out of one field report: a macOS install that stopped at **b) Installing what tracebloc needs** and printed nothing — no reason on screen, and nothing in the install log either, even though that log is the entire session `tee`'d. The machine turned out to have a healthy environment from weeks earlier; Docker Desktop simply wasn't running.

Auditing outward from that, the bash installer was missing three diagnostic affordances the PowerShell installer already had. This PR closes them.

## 1 — A step could fail with zero diagnostics (#681)

`grep -rn 'trap.*ERR\|BASH_LINENO\|BASH_COMMAND' scripts/` returned nothing. Under `set -euo pipefail`, any command failing outside an `if`/`&&`/`||` context killed the installer silently, and `install_cleanup` printed a fixed string with no file, line, command, or exit code.

- Adds an ERR trap recording `file:line`, the command, and the status. **`set -E` is the load-bearing part** — without errtrace an ERR trap fires only at top level, so every failure inside `install_macos`/`install_linux` (nearly all of them) would still be invisible.
- The closer names the site on screen; the command text goes to the log only. `BASH_COMMAND` is unexpanded (`cmd "$VAR"`, never the value), so it cannot leak a credential.
- **Ctrl-C is no longer reported as a failure.** `install-k8s.sh` routes SIGINT/SIGTERM to `exit 130`/`143`, and those fell into the same "Installation did not complete" branch as a real error — which is a large part of why the original report was unresolvable.
- Per-stage `log()` breadcrumbs in step b on both platforms.

Counterparts already shipped on Windows: `Show-FatalError` / `Show-Interrupted` (#577), `Err`'s detail lines (#423).

## 2 — A stopped runtime was classified as a fresh machine (#682)

`_cluster_exists` tries three probes and **all three send stderr to `/dev/null` and return 1**, so a down daemon returns exactly what an empty machine returns. `_assess_classify` then set `fresh`/`no-cluster` and announced a first-time setup over a working environment.

- Classifies an installed-but-unreachable runtime as `degraded`/`runtime-down`, before the cluster probe, and says so honestly.
- Deliberately narrow: **no docker binary at all is still `fresh`** (correct — that machine really is new), and **`permission denied` is not "down"** (a Linux user outside the docker group has a different fix). A wedged daemon that won't answer inside the bound counts as down.

Counterpart already shipped on Windows: the tri-state `Get-ClusterRunStateFromList` (#557).

### One design note worth flagging

My first cut *blocked* on a down runtime and told the user to start Docker. That was wrong and I backed it out: `install_docker_desktop` already launches Docker Desktop and waits, and `create_cluster` reconciles an existing cluster — so blocking would have removed a step that works today, trading one bad outcome for another. **The bug was the claim, not the flow**, so the run now continues and only the misleading "first time on this machine" line is gone.

## Test plan

- `make check` — green (lint, shellcheck, drift, helm lint).
- `make bats` — **940 tests, 0 failures** (15 new).
- New coverage: ERR capture (location / command / status), errtrace reaching nested functions, first-failure-wins, the command staying out of the on-screen output, 130/143 vs a real failure, the six `_assess_runtime_down` cases, and two mutation-real guards that a down runtime is never reported as fresh.
- The pipefail/SIGPIPE death class is covered explicitly — the exact shape that produced no output at all before.
- `scripts/manifest.sha256` regenerated (`scripts/gen-manifest.sh`), as the changed files are in the signed-installer trust root.

## What this does not do

It does not identify the root cause of the original report — that machine's step b emitted zero bytes and I could not reproduce the silent exit on comparable hardware. This makes the next occurrence self-reporting: the log will name the file, line, command, and status.

---

> [!NOTE]
> **Medium Risk**
> Touches installer-wide ERR/EXIT handling and the stop-and-check gate that decides fresh vs reconcile, so a regression could misroute re-runs or obscure real failures. Changes are narrowly scoped and covered by new bats tests.
> 
> **Overview**
> Makes silent installer deaths self-reporting, and stops telling machines with a stopped Docker daemon that they are a first-time install.
> 
> **Failure diagnostics (#681).** Arms an ERR trap with `set -E` so failures inside `install_macos` / `install_linux` record file, line, command, and exit status. `install_cleanup` names the site on screen and logs the command; Ctrl-C / SIGTERM (130/143) now report as interrupted instead of "did not complete". Adds per-stage `log()` breadcrumbs in step b on both platforms.
> 
> **Runtime classification (#682).** New `_assess_runtime_down` probe runs before the cluster check and marks an installed-but-unreachable Docker as `degraded`/`runtime-down`. The run continues (Docker start + reconcile still work); only the misleading "first time" claim is removed. Missing Docker stays `fresh`; `permission denied` is not treated as down.
> 
> **Pipefail / SIGPIPE (#680).** Switches early-exit pipes to capture-then-match in `_cluster_exists`, macOS admin/arch detection, and related probes so SIGPIPE under `pipefail` no longer reads as "no cluster" / "not admin" / wrong arch — a second route into the #682 misclassification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8b5b7f53da3b119ed941da1d74b9b697716c278e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saadqbal (approved):

saadqbal (dismissed):
Really careful PR — both routes to the #682 misclassification are genuinely closed (runtime-down probe ahead of the cluster check, and the capture-then-match SIGPIPE fix in _cluster_exists), the #681 breadcrumbs + ERR trap attribute the right site, and the new bats are mutation-real. Ran assess/cluster/common/setup-macos-lifecycle locally: 226/226 green, manifest hashes match. One tiny non-blocking nit inline.

saadqbal (approved):

PR #694, opened 2026-08-12
fix(chart): the ingestor tag fallback is per-environment, not always prod's

LukasWodka:
## Summary

`tracebloc.ingestorTag`'s last-resort literal was a bare `"0.8"` for **every**
environment. On the `--reuse-values` replay it exists for — a release predating
`channelTags`, where no chart default is adopted — a **dev or staging edge
spawned the PROD ingestor line**.

That inverts the whole point of backend#1360: dev/stg channels exist so an
ingestor change can be validated on a real edge *without* a prod release. An
edge silently validating prod's image reports on the wrong artifact.

**It is no longer merely a wrong tag.** The prod float has moved past the
ordering ceiling documented at `values.yaml` `prodDigest`: the 0.8 line no
longer carries the ingestor's `edgeuser` `DB_USER` default that
data-ingestors#468 removed (**backend#1853**). `serviceDbAccountsByEnv` supplies
`DB_USER` on dev/stg, so those two survive it — but the coupling is accidental,
and the same literal is where an out-of-vocabulary `CLIENT_ENV` lands, where
`serviceDbAccountsByEnv` misses too and nothing supplies `DB_USER`. That is
backend#1752 reconstructed from a typo.

Fallback is now keyed on the **resolved** environment (`dev`→`dev`, `stg`→`stg`,
`prod`→`0.8`), so the documented aliases reach it too.

## Verified on the real chart — input reaching output, with a control

Rendered against `client/ci/bm-values.yaml`, `helm v4.1.1`.

**Before (`origin/develop` @ `ead58c6`)** — `channelTags: null`, i.e. the replay
this branch serves:

```
channelTags:null CLIENT_ENV=dev  -> INGESTOR_IMAGE_TAG value: "0.8"   <-- prod line on a dev edge
channelTags:null CLIENT_ENV=stg  -> INGESTOR_IMAGE_TAG value: "0.8"   <-- prod line on a stg edge
channelTags:null CLIENT_ENV=prod -> INGESTOR_IMAGE_TAG value: "0.8"
```

Control proving the opposite input differs (same command, `channelTags` present):

```
channelTags present, CLIENT_ENV=dev -> INGESTOR_IMAGE_TAG value: "dev"
```

**After (this branch):**

```
dev          channelTags:null -> value:"dev"
stg          channelTags:null -> value:"stg"
prod         channelTags:null -> value:"0.8"      <-- control: prod unchanged
development  channelTags:null -> value:"dev"      <-- alias reaches the fallback
[...]

PR #695, opened 2026-08-12
fix(chart): close the CLIENT_ENV + channelTags vocabularies, and correct a false schema doc

LukasWodka:
> **⚠️ NEEDS A CALL FROM @LukasWodka BEFORE MERGE — see "Blast radius" below.**
> Finding 1 turns an open vocabulary into a closed one. That is breaking for any
> edge currently deploying an out-of-vocabulary `CLIENT_ENV`. I believe failing
> closed is right and consistent with the chart's own conventions, but it is a
> policy call, not a correctness one, and I am not making it unilaterally.
> Findings 2 and 3 are unambiguous and could ship alone if you'd rather split.

## Summary

Three findings from a mutable-tags sweep, all the same shape: the chart
**documented a vocabulary it never enforced**, and in one case **documented a
fallback it does not have**. Follow-up to backend#1723, which fixed the
*resolution* and left the *validation* open.

Every claim below was verified by rendering the real chart, each with a control
proving the opposite input behaves differently. Helm `v4.1.1`, values
`client/ci/bm-values.yaml`, base `origin/develop` @ `ead58c6`.

---

### 1. `env.CLIENT_ENV` had no `enum` — the important one

`tracebloc.clientEnv` normalized three aliases and passed **anything else
through raw**. Measured on `develop`:

```
CLIENT_ENV=prod  -> jobs-manager:prod  pods-monitor:prod  resource-monitor:prod   INGESTOR_IMAGE_DIGEST="sha256:05e1249…"
CLIENT_ENV=prd   -> jobs-manager:prd   pods-monitor:prd   resource-monitor:prd    INGESTOR_IMAGE_DIGEST=""      <-- pin GONE
CLIENT_ENV=qa    -> jobs-manager:qa    pods-monitor:qa    resource-monitor:qa     INGESTOR_IMAGE_DIGEST=""
CLIENT_ENV=Prod  -> jobs-manager:Prod  pods-monitor:Prod  resource-monitor:Prod   INGESTOR_IMAGE_DIGEST=""
```

Published tags, checked against Docker Hub 2026-08-12 — no `prd`, no `qa`, no `Prod`:

```
tracebloc/jobs-manager      ['arm-canary', 'dev', 'prod', 'staging', 'stg']
tracebloc/pods-monitor      ['arm-canary', 'dev', 'latest', 'prod', 'staging', 'stg']
tracebloc/resource-monitor  ['arm-canary', 'dev', 'prod', 'staging', 'stg']
```

**The clearest demonstration is a one-letter typo.** `development` is an alias;
`develop` is not:

```
CLIENT_ENV=development -> tag: dev       INGESTOR_IMAGE_TAG="dev"   SERVICE_DB_ACCOUNTS="1"
CLIENT_ENV=develop     -> tag: develop   INGESTOR_IMAGE_TAG="0.8"   SERVICE_DB_ACCOUNTS=(absent)
```

Four things break at once, none of them loudly: a control-plane tag that does
not exist, `images.ingestor.channelTags` missed, `serviceDbAccountsByEnv`
missed so **no service DB account is provisioned**, and the prod digest pin
dropped (it applies only where the env resolves to exactly `prod`). That last
pair is backend#1752 reconstructed from a typo — and it matters more now that
the `0.8` float has moved past the `edgeuser` ceiling (**backend#1853**).

The only validator that existed was `client-runtime` `jobs_manager.py`'s
`sys.exit(1)` on "Unknown CLIENT_ENV" — which lives **inside the container that
cannot start**, so it cannot help.

**Fix:** an `enum` in `values.schema.json` (primary gate) plus a `fail` in
`tracebloc.clientEnv` (backstop — the enum is only checked where the packaged
schema is read, and `--skip-schema-validation` skips it). Failing closed matches
the chart's own conventions: it already `fail`s on placeholder `clientId`, empty
training CIDRs, non-alphanumeric service passwords, `perDatasetPvcs` without
`clusterScope`, and a missing metrics API.

**Accepted vocabulary — the complete list:**

| value | resolves to |
|---|---|
| `dev` | `dev` |
| `stg` | `stg` |
| `prod` | `prod` |
| `development` | `dev` |
| `staging` | `stg` |
| `production` | `prod` |
| `""` / unset | `prod` |

Everything else is refused.

### 2. `channelTags` accepted arbitrary keys

Proven with the control, on `develop`:

```
CLIENT_ENV=staging  channelTags.staging=0.7  ->  INGESTOR_IMAGE_TAG "stg"   <-- the key is IGNORED
CLIENT_ENV=staging  channelTags.stg=0.7      ->  INGESTOR_IMAGE_TAG "0.7"   <-- takes effect
channelTags.totallyBogusKey=9.9              ->  validates fine, silently inert
```

The lookup is on the **resolved** env, so only `dev`/`stg`/`prod` can ever
match. The same word "staging" is normalized in one place and meaningless one
key over — a customer writing the alias gets silence.
**Fix:** `additionalProperties: false`.

### 3. A schema description that was false

`values.schema.json` said the mysql-client tag *"Empty falls back to
env.CLIENT_ENV"*. The template is `.Values.images.mysqlClient.tag | default
"prod"`:

```
CLIENT_ENV=dev  -> mysql-client:prod
CLIENT_ENV=stg  -> mysql-client:prod
CLIENT_ENV=prod -> mysql-client:prod
```

Published tags (Docker Hub, 2026-08-12) — **no `dev`, no `stg`**:

```
['8.0', '8.4', '8.4-528d7cee…', '8.4-5af5e835…', '8.4-87a7b276…', '8.4-b9c6b7a8…', 'latest', 'prod']
```

So the doc was not merely imprecise, it was an instruction to break the edge.
And it costs more than a failed pull — an env-derived tag is unrecognized by
`tracebloc.mysqlEngineMajor`, so the **mysql-format-guard silently stands
down**:

```
mysqlClient.tag=prod -> TB_EXPECTED_ENGINE="5.7"       <-- armed
mysqlClient.tag=dev  -> TB_EXPECTED_ENGINE="unknown"   <-- disarmed
mysqlClient.tag=stg  -> TB_EXPECTED_ENGINE="unknown"
mysqlClient.tag=8.4  -> TB_EXPECTED_ENGINE="8.4"
```

Worth noting: **`values.yaml` already said this correctly** ("mysql-client is
only published under the `prod` tag — it has no dev/staging variants, so we
decouple it from `env.CLIENT_ENV`"). The two files contradicted each other and
the schema — the copy Helm shows customers — held the false half. Description
corrected; **the template is right and is left alone.**

---

## Blast radius — the part that needs your call

Closing an open vocabulary is **breaking for any edge currently deploying an
out-of-vocabulary value**. Such an edge is *already* broken — it is pulling
control-plane tags that do not exist, running with no service DB account, and
has no prod digest pin — but the failure moves:

| | before | after |
|---|---|---|
| where it fails | pod start (`ImagePullBackOff`, or `sys.exit(1)` inside the container) | `helm upgrade` / `helm install`, with a message naming the accepted values |
| the fleet auto-upgrade | proceeds, edge stays broken | **the upgrade fails and the edge stays on its current release** |

That second row is the real risk and the reason I am not merging this myself:
the auto-upgrade CronJob runs `helm upgrade --reset-then-reuse-values`, so a
misconfigured edge's next auto-upgrade would start failing. It would fail
*visibly*, which is the point — but it is a behaviour change on live edges.

**I could not find such an edge from inside this repo:** every `CLIENT_ENV`
value anywhere in the tree is one of the six accepted spellings (the only
exceptions were two test fixtures deliberately exercising the pass-through, both
handled here). `arm-canary` is a published image tag but is referenced nowhere
in the repo as a `CLIENT_ENV`. **Confirming no live edge is out of vocabulary is
outside what I can check** — that needs a look at the actual installed values,
and it is the thing I would want checked before this merges.

If you'd rather not take that on now: findings **2 and 3 are non-breaking** and
can ship on their own.

## Tests

**The gates cannot be asserted from helm-unittest** — verified, not assumed: it
validates values against the packaged schema and reports a violation as a
plugin-level **error** rather than a template failure, so `failedTemplate`
cannot catch it, and it exposes no flag to skip validation and reach the
helper's `fail`. So they are exercised from
`scripts/tests/chart-env-vocabulary.sh`, wired into `make check` and the
`helm-ci` lint job. Every rejection is paired with an **accept control on the
same command line**, and matched against the specific error text rather than
merely a non-zero exit.

```
== env.CLIENT_ENV: the six accepted spellings plus empty all resolve ==
  ok    CLIENT_ENV unset -> tracebloc/jobs-manager:prod
  ok    CLIENT_ENV='' -> tracebloc/jobs-manager:prod
  ok    CLIENT_ENV=dev -> tracebloc/jobs-manager:dev
  ok    CLIENT_ENV=stg -> tracebloc/jobs-manager:stg
  ok    CLIENT_ENV=prod -> tracebloc/jobs-manager:prod
  ok    CLIENT_ENV=development -> tracebloc/jobs-manager:dev
  ok    CLIENT_ENV=staging -> tracebloc/jobs-manager:stg
  ok    CLIENT_ENV=production -> tracebloc/jobs-manager:prod
== env.CLIENT_ENV: everything else is refused at install time ==
  ok    CLIENT_ENV=prd (rejected: values don't meet the specifications of the schema)
  ok    CLIENT_ENV=qa (rejected: ...)
  ok    CLIENT_ENV=Prod (rejected: ...)
  ok    CLIENT_ENV=PROD (rejected: ...)
  ok    CLIENT_ENV=produktion (rejected: ...)
  ok    CLIENT_ENV=develop (rejected: ...)
  ok    CLIENT_ENV=stage (rejected: ...)
  ok    CLIENT_ENV=test (rejected: ...)
  ok    CLIENT_ENV=<tab> (rejected: ...)
== the template fail is a real backstop, not decoration ==
  ok    CLIENT_ENV=prd with the schema skipped (rejected: is not a recognized environment)
  ok    CLIENT_ENV=staging with the schema skipped -> tracebloc/jobs-manager:stg
== images.ingestor.channelTags: keys are closed ==
  ok    channelTags.dev -> value: "9.9"
  ok    channelTags.stg -> value: "9.9"
  ok    channelTags.prod -> value: "9.9"
  ok    channelTags.staging (rejected: ...)
  ok    channelTags.development (rejected: ...)
  ok    channelTags.production (rejected: ...)
  ok    channelTags.dev-1 (rejected: ...)
  ok    channelTags.PROD (rejected: ...)
  ok    channelTags.totallyBogus (rejected: ...)

chart-env-vocabulary: all 28 checks passed
```

**Mutation-checked** — each guard is independently load-bearing:

```
MUT A: drop the enum (keep the helper fail)          -> 9 of 28 checks FAILED   (= the 9 CLIENT_ENV rejects)
MUT B: drop additionalProperties on channelTags      -> 6 of 28 checks FAILED   (= the 6 bad keys)
MUT C: drop the helper fail (keep the enum)          -> 1 of 28 checks FAILED   (= the skip-schema backstop)
restored                                             -> all 28 checks passed
```

MUT C is the one that matters: it shows the backstop case is not just
re-testing the enum.

`make check` (shell parse + shellcheck + 3 drift guards + helm lint, 5 value files):

```
check-facts: all installer facts match scripts/spec/facts.env.
== tracebloc style guard ==
  ok: style + terminology clean
1 chart(s) linted, 0 chart(s) failed        (x5)
chart-env-vocabulary: all 28 checks passed
==> check: green
```

`make helm-unittest` — 5 new cases pin the mysql-client behaviour the corrected
description now describes:

```
Charts:      1 passed, 1 total
Test Suites: 30 passed, 30 total
Tests:       389 passed, 389 total
```

Mutation-checked too — someone "fixing" the template to match the old false doc
gets caught, and the prod control correctly stays green because prod is
indistinguishable either way:

```
MUT D: make the mysql tag env-derived, as the old doc claimed
  Tests: 2 failed, 387 passed, 389 total
    - an empty tag renders :prod on a dev edge, not :dev
    - an empty tag renders :prod on a stg edge, not :stg
```

`make helm-template` — all four platforms render; `shellcheck -S warning -x` on
the new script is clean.

**Manifest:** `scripts/gen-manifest.sh --check` passes and needs no
regeneration — the manifest is the installer bootstrap's integrity surface
(`scripts/install.sh`'s fetch list), and a CI-only test script is correctly not
in it. Confirmed by reading the `FILES` array rather than assuming.

## Type

- [x] Bug fix
- [x] **Breaking change** (finding 1 only)
- [x] Documentation

## Notes for the reviewer

- Chart version bumped 1.9.34 → 1.9.36 (`values.schema.json` is chart content
  per `scripts/chart-version-guard.sh`; I read the guard to confirm).
- **1.9.35 is taken by #694**, the sibling PR that makes the ingestor tag
  fallback per-environment. The two are independent; whichever lands second
  will want a trivial `Chart.yaml` rebase.
- One existing test case (`CLIENT_ENV: produktion`, asserting the pass-through)
  is replaced rather than re-pointed, because the *input* is now illegal. Its
  replacement deliberately asserts on **prod**, whose fallback is the same on
  this branch and on #694, so it does not depend on merge order.
- After both land, the `| default "0.8"` tail in #694's fallback becomes
  unreachable-by-construction. Left in as a belt-and-braces guard rather than
  coupling the two PRs; say the word if you'd rather it go.

---

> [!NOTE]
> **High Risk**
> Closing `CLIENT_ENV` is a breaking change: fleet auto-upgrade on edges with out-of-vocabulary values will fail at `helm upgrade` instead of proceeding with a broken cluster.
> 
> **Overview**
> **Closes undocumented open vocabularies** on the client Helm chart so bad `env.CLIENT_ENV` and `images.ingestor.channelTags` keys fail at install/upgrade instead of silently becoming wrong image tags, missed lookups, and dropped prod digest pins.
> 
> `values.schema.json` adds a **`CLIENT_ENV` enum** (six spellings + empty) and **`channelTags` with `additionalProperties: false`**. `tracebloc.clientEnv` adds a **`fail` backstop** when schema validation is skipped. The mysql-client **`tag` description** is corrected to match the template: empty tag always falls back to **`prod`**, not `CLIENT_ENV`.
> 
> **`scripts/tests/chart-env-vocabulary.sh`** exercises both gates via `helm template` (paired accept/reject controls); wired into **`make check`** / **`helm-ci`** and shellcheck. Helm unittest cases are updated (illegal `produktion` removed; mysql empty-tag → `:prod` pinned). Chart version **1.9.34 → 1.9.36**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8c3354630e764d369053c1802c5f792a0ca561bf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -10,6 +10,7 @@ on:
       - 'scripts/tests/e2e-seal-check.sh'
       - 'scripts/tests/e2e-full-seal.sh'
       - 'scripts/tests/lib/e2e-common.sh'
+      - 'scripts/tests/chart-env-vocabulary.sh'
       - 'scripts/lib/**'
       - '.github/workflows/helm-ci.yaml'
   pull_request:
@@ -21,6 +22,7 @@ on:
       - 'scripts/tests/e2e-seal-check.sh'
       - 'scripts/tests/e2e-full-seal.sh'
       - 'scripts/tests/lib/e2e-common.sh'
+      - 'scripts/tests/chart-env-vocabulary.sh'
       - 'scripts/lib/**'
       - '.github/workflows/helm-ci.yaml'
   # Manual runs — mainly to fire full-seal-e2e on demand (e.g. right after the
--- a/.github/workflows/helm-ci.yaml
+++ b/.github/workflows/helm-ci.yaml
@@ -65,6 +65,21 @@ jobs:
         # now packages both charts).
         run: helm lint --strict ./ingestor
 
+      - name: CLIENT_ENV + channelTags vocabularies are closed
+        # `helm lint` does not exercise the values.schema.json `enum` /
+        # `additionalProperties` against out-of-vocabulary inputs, and
+        # helm-unittest structurally cannot: it reports a schema violation as a
+        # plugin-level error rather than a template failure, and exposes no flag
+        # to skip validation and reach the `fail` in tracebloc.clientEnv. So
+        # both gates are asserted from outside the plugin, every rejection
+        # paired with an accept control on the same command line. ~2 s.
+        #
+        # The helper-backstop cases self-skip when helm predates
+        # --skip-schema-validation (3.16); the pin above is v3.15.4, so they are
+        # skipped here today and run for anyone local on 3.16+. Bumping the pin
+        # turns them on with no change to this file.
+        run: bash scripts/tests/chart-env-vocabulary.sh
+
   template:
     timeout-minutes: 10
     name: Template render
--- a/.github/workflows/installer-tests.yaml
+++ b/.github/workflows/installer-tests.yaml
@@ -77,11 +77,11 @@ jobs:
           # below for visibility but don't fail the gate.
           shellcheck --severity=error --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-full-seal.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-full-seal.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh scripts/tests/chart-env-vocabulary.sh
           echo "── shellcheck warnings (advisory, non-blocking) ──"
           shellcheck --severity=warning --shell=bash \
             scripts/install.sh scripts/install-k8s.sh scripts/gen-manifest.sh scripts/check-facts.sh scripts/check-style.sh scripts/resolve-ingestor-digest.sh scripts/lib/*.sh \
-            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-full-seal.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh || true
+            scripts/tests/check-drift.sh scripts/tests/distro-prereqs.sh scripts/tests/e2e-auto-upgrade.sh scripts/tests/e2e-seal-check.sh scripts/tests/e2e-full-seal.sh scripts/tests/e2e-cluster.sh scripts/tests/e2e-journey.sh scripts/tests/e2e-proxy.sh scripts/tests/lib/e2e-common.sh scripts/tests/path-persist.sh scripts/tests/chart-env-vocabulary.sh || true
 
       - name: Installer manifest is current (supply-chain, R8)
         # The bootstrap verifies each sub-script against scripts/manifest.sha256
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,8 @@ SHELLCHECK_FILES := \
 	scripts/tests/e2e-journey.sh \
 	scripts/tests/e2e-proxy.sh \
 	scripts/tests/lib/e2e-common.sh \
-	scripts/tests/path-persist.sh
+	scripts/tests/path-persist.sh \
+	scripts/tests/chart-env-vocabulary.sh
 
 # The bats total, DERIVED — never written down. It moves on most PRs that add a
 # test, nothing enforces it, and the help text had drifted from its hardcoded
@@ -57,7 +58,7 @@ help:
 	@echo "  setup       check for / point at the tools these targets need; installs the pre-push hook"
 	@echo "  install-hooks  (re)install the git pre-push hook that runs 'make check'"
 	@echo
-	@echo "  individual: lint bats helm-lint helm-template helm-unittest drift"
+	@echo "  individual: lint bats helm-lint helm-vocab helm-template helm-unittest drift"
 	@echo
 	@echo "  NOT here (CI-only, by name): the 9-distro prereq matrix, e2e-cluster"
 	@echo "  (k3d), e2e-proxy (squid), path-persist, Pester, windows-e2e,"
@@ -76,11 +77,11 @@ help:
 # someone does it. A `check` that quietly took two minutes would be a
 # `check` nobody runs.
 .PHONY: check
-check: lint drift helm-lint
+check: lint drift helm-lint helm-vocab
 	@echo "==> check: green (bats + helm unit tests are in 'make check-all')"
 
 .PHONY: check-all
-check-all: lint drift helm-lint bats helm-template helm-unittest
+check-all: lint drift helm-lint helm-vocab bats helm-template helm-unittest
 	@echo "==> check-all: green"
 
 # setup: no dependency is installed — this only tells you what is missing —
@@ -185,6 +186,16 @@ helm-lint:
 	  helm lint --strict ./client -f "$$f" || exit 1; \
 	done
 	helm lint --strict ./ingestor
+
+# helm-vocab: helm-ci.yaml `lint` job's vocabulary step. The CLIENT_ENV and
+# channelTags vocabularies are closed by a values.schema.json `enum` /
+# `additionalProperties` plus a `fail` in tracebloc.clientEnv, and helm-unittest
+# can assert NEITHER: it treats a schema violation as a plugin-level error
+# rather than a template failure, and offers no way to skip validation and reach
+# the helper. So the gates are exercised from outside the plugin. ~2 s.
+.PHONY: helm-vocab
+helm-vocab:
+	bash scripts/tests/chart-env-vocabulary.sh
 
 # helm-template: helm-ci.yaml `template`. kubeconform is pinned by
 # version AND digest in CI; rather than re-implement that download here,
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.34
-appVersion: "1.9.34"
+version: 1.9.36
+appVersion: "1.9.36"
 keywords:
   - tracebloc
   - kubernetes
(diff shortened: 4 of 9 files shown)


shujaatTracebloc (approved):
Approving. Verified all four Bugbot findings rather than trusting the resolve — three are genuinely fixed, and I'm accepting the fourth's deferral.

## Bugbot fixes, checked in the tree

1. **Help probe under pipefail** → now capture-then-match: `_helm_help="$(helm template --help 2>&1 || true)"` then `grep -q <<<"$_helm_help"`. No pipe, so no SIGPIPE misbranch. Fixed ✓
2. **Vocabulary test missing from CI paths** → added to **both** the `push` and `pull_request` filters, not just one ✓
3. **Shellcheck list drift** → `chart-env-vocabulary.sh` added to the Makefile's `SHELLCHECK_FILES` *and* to both invocations in `installer-tests.yaml` (error and advisory). Both sides in sync ✓
4. **Fallback test cannot prove path** → resolved as intended, not fixed. **I checked the claim it rests on and it holds:** `client/tests/ingestor_channel_tag_test.yaml:93` really does have a `images.ingestor.channelTags: null` case, so the fallback-fired path has independent coverage. Combined with helm-unittest's deep-merge making "one entry absent, others present" inexpressible, strengthening the `prod` case would couple it to #694's merge order for no new coverage. Reasonable trade — not reopening.

## Two things to handle at merge time

**The chart version assumes #694 lands first.** `develop` is at `1.9.34`; #694 → `1.9.35`, this → `1.9.36`. That only works in that order. If this merges first, #694 hits a conflict on the same `version`/`appVersion` lines and has to re-bump to `.37` — and the two PRs share three files (`Chart.yaml`, `_helpers.tpl`, `ingestor_channel_tag_test.yaml`), so the second one through needs a rebase regardless. Worth flagging because #694 is on a different reviewer, so nobody is watching the ordering from both sides.

**`additionalProperties: false` on `channelTags` is a breaking chart change.** The chart's own `values.yaml` only uses `dev`/`stg`/`prod`, so nothing here breaks itself ✓ — but the PR's own rationale is that `channelTags.staging: "0.8.2"` *"validated fine and was then ignored"*, which means it may exist in a live edge's values. Those edges now fail at `helm upgrade` rather than silently ignoring the key. That's the correct outcome and the point of the change, but it deserves a line in **Deployment notes** so a rollout doesn't discover it by surprise.

## On the design

Putting the enum in `values.schema.json` *and* a `fail` in `tracebloc.clientEnv` is right, and the reason given is the part that matters: the enum only binds where the packaged schema is read, and `--skip-schema-validation` or a repackage without the schema walks past it. Since the helper is already the single chokepoint every consumer goes through, the backstop costs nothing and closes the hole the enum can't.

The justification for `fail` over pass-through is the strongest argument in the PR — an unrecognised value wasn't graceful degradation, it silently reconfigured the edge in four places at once (three control-plane tags at unpublished names, a missed `channelTags` lookup, a missed `serviceDbAccountsByEnv` lookup, and a dropped prod digest pin, since the pin only applies where the env resolves to exactly `prod`). And noting that the only existing validator was `jobs_manager.py`'s `sys.exit(1)` *inside the container that can't start* is exactly why failing at `helm upgrade` is the right layer.

`chart-env-vocabulary.sh` earns being a shell test rather than a helm-unittest suite, and the reason is stated concretely (plugin 0.5.2 reports a schema violation as a plugin-level error, so `failedTemplate` can't assert it, and there's no flag to skip validation and reach the `fail`). Two details I'd have asked for and didn't have to:

- **every reject paired with an accept control on the same command line**, matched against specific error text rather than just non-zero exit — a rejection test that failed to render for an unrelated reason would otherwise pass for free
- **the `--skip-schema-validation` gate self-skips on the pinned helm 3.15.4 and turns on for anyone on 3.16+**, with the skip announced rather than silent, so bumping the pin activates it with no edit here

`BAD_ENVS` choosing `prd`, `Prod` and `develop` is well-judged — an abbreviation, a capitalisation, and an alias one letter off `development` are what people actually type. Keeping the whitespace-only case out of the array so the tab survives word-splitting is a nice touch.

The `mysql_test.yaml` additions are the sleeper value here: the schema said *"Empty falls back to env.CLIENT_ENV"* when the template is `| default "prod"`, and the five new cases pin why that doc had teeth beyond a failed pull — an env-derived tag is unrecognised by `tracebloc.mysqlEngineMajor`, so the mysql-format-guard stands down at `unknown` instead of staying armed at 5.7. Asserting the believing-the-old-doc case *and* the real-fallback control is what makes that provable rather than asserted.

PR #705, opened 2026-08-13
fix(chart): re-image control-plane pods by digest instead of relying on Always (#569)

shujaatTracebloc:
Closes #569. Splits out of #552 (the requests-proxy hardcoded-`Always` half landed in #570).

## The problem

The always-running control-plane pods rendered `imagePullPolicy: Always`, so a Docker Desktop / WSL2 restart forced a registry round-trip and landed in `ImagePullBackOff` **even with the image already cached in containerd**. The edge showed Offline until docker.io was reachable again — up to ~6h behind Docker Hub's anonymous pull-rate limit.

`Always` could not simply be flipped, because it *is* the update mechanism: image-refresh's `kubectl rollout restart` only picks up a new build because the pull policy re-resolves the floating tag. Offline-safety and restart-driven updates are mutually exclusive unless the image **reference** changes on update.

## The decision: option B (digest-on-update), not A (prodDigest pin)

The ticket framed both. B won on two grounds:

- **A is prod-only by construction** (gated on `CLIENT_ENV == "prod"`). The incident client runs dev, so A would not have fixed the reported incident.
- **A couples releases across repos.** These images ship from `client-runtime`, whose release train ran twice on 2026-08-12 and again on 08-13. Under A, every prod cut needs a `client` chart PR + lockstep version bump + chart publish before prod edges see it. `ingestor.prodDigest` shows how that ages: it is currently pinned to v0.8.2 while its own float resolves v0.8.4. For the ingestor that lag is deliberate (an ordering ceiling is being enforced); for jobs-manager there is no ceiling, so lag is just an unshipped fix.

The objections to B were weaker than the ticket implied. Auto-upgrade only runs `helm upgrade` when the published chart version is **newer** (`sort -V` compare, `auto-upgrade-cronjob.yaml`), so the re-render is not an hourly revert — it happens only on a chart release.

## What changed

**Pull policy.** All four call sites render `IfNotPresent` unconditionally — jobs-manager (`api` + the `pods-monitor-container` sidecar), requests-proxy, resource-monitor.

**Update path.** image-refresh swaps `kubectl rollout restart` for `kubectl set image <workload> <container>=repo@digest`, across three workloads instead of one:

| workload | image | before this PR |
|---|---|---|
| `deployment/<release>-jobs-manager` | jobs-manager + pods-monitor | reconciled (both containers now re-imaged in one patch → one rollout) |
| `deployment/<release>-requests-proxy` | **the same** jobs-manager image | **never reconciled** |
| `daemonset/<release>-resource-monitor` | resource-monitor | **never reconciled** |

Two bugs the ticket did not name, both fixed here:

- **requests-proxy was never refreshed.** The CronJob's `DEPLOYMENT_NAME` was only `<release>-jobs-manager`. So requests-proxy stayed on its old build until some unrelated restart, then jumped to whatever the tag pointed at — two pods routinely running different builds of one image. It now follows the jobs-manager digest in the same tick, with no second registry HEAD. `images.requestsProxy.digest` opts it out.
- **resource-monitor had no deliberate update path at all.** A chart release does not change its image ref either, so it only ever moved when a pod happened to restart. That is also why a naive flip to `IfNotPresent` would have frozen it permanently — it needed the reconcile to come with it.

**RBAC.** requests-proxy needed no new rule (same-namespace Deployment, already covered). resource-monitor needed a second Role in `tracebloc-node-agents` — the one genuinely new piece of trust surface here:

- `get` + `patch` are `resourceNames`-scoped to this chart's own DaemonSet. That namespace runs under the `privileged` PSA profile, so the CronJob cannot touch any *other* DaemonSet there.
- `list` + `watch` **cannot** be `resourceNames`-scoped (the authorizer ignores `resourceNames` for collection verbs) and `rollout status` needs both, so they stay namespace-wide and read-only. That residual widening is deliberate and asserted with `equal` so it cannot grow silently.
- Not rendered at all when `resourceMonitor: false`.

**Helper.** `imageRefreshEnabled` now requires all three refreshed images pinned before retiring the CronJob. Pinning only the two class-1 images used to render it away, which would have left the DaemonSet with no update path.

**Private mirrors.** The script resolves digests from docker.io, so pinning one onto a `global.imageRegistry` reference could pin an image the mirror does not hold. It logs and goes inert. Under `rollout restart` that mismatch was merely useless; with `set image` it has to fail closed.

## Bug caught during verification

`IMAGE_REGISTRY` first rendered as `""` — values.yaml ships `global.imageRegistry: ""`, so the key *exists* and `dig`'s default never applies. The mirror guard would then have seen `"" != "docker.io"` and gone inert on **every default install**, silently disabling auto-refresh fleet-wide. Fixed with a trailing `| default "docker.io"`; there is a test whose comment explains why that is not redundant.

## Deliberate trade-offs — please review these specifically

**The first-tick contract is kept.** Re-imaging on the first tick would rewrite `repo:tag` to `repo@digest` for byte-identical content on every fresh install — rolling the Deployment, and the DaemonSet on every node, for nothing. So a fresh edge runs `repo:tag` until the first real digest change: restart-safe offline (which is the point of this PR), just not yet reproducible.

**Two bounded limitations, documented in the script header rather than hidden.** Both self-heal at the next upstream release, and neither can break a running edge — `IfNotPresent` holds regardless.

1. **Helm re-render.** A chart version bump re-renders `repo:tag` and reverts the pin. This tick will *not* re-pin, because the annotation still records that digest and `recorded == latest` no-ops. Contrary to what the ticket predicted, refresh does not self-heal this; the edge floats until the next upstream release.
2. **Pre-existing skew is prevented, not repaired.** An edge upgrading *into* this version may already have requests-proxy and jobs-manager on different builds. They converge on the next digest change; this script never compares pod-vs-pod.

The proper fix for both is the same: reconcile against each workload's **live container image** instead of a shared annotation. `set image` makes that possible for the first time — `rollout restart` was a blind action, which is why the annotation exists at all. I kept it out of this PR to hold the diff to one change; happy to fold it in if you would rather have it here.

## Verification

```
helm unittest ./client
Tests: 5 failed, 5 errored, 405 passed, 410 total
```

405 passed, up from 395. The failures are **exactly** develop's pre-existing baseline (5 failed / 5 errored) — confirmed by diffing failing test names against a stashed baseline run on develop. **Zero new failures.**

Also green: `helm lint`, all four `client/ci/*-values.yaml` render, `scripts/check-style.sh`, `scripts/check-facts.sh`, `scripts/gen-manifest.sh --check`.

Chart bumped 1.9.38 → 1.9.39, `version` + `appVersion` in lockstep.

## Note

I could not read **#565**, referenced by the ticket as the installer-side context — it does not resolve in `client` or any other tracebloc repo I have access to. If it constrains this design, it did not feed into it.

---

> [!NOTE]
> **High Risk**
> Changes core control-plane image update mechanics, CronJob RBAC across namespaces, and pull policies—misconfiguration could freeze images or break refresh fleet-wide.
> 
> **Overview**
> Fixes **offline Docker/WSL restarts** that hit `ImagePullBackOff` because control-plane pods used **`imagePullPolicy: Always`**. Updates now come from **`kubectl set image repo@digest`** (replacing **`rollout restart`**) so pods can use **`IfNotPresent`** when a real update path exists.
> 
> **Image-refresh** reconciles **three workloads**: jobs-manager (+ pods-monitor containers), **requests-proxy** (same jobs-manager image, same digest in one tick), and **resource-monitor** (cross-namespace DaemonSet). Adds a **separate RBAC Role** in the node-agents namespace (`tracebloc.imageRefreshNodeAgentsName`) so Role names do not collide when node-agents shares the release namespace. **`requests-proxy`** falls back to **`images.jobsManager.digest`** when its own digest is unset; deployment name is centralized via **`tracebloc.requestsProxyName`**.
> 
> **`tracebloc.controlPlanePullPolicy`** unifies pull policy: **`IfNotPresent`** when digest-pinned or when refresh is enabled on **docker.io**; **`Always`** on private mirrors or **`imageRefresh.enabled: false`** so floating tags still update via restart. **`tracebloc.resourceMonitorRefreshPinned`** aligns CronJob retirement with runtime skip when **`resourceMonitor: false`** or digest pinned.
> 
> **Operational fixes**: mirror installs fail closed (script exits 0, reconcile inert); **`activeDeadlineSeconds`** default **3600** for three sequential rollouts; resource-monitor **`maxUnavailable: 10%`**. Chart **1.9.39**; schema/values/docs and helm unittest coverage expanded.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4d6015c49fada61ff5a8c3217b3168bf27c0a56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.38
-appVersion: "1.9.38"
+version: 1.9.39
+appVersion: "1.9.39"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/templates/_helpers.tpl
+++ b/client/templates/_helpers.tpl
@@ -187,6 +187,29 @@ mysql-pvc
 {{- end }}
 
 {{/*
+  Name of the requests-proxy Deployment. ONE definition, because #569 gave it a
+  second consumer: image-refresh reconciles it by name with `kubectl set image`,
+  so a rename that reached only the Deployment would leave the CronJob patching
+  a workload that does not exist — failing the tick, freezing the digest record,
+  and eventually tripping the shared flap lockout for every control-plane image.
+
+  This is the third instance on #569 of one side of a two-sided contract moving
+  without the other (the requests-proxy digest pin and the resource-monitor pin
+  signal were the first two, both found in review). Same remedy: collapse to a
+  single definition rather than keep two literals in sync by hand.
+
+  The jobs-manager Deployment has the same shape — image-refresh's
+  DEPLOYMENT_NAME re-derives `<release>-jobs-manager` with its own printf, and
+  five other call sites spell it out too (NOTES.txt, the PDB,
+  tracebloc.serviceAccountName, ...). Unifying THAT is a mechanical refactor
+  across templates this change does not otherwise touch, so it is deliberately
+  left for its own PR; a contract test pins the two sides here in the meantime.
+*/}}
+{{- define "tracebloc.requestsProxyName" -}}
+{{ .Release.Name }}-requests-proxy
+{{- end }}
+
+{{/*
   Whether the image-refresh CronJob has anything to do. When ALL THREE
   managed images (jobs-manager, pods-monitor, resource-monitor) are
   digest-pinned, the operator has explicitly opted into reproducible
--- a/client/templates/_helpers.tpl
+++ b/client/templates/_helpers.tpl
@@ -165,16 +165,18 @@ mysql-pvc
 
 {{/*
   Whether the image-refresh CronJob has anything to do. When ALL THREE
-  managed images (jobs-manager, pods-monitor, ingestor) are
+  managed images (jobs-manager, pods-monitor, resource-monitor) are
   digest-pinned, the operator has explicitly opted into reproducible
   pinning for every image this CronJob would refresh, so we render
   nothing — no CronJob, no RBAC, no ConfigMap. When at least one is
   unpinned, the CronJob is rendered and the script skips the pinned
   images at runtime via env flags.
 
-  Three pins because #158 added ingestor refresh on top of #154's
-  jobs-manager + pods-monitor. Keep this list in sync if more images
-  come under auto-refresh in future.
+  The three have changed over time: #154 started with jobs-manager +
+  pods-monitor, #158 added the ingestor, the floating-tag migration
+  retired the ingestor pass, and #569 brought resource-monitor under
+  refresh (it had no deliberate update path before). Keep this list in
+  sync if more images come under auto-refresh in future.
 
   Nil-guarded with `default dict` on every dereference: these are
   newer top-level keys, and a customer who runs
@@ -203,8 +205,18 @@ mysql-pvc
 */}}
 {{- $jmPinned := $jm.digest -}}
 {{- $pmPinned := $pm.digest -}}
+{{/*
+  #569: resource-monitor came under refresh too, so it joins the "nothing left
+  to do" test. requests-proxy deliberately does NOT: it runs the SAME
+  tracebloc/jobs-manager image and follows the jobs-manager digest, so
+  `$jmPinned` already covers it. (`images.requestsProxy.digest` remains an
+  operator override that pins requests-proxy alone; it is checked at runtime,
+  not here, because pinning only requests-proxy still leaves jobs-manager
+  itself to refresh.)
+*/}}
+{{- $rmPinned := (default dict $imgs.resourceMonitor).digest -}}
 {{- if not $ir.enabled -}}
-{{- else if and $jmPinned $pmPinned -}}
+{{- else if and $jmPinned $pmPinned $rmPinned -}}
 {{- else -}}
 true
 {{- end -}}
--- a/client/templates/_helpers.tpl
+++ b/client/templates/_helpers.tpl
@@ -213,6 +213,14 @@ mysql-pvc
   operator override that pins requests-proxy alone; it is checked at runtime,
   not here, because pinning only requests-proxy still leaves jobs-manager
   itself to refresh.)
+
+  "Follows the jobs-manager digest" is only true because
+  requests-proxy-deployment.yaml FALLS BACK to `images.jobsManager.digest` when
+  its own key is empty. Without that fallback this test is a silent trap
+  (Bugbot): retiring the CronJob here on a jobs-manager pin would leave the
+  proxy rendering the floating tag with nothing left to reconcile it, running a
+  different build of the same image forever. If that fallback is ever removed,
+  requests-proxy must get its own entry in this test.
 */}}
 {{- $rmPinned := (default dict $imgs.resourceMonitor).digest -}}
 {{- if not $ir.enabled -}}
--- a/client/templates/_helpers.tpl
+++ b/client/templates/_helpers.tpl
@@ -231,6 +231,62 @@ true
 {{- end }}
 
 {{/*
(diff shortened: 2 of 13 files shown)


saqlainsyed007 (approved):
**APPROVE** — reviewed for correctness; no defects found.

### What this does
Fixes the offline-restart `ImagePullBackOff` (#569) by dropping `imagePullPolicy: Always` on the four always-running control-plane images in favour of `IfNotPresent` **only where a non-`Always` update path exists**, and swaps image-refresh's action from `kubectl rollout restart` to `kubectl set image repo@digest` across three workloads (jobs-manager+pods-monitor, requests-proxy, resource-monitor).

### Correctness verification
Traced the load-bearing logic end-to-end; all internally consistent:
- **`tracebloc.controlPlanePullPolicy`** — `IfNotPresent` on a digest pin, or when refresh can actually reconcile the image (`imageRefreshEnabled` ∧ docker.io); `Always` otherwise. No path renders `IfNotPresent` for an image with no update mechanism (verified across mirror / `imageRefresh.enabled:false` / per-image-pin permutations, incl. requests-proxy's jobs-manager-digest fallback).
- **Mirror guard** fails closed (exit 0, reconcile inert) and the pull policy stays `Always` there — the two agree, so no frozen-control-plane-with-green-CronJob.
- **`IMAGE_REGISTRY` / `$mirror`** both carry the `| default "docker.io"` needed because `global.imageRegistry: ""` exists (dig's default never applies) — consistent in template and env.
- **Cross-namespace RBAC**: second Role in node-agents is `resourceNames`-scoped for get/patch; list/watch namespace-wide+read-only (required by `rollout status`, cannot be name-scoped); RoleBinding correctly references the release-namespace SA; distinct name avoids the collapsed-namespace collision.
- **DaemonSet `maxUnavailable: "10%"`** — confirmed against k8s DaemonSet controller behaviour that it floors at 1 when the percentage rounds to 0, so small clusters keep one-at-a-time (the comment's claim is accurate).
- **`activeDeadlineSeconds` 3600** ≥ 3×rolloutTimeout with slack; `set image` idempotency makes partial-tick retries safe; shared flap-counter blast radius is documented as a deliberate tradeoff.

All prior cursor[bot] findings (refresh budget, jobs-manager-pin-skips-proxy, mirror update path, disabled-monitor CronJob, RBAC name collision) are confirmed fixed and match the shipped code.

### Evidence
- `helm unittest ./client`: **431/431 passed** on this branch (develop baseline 402/402; +29 tests, zero failures). Note: the PR body's "5 failed / 5 errored baseline" is stale — both develop and this branch are fully green now.
- `scripts/check-style.sh`: clean · `scripts/check-facts.sh`: clean
- `helm lint` shows only the pre-existing `clientId`/`clientPassword` minLength error (bare lint without secrets), identical on develop — unrelated to this PR.

saadqbal (approved):
Very careful PR 👍 Every long-running control-plane container (jobs-manager api + pods-monitor-container, requests-proxy, resource-monitor) now goes through tracebloc.controlPlanePullPolicy — IfNotPresent where an update path exists (digest pin, or a reconcile that can actually run), Always kept exactly on the edges that still need it (private mirror, imageRefresh disabled). Nothing left on unconditional Always, so the offline-restart goal is genuinely met. The node-agents Role is the only new trust surface and it's as tight as RBAC allows (get/patch resourceNames-scoped to the one DaemonSet; list/watch namespace-wide read-only only because rollout-status needs them), the distinct name avoids the same-namespace collision, and the CronJob can't thrash — set image is idempotent and the shared flap lockout still backstops a bad digest. New keys are nil-guarded/defaulted and the schema only adds descriptions + one optional key, so reuse-values upgrades are safe. helm unittest green: 431/431 across the whole chart (image_refresh 36, jobs_manager 46, requests_proxy 20, resource_monitor 11). Version bumped 1.9.38→1.9.39.

Issue #85, opened 2026-04-29
Ingestor Update — declarative YAML, official image, scoped tokens


saadqbal:
## Goal

Redesign data ingestion to be declarative, secure, and future-proof. Customers should be able to ingest a new dataset by writing an `ingest.yaml` and running `helm install` — no Dockerfile, no `docker push`, no user-built images for the common path.

## Why now

Current flow (per-customer Dockerfile + image build + manual `kubectl apply` of [ingestor-job.yaml](https://github.com/tracebloc/data-ingestors/blob/develop/ingestor-job.yaml)) has accumulated:

- Hardcoded credentials in [tracebloc_ingestor/config.py](https://github.com/tracebloc/data-ingestors/blob/develop/tracebloc_ingestor/config.py)
- `:latest` tag and unsigned images in the job manifest
- Username/password auth at job start (training pods already use the better pre-minted-token pattern via jobs-manager)
- `data_id` column-mapping that can leak source PII (e.g., `patient_id`) to the central backend
- Silent passthrough of regression target values (`label` field for regression / time-series / time-to-event tasks is the prediction target itself, not metadata)
- Per-dataset images = supply-chain surface that's hard to scan, sign, and govern

## Architecture (locked)

1. **Declarative YAML config** with a Python escape hatch for genuine edge cases.
2. **PVC stays** as the data plane (no object-store migration in this epic).
3. **Option A**: jobs-manager (in `client-runtime`) owns ingestor Job creation and mints scoped tokens. The Helm chart becomes a thin client that submits the run to jobs-manager.
4. **`data_id` defaults to UUID**; mapping a source column requires explicit opt-in in the YAML.
5. **`label_policy` field, required for regression-class tasks**:
   - `bucket` (recommended): bin index + boundaries; default `strategy: quantile`, `bins: 10`
   - `passthrough`: customer explicitly accepts that exact target values leave the cluster
   - Classification / object-detection / text-classification: defaults to `passthrough` (categorical metadata, no behavior change)

## What does NOT change

- PVC + cluster-local MySQL remain the data plane.
- The six existing central-backend endpoints stay as they are; the token minted by jobs-manager is scoped exactly to those.
- Customer feature columns and file contents continue to never leave the cluster.

## Cross-cutting acceptance criteria

- A customer can ingest a new dataset with: `helm install my-dataset tracebloc/ingestor -f ingest.yaml`. No Dockerfile.
[...]

PR #726, opened 2026-08-14
fix(preflight): match FUSE network mounts by normalised name, not by spelling (client#725)

### LINKED ISSUE #725: preflight: the network-FS guard's FUSE entries are unreachable on macOS and on the stat fallback
## What

`_pf_is_network_fstype` in `scripts/lib/preflight.sh` — the classifier shared by the full storage check (`_pf_storage_type`) and the pre-log `early_data_dir_guard` (#432) — matches the FUSE network filesystems **only** in their `fuse.`-prefixed spelling:

```
fuse.glusterfs  fuse.ceph  fuse.sshfs  fuse.s3fs  fuse.davfs  fuse.rclone
```

`glusterfs`, `ceph` and `davfs` also have bare entries; `sshfs`, `s3fs` and `rclone` do not.

That matters because the prefix depends on **which reader in `_pf_fstype` answered**, and two of its three readers never produce a `fuse.`-prefixed name at all — so on those platforms the FUSE entries are unreachable and a network `HOST_DATA_DIR` is accepted, which is exactly what the guard exists to prevent (MySQL/InnoDB corrupts or crash-loops on network storage).

## Measured

Mounted each filesystem for real and read it back through each of `_pf_fstype`'s readers:

| reader (in `_pf_fstype` order) | sshfs | rclone | bindfs |
|---|---|---|---|
| `findmnt -nro FSTYPE` (util-linux) | `fuse.sshfs` ✅ caught | `fuse.rclone` ✅ | `fuse` ❌ missed |
| `stat -f -c %T` (Linux fallback when findmnt is absent) | `fuseblk` ❌ | `fuseblk` ❌ | `fuseblk` ❌ |
| `df` + `mount` (the macOS path) | `macfuse` ❌ | `macfuse` ❌ | `macfuse` ❌ |

Environments: Ubuntu 24.04 (libfuse 3.14.0, util-linux 2.39.3, coreutils 9.4) and Ubuntu 20.04 (libfuse 2.9.9) — identical answers on both. macOS row from macFUSE's own mount output format, e.g. `sshfs#user@host:/d on /mp (macfuse, nodev, nosuid, synchronous, …)`; `_pf_fstype`'s awk takes the first token inside the parens.

Three findings, in order of severity:

1. **macOS is blind to every FUSE entry.** macOS has no `findmnt`, and the `stat` reader is skipped there on purpose (BSD `stat -f` is a format string), so the `df`+`mount` path always answers — and it reports the macFUSE vfs name `macfuse` (`osxfuse` before 4.x), never a subtype. None of the six `fuse.*` entries can ever match on macOS.
2. **`stat -f -c %T` erases the subtype.** It answers `fuseblk` for *every* FUSE mount (the `fuse` and `fuseblk` kernel filesystems share one magic number), so wherever `findmnt` is missing, sshfs/s3fs/rclone all read back as `fuseblk` — unmatched.
3. **A subtype-less FUSE mount reports bare `fuse`** even under `findmnt` — reproduced with `bindfs`; libfuse only defaults the subtype to the program name, it is not guaranteed.

Note the original hypothesis (a bare `sshfs` in the mount table) did **not** reproduce: on libfuse 2.9.9 and 3.14.0 alike, sshfs and rclone always set a subtype. The list's prefixed spellings are correct for `findmnt` — the defect is that they are the *only* spellings, and the other two readers never speak that language.

## Fix

- Normalise: strip a leading `fuse.` before matching, so the case list carries **bare names only** and cannot drift into "prefixed entry with no bare twin" again.
- Classify `fuse` / `fuseblk` / `macfuse` / `osxfuse` as *opaque FUSE* — FUSE with the subtype erased. These cannot be resolved to network-vs-local (sshfs and a local ntfs-3g/gocryptfs read identically), so they **warn** rather than hard fail: refusing every unnamed FUSE mount would block legitimate local installs, and `TRACEBLOC_ALLOW_NETWORK_FS` is nonsense advice to someone who is not on a network filesystem. Silence is what let them through.

Found while fact-checking the guard's documentation in tracebloc/e2e-test-agent#125, which notes the gap but correctly leaves the fix here.

LukasWodka:
Fixes tracebloc/client#725

## Summary

`_pf_is_network_fstype` (shared by `_pf_storage_type` and the pre-log `early_data_dir_guard`, #432) matched the FUSE network filesystems **only** as `fuse.sshfs` / `fuse.s3fs` / `fuse.rclone` / `fuse.glusterfs` / `fuse.ceph` / `fuse.davfs` — and three of those six had no bare twin.

Which spelling arrives depends on **which reader in `_pf_fstype` answered**, and two of its three readers never emit a `fuse.` prefix. Measured by mounting each filesystem for real on Ubuntu 24.04 (libfuse 3.14.0, util-linux 2.39.3, coreutils 9.4) and Ubuntu 20.04 (libfuse 2.9.9) — identical on both:

| reader (in `_pf_fstype` order) | sshfs | rclone | bindfs |
|---|---|---|---|
| `findmnt -nro FSTYPE` | `fuse.sshfs` ✅ | `fuse.rclone` ✅ | `fuse` ❌ |
| `stat -f -c %T` (Linux fallback, no findmnt) | `fuseblk` ❌ | `fuseblk` ❌ | `fuseblk` ❌ |
| `df` + `mount` (**the macOS path**) | `macfuse` ❌ | `macfuse` ❌ | `macfuse` ❌ |

So **on macOS none of the six entries can ever match** — macOS has no `findmnt` and deliberately skips the `stat` reader (BSD `stat -f` is a format string), so the `df`+`mount` path always answers, with the macFUSE vfs name. And wherever `findmnt` is absent on Linux, `stat -f -c %T` collapses every FUSE mount to `fuseblk` (the `fuse` and `fuseblk` kernel filesystems share one magic number). In both cases the guard read "Local storage" and MySQL was installed onto exactly the filesystem it exists to reject.

The originally-suspected mechanism — a bare `sshfs` in the mount table — did **not** reproduce on either libfuse generation; sshfs and rclone always set a subtype. The prefixed spellings are right for `findmnt`; the defect is that they were the *only* spellings.

## Change

- **Normalise, don't enumerate.** Strip a leading `fuse.` before matching; the case list now carries **bare names only**, so it cannot regrow a prefixed entry with no bare twin. A guard test asserts the list contains no `fuse.`-prefixed entry, and fails closed if it can't read the function.
- **Opaque FUSE** (`fuse`, `fuseblk`, `macfuse`, `osxfuse`) — FUSE with the subtype erased — gets its own predicate and **warns instead of hard failing**. sshfs and a local ntfs-3g/gocryptfs are indistinguishable at that point, so a refusal would block legitimate local installs, and `TRACEBLOC_ALLOW_NETWORK_FS` reads as nonsense to someone who is not on a network filesystem. The silent "Local storage" pass was the defect; the warn names what it might be and continues.

Deliberately **not** in this PR: recovering the real subtype when a reader degrades (readable from `/proc/mounts` on Linux and from the `sshfs#…` device field on macOS). That would let the named cases hard fail on every platform instead of only warning. Happy to file it as a follow-up.

## Test plan

New tests write their input filesystem names down **from the measured reader outputs, not from the case list** — a list checked against itself is self-consistent and blind. The measurement table is recorded in the test file header.

- `bats scripts/tests/preflight.bats` — **134/134 pass** (5 new tests, 0 failures)
- `bats scripts/tests/bats-hygiene.bats` — green; every new assertion is `|| return 1`-hardened
- `bats scripts/tests/copy-catalog.bats` — green (no user-facing copy in the goldens changed)
- `make check` — green; `make check-all` — green (465 helm unit tests, 31 suites)
- `scripts/gen-manifest.sh` re-run; `scripts/manifest.sha256` committed (R8 gate)

**Mutation-proof** — each mutation was verified to have actually applied (anchor gone) before trusting the result:

| mutation | result |
|---|---|
| drop the `fuse.` strip | 5 tests red |
| regrow the original bug (`fuse.sshfs` entry, no bare `sshfs`) | 7 tests red, incl. the no-prefix structural guard |
| opaque predicate always false | 3 tests red |
| **restore the classifier exactly as it ships on `develop`** | **5 of the new tests red** |

The last row is the one that matters: the tests fail against the real pre-fix code, not just against a synthetic mutation.

Found while fact-checking tracebloc/e2e-test-agent#125, which documents this guard and flags the gap without fixing it (the defect is here, not there).

---

> [!NOTE]
> **Medium Risk**
> Changes install-time refusal/warning for MySQL data paths—affects whether installs proceed on FUSE/macOS/Linux—but behavior is well-tested and opaque FUSE stays advisory rather than blocking.
> 
> **Overview**
> Fixes **HOST_DATA_DIR** preflight so network FUSE mounts are caught regardless of how `_pf_fstype` spells them, and generic FUSE no longer passes as “local storage.”
> 
> **`_pf_is_network_fstype`** now lowercases and strips a leading `fuse.` before matching a **bare-name** allowlist (adds **gcsfuse**, **blobfuse**, **blobfuse2**, etc.). Unlisted `fuse.*` subtypes are no longer treated as local.
> 
> **`_pf_is_opaque_fuse_fstype`** and **`_pf_opaque_fuse_warn`** handle subtype-erased readings (`fuseblk`, `macfuse`, bare `fuse`) and unknown named FUSE drivers: **warn** and continue (not hard-fail), with wording that distinguishes “subtype unknown” vs “driver not recognised.” **`early_data_dir_guard`** and **`_pf_storage_type`** use the same classifiers.
> 
> **`docs/INSTALL.md`** checklist and **`scripts/manifest.sha256`** are updated; **`scripts/tests/preflight.bats`** adds broad coverage (prefix invariance, mutual exclusivity, guard agreement).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f802292ec4ceca7354f4bd84fd2694c10098af7f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/preflight.sh
```diff
@@ -578,13 +578,58 @@ _pf_disk() {
...
+# list twice. (Measured on Ubuntu 24.04 / libfuse 3.14 / util-linux 2.39 and
+# Ubuntu 20.04 / libfuse 2.9: sshfs and rclone both mount as `fuse.<subtype>`.)
 _pf_is_network_fstype() {
-  case "$1" in
-    nfs|nfs3|nfs4|nfsd|cifs|smb|smbfs|smb2|smb3|afpfs|9p|ncpfs|gfs|gfs2|ocfs2|lustre|glusterfs|fuse.glusterfs|ceph|fuse.ceph|beegfs|fuse.sshfs|fuse.s3fs|davfs|fuse.davfs|webdav|fuse.rclone) return 0 ;;
+  local t
+  t="$(printf '%s' "${1:-}" | tr '[:upper:]' '[:lower:]')"
+  t="${t#fuse.}"                       # fuse.sshfs -> sshfs; nfs4 -> nfs4
+  case "$t" in
+    nfs|nfs3|nfs4|nfsd|cifs|smb|smbfs|smb2|smb3|afpfs|9p|ncpfs|gfs|gfs2|ocfs2|lustre|glusterfs|ceph|beegfs|sshfs|s3fs|davfs|webdav|rclone) return 0 ;;
```
saadqbal: Optional, non-blocking: a named FUSE cloud mount that isn't on this list still slips to "Local storage" silently — e.g. `fuse.gcsfuse` (GCS) or `fuse.blobfuse2` (Azure), the cloud twins of the `s3fs` you already cover. The bare-`fuse` reporters are caught by the opaque-warn tier, so this only bites the ones that set a subtype. Worth adding `gcsfuse`/`blobfuse`/`blobfuse2` while you're here, but the common trio (sshfs/s3fs/rclone) is handled and nothing here regresses, so not a blocker.

saadqbal (dismissed):
Nice, careful PR 👍 Normalise-then-match is the right shape, and the opaque-FUSE warn tier is a good call — refusing every unnamed `fuse`/`fuseblk`/`macfuse` would block legit local FUSE. Tests are non-vacuous (names come from real reader output, and #85 greps the source to stop the `fuse.`-prefix regrowing). Ran the suite locally: 134/134 green, manifest hash matches. One optional nit inline.

scripts/lib/preflight.sh
LukasWodka: Not optional in the end — you were right that it bites, and taking it turned up two things worth more than the three names.

**Added** `gcsfuse`, `blobfuse` and `blobfuse2` to the network list.

**And the structural half**, because a list only ever holds the drivers someone thought of and the next cloud FUSE ships without asking us: an unrecognised `fuse.<subtype>` is now **opaque** rather than local. We have no evidence it is network — so it stays a warning, never a hard fail — but we have no evidence it is local either, and "Local storage" was a confident answer about a mount we could not identify.

**Your existing suite caught two things while I did it, both more interesting than the change:**

**1 · The two predicates are mutually exclusive by design, and bats asserts it.** A bare `fuse.*) return 0` broke that — `fuse.sshfs` satisfied *both*, and the only thing keeping it a hard fail rather than a notice was the order the caller happens to ask in. A future caller asking "opaque?" first would have silently downgraded a database-corrupting mount to a notice. The arm now negates the network classifier, so the answer is independent of call order. That test earned its keep.

**2 · The opaque warning would have become a false sentence.** *"the mount table doesn't say which one"* is true for `fuseblk` and **false** for `fuse.gocryptfs`, where the table says precisely which one and the gap is that *we* have no opinion. The wording now branches on which unknown it is.

Verified end to end through the real predicates:

```
fuse.gcsfuse    NETWORK hard fail      fuse.gocryptfs  opaque warn
fuse.blobfuse2  NETWORK hard fail      fuse.ntfs-3g    opaque warn
fuse.s3fs       NETWORK hard fail      fuseblk         opaque warn
fuse.sshfs      NETWORK hard fail      ext4            local success
mutually exclusive on every shape: yes
```

138 bats pass (was 134). Four mutations, each anchor asserted unique first, all reddened — including restoring the false wording and dropping the exclusivity negation. `manifest.sha256` regenerated.

**One judgement call to overrule if you disagree:** local FUSE (`ntfs-3g`, `gocryptfs`) moves from a silent pass to a **notice**. I think that is the honest reading — we genuinely cannot tell those from a cloud mount by name alone — and it is a warning, never a failure. If you would rather ship only the three names, the `fuse.*` arm is a one-line revert and I will take it out.

LukasWodka:
/fr-pass

PR #731, opened 2026-08-17
sec(chart): drop unused cluster-wide pods/log from resource-monitor (backend#950)

saadqbal:
Finding 3 of 4 in tracebloc/backend#950 (epic tracebloc/backend#1078).

The ticket filed this against `client-runtime` `Node-deploy/rabc.yaml`, which no longer exists — those manifests moved into this chart in client-runtime#155. Re-triaged against the chart as it stands today.

## What was actually still wrong

The resource-monitor ClusterRole granted `get/list/watch` on `pods/log` and `namespaces` cluster-wide. The DaemonSet image never calls either.

`pods/log` is the one that matters: cluster-wide it lets the resource-monitor ServiceAccount read the **log contents of every pod in the cluster** — including other tenants' training pods, which run customer-submitted code and print customer data. That's a real exfiltration path for an agent whose job is node telemetry and which never reads a log.

## What the image actually calls

The image (`Node-deploy/Dockerfile`) ships exactly `resource_monitor.py`, `resource_utils.py`, `constants.py`, `proxy_config.py`. Their entire Kubernetes surface:

| call | resource | verb |
|---|---|---|
| `read_namespaced_pod` | `pods` | get |
| `list_pod_for_all_namespaces` | `pods` | list |
| `read_node` / `read_node_status` | `nodes`, `nodes/status` | get |
| `list_namespaced_custom_object` | `metrics.k8s.io/pods` | list |
| `get_cluster_custom_object` | `metrics.k8s.io/nodes` | get |

So the core rule becomes `["pods", "nodes", "nodes/status"]`. The `metrics.k8s.io` rule is **untouched** — the telemetry reads the image does make must survive the trim, and the asserts cover that explicitly.

Verbs stay `get/list/watch`. `watch` is also unused, but trimming verbs is a wider blast radius than this finding needs — happy to do it in a follow-up if you'd prefer.

The separate controller/jobs-manager role in `templates/rbac.yaml` **keeps** its `pods/log`: `controller.py:465` genuinely calls `read_namespaced_pod_log`. Left alone deliberately.

## Already fixed, no action needed

The same #950 bullet also claimed "DaemonSet tolerates all taints; host /proc,/sys mounts; no securityContext". Against current `develop` that's mostly stale:

- pod: `runAsNonRoot: true` + `seccompProfile: RuntimeDefault` ✅
- container: `allowPrivilegeEscalation: false` + `capabilities.drop: [ALL]` ✅
- `/proc` and `/sys` hostPath mounts are `readOnly: true` ✅ and are inherent to node telemetry
- `tolerations: operator: Exists` is by design — a node-telemetry DaemonSet that skips tainted nodes has blind spots

I'll note that on the ticket rather than churn the chart.

## Verification

```
make helm-unittest    31 suites, 466 tests passed
make check            green
```

Added a helm-unittest case pinning the resource list and asserting `pods/log` / `namespaces` are absent. Verified it **fails** against the pre-trim rule:

```
FAIL  Resource-monitor DaemonSet
  - ClusterRole grants no pod-log or namespace read
    asserts[1] `equal` fail — Path: rules[0].resources
```

## Upgrade note

This removes permissions from a live ClusterRole. Safe by construction — the removed grants are uncalled — but it does mean a `helm upgrade` narrows the SA's access, so it's worth landing before the next fleet rollout rather than alongside unrelated changes. No version bump here; that's the release train's call.

---

> [!NOTE]
> **Medium Risk**
> Live ClusterRole permissions shrink on upgrade, which is safe because the removed grants are unused but should be rolled out deliberately so fleet SA access narrows as intended.
> 
> **Overview**
> **Narrows the resource-monitor ClusterRole** so it only grants what the DaemonSet image actually uses: core `pods`, `nodes`, and `nodes/status`, plus the existing `metrics.k8s.io` rule unchanged.
> 
> **Removes cluster-wide `pods/log` and `namespaces` read access**, which were never called by the monitor but allowed the ServiceAccount to read every pod’s log contents cluster-wide (a tenant data exfiltration path for a node-telemetry agent). Template comments document the allowed API surface; a helm-unittest case pins `rules[0].resources` and blocks re-adding `pods/log` or `namespaces`.
> 
> Chart version bumps to **1.9.45**. Controller/jobs-manager RBAC that legitimately uses `pods/log` is intentionally untouched.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e6ff60378f70bde09aca347d0081cbee6dda242. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/templates/resource-monitor-rbac.yaml
+++ b/client/templates/resource-monitor-rbac.yaml
@@ -40,9 +40,24 @@ metadata:
   labels:
     {{- include "tracebloc.labels" . | nindent 4 }}
     app: {{ include "tracebloc.resourceMonitorName" . }}
+{{/*
+  Keep this rule matched to what the DaemonSet image actually calls. The image
+  (Node-deploy/Dockerfile) ships exactly resource_monitor.py, resource_utils.py,
+  constants.py and proxy_config.py, whose entire Kubernetes surface is:
+    read_namespaced_pod            -> pods            get
+    list_pod_for_all_namespaces    -> pods            list
+    read_node / read_node_status   -> nodes, nodes/status  get
+    list_namespaced_custom_object  -> metrics.k8s.io/pods  list
+    get_cluster_custom_object      -> metrics.k8s.io/nodes get
+  `pods/log` and `namespaces` were granted but never called; `pods/log` is
+  removed because cluster-wide it lets this ServiceAccount read the LOG CONTENTS
+  of every pod in the cluster -- including other tenants' training pods, which
+  run customer code and print customer data. That is a real exfiltration path
+  for a node-telemetry agent that never reads a log. backend#950.
+*/}}
 rules:
   - apiGroups: [""]
-    resources: ["pods", "pods/log", "nodes", "nodes/status", "namespaces"]
+    resources: ["pods", "nodes", "nodes/status"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["metrics.k8s.io"]
     resources: ["pods", "nodes"]
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.44
-appVersion: "1.9.44"
+version: 1.9.45
+appVersion: "1.9.45"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/tests/resource_monitor_test.yaml
+++ b/client/tests/resource_monitor_test.yaml
@@ -84,6 +84,34 @@ tests:
           path: metadata.name
           value: stg-resource-monitor
 
+  # backend#950: the ClusterRole granted cluster-wide `pods/log`, which lets the
+  # resource-monitor SA read the log CONTENTS of every pod in the cluster —
+  # other tenants' training pods included — while the image never reads a log.
+  # `namespaces` was likewise never called. Pin the rule to what the image
+  # actually calls so neither can be re-added without a deliberate edit here.
+  - it: ClusterRole grants no pod-log or namespace read
+    template: templates/resource-monitor-rbac.yaml
+    documentIndex: 1
+    asserts:
+      - isKind:
+          of: ClusterRole
+      - equal:
+          path: rules[0].resources
+          value: ["pods", "nodes", "nodes/status"]
+      - notContains:
+          path: rules[0].resources
+          content: pods/log
+      - notContains:
+          path: rules[0].resources
+          content: namespaces
+      # The telemetry reads the image DOES make must survive the trim.
+      - equal:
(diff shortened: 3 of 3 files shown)


client/templates/resource-monitor-rbac.yaml
```diff
@@ -40,9 +40,24 @@ metadata:
...
+  removed because cluster-wide it lets this ServiceAccount read the LOG CONTENTS
+  of every pod in the cluster -- including other tenants' training pods, which
+  run customer code and print customer data. That is a real exfiltration path
+  for a node-telemetry agent that never reads a log. backend#950.
+*/}}
 rules:
   - apiGroups: [""]
-    resources: ["pods", "pods/log", "nodes", "nodes/status", "namespaces"]
+    resources: ["pods", "nodes", "nodes/status"]
     verbs: ["get", "list", "watch"]
```
saadqbal: Same logic that justifies dropping pods/log and namespaces applies to `watch`: the image only does one-shot get/list (no watch=True or Watch() anywhere in the source), so watch is granted-but-unused too. Out of scope here -- just noting it, since the comment above pins the rule to "what the image actually calls" and watch isn't one of them. Fine as a later follow-up.

saadqbal (commented):
Clean least-privilege drop, and the safety case holds up. I traced the image source in client-runtime/Node-deploy (built into tracebloc/resource-monitor by build-images.sh via Node-deploy/Dockerfile): the entire k8s surface is read_namespaced_pod, list_pod_for_all_namespaces, read_node / read_node_status, and the two metrics.k8s.io reads. read_namespaced_pod_log (pods/log) and list/read_namespace (namespaces) are never called -- the pod's own namespace is read off the SA token mount, not the API -- so the grant really is dead, and the retained rules still cover every call. helm unittest green (12/12), and the notContains guards will catch a re-add. The other pods/log/namespaces grants in the tree are jobs-manager / auto-upgrade, correctly untouched. One optional follow-up inline.

client/templates/resource-monitor-rbac.yaml
saadqbal: Agreed on all counts — `watch` is granted-but-unused by the same evidence, and you're right that the comment pins the rule to "what the image actually calls" while leaving a verb that isn't one of them. Slightly inconsistent as it stands.

Keeping it out of this PR as you suggest: removing resources closes a real exfil path, whereas trimming a verb is tidiness, and I'd rather not widen the blast radius of a change that narrows a live ClusterRole.

Per the trivial-follow-up rule this rolls up under backend#950 rather than earning its own card — I've noted it there so it isn't lost, and it can ride along with the next resource-monitor chart change.

Happy to just do it here instead if you'd rather have it consistent in one go — it's a one-line edit plus an assert.

shujaatTracebloc (approved):

LukasWodka:
/fr-pass

PR #745, opened 2026-08-18
feat(2036): add-to-kanban authenticates as the App

LukasWodka:
Refs backend#2036 — the last board writer to move off `PROJECTS_KANBAN_TOKEN`.

`add-to-kanban.yml` is a per-repo **copy**, not a reusable, so this is one PR per repo. The content is byte-identical across every repo that carries it, and the guard compares it that way.

## The change

Mint a `tracebloc-release-train` installation token and use it instead of the PAT.

`owner:` makes the token **org-scoped** — a repo-scoped one cannot write the org project at all.

**No fallback to the PAT**, consistent with the rest of this migration: a fallback would let a broken App path keep working silently, which is the failure class backend#1680 exists to remove.

## Dependabot

This workflow also fires on Dependabot PRs, and GitHub gates those on a **separate secret scope**. Both app secrets are set in the Dependabot scope as well as Actions — without that, Dependabot PRs would stop reaching the board with `Input required and not supplied: github-token`. That is the exact failure `PROJECTS_KANBAN_TOKEN` already had to be dual-scoped to avoid.

## Ordering

`caller-drift` compares each repo's copy against `.github`'s as the source of truth, so a fleet-wide change to a copy has an unavoidable drift window. Every other repo merges **before** `.github`, which keeps the window on `develop` rather than on `main`. `.github` is the last PR in the sweep.

---

> [!NOTE]
> **Medium Risk**
> Changes CI credentials for org kanban writes; misconfigured App secrets or Dependabot-scoped secrets would stop issues/PRs from landing on the board until fixed.
> 
> **Overview**
> **Kanban automation** now authenticates to the org project with a **minted `tracebloc-release-train` GitHub App installation token** instead of the human **`PROJECTS_KANBAN_TOKEN` PAT**, completing the board-writer migration for this per-repo workflow copy (backend#2036).
> 
> A new step runs **`actions/create-github-app-token`** with **`owner: ${{ github.repository_owner }}`** so the token is **org-scoped** (required to write the org kanban). **`add-to-project`** receives **`steps.app-token.outputs.token`**. There is **no PAT fallback** so a broken App path fails visibly rather than silently continuing on the old token.
> 
> Inline comments document that **Dependabot-triggered runs** need the same app secrets in GitHub’s **Dependabot secret scope**, not only Actions—otherwise those PRs would fail with missing **`github-token`** input.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ed938c44647aa59fd8bc185e97e32e8a30e4187. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/add-to-kanban.yml
+++ b/.github/workflows/add-to-kanban.yml
@@ -10,7 +10,24 @@ jobs:
   add-to-project:
     runs-on: ubuntu-latest
     steps:
+      # Board writes authenticate as the tracebloc-release-train App (backend#2036),
+      # not a human's PAT. `owner:` yields an ORG-scoped installation token; a
+      # repo-scoped one cannot write the org project. No fallback to the PAT: a
+      # fallback would let a broken App path keep working silently.
+      #
+      # This workflow also fires on DEPENDABOT PRs, which GitHub gates on a separate
+      # secret scope -- both app secrets are set there too, or Dependabot PRs would
+      # stop reaching the board with `Input required and not supplied`.
+      - name: Mint an installation token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.RELEASE_TRAIN_APP_ID }}
+          private-key: ${{ secrets.RELEASE_TRAIN_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
+
       - uses: actions/add-to-project@5afcf98fcd03f1c2f92c3c83f58ae24323cc57fd # v2.0.0
         with:
           project-url: https://github.com/orgs/tracebloc/projects/2
-          github-token: ${{ secrets.PROJECTS_KANBAN_TOKEN }}
+          github-token: ${{ steps.app-token.outputs.token }}
+


shujaatTracebloc (approved):
Reviewed for correctness. No findings — approving.

**Verified against the live run on this head** ([run 32109990233](https://github.com/tracebloc/client/actions/runs/32109990233/job/95627121191), triggered by this PR's own `opened` event, so the new workflow file is what executed):

- `Input 'repositories' is not set. Creating token for all repositories owned by tracebloc.` — the `owner:` input does yield the org-scoped token the org ProjectV2 write needs.
- `Creating project item` → success, then `Token revoked` in post-cleanup. The App path works end to end; this is evidence, not assertion.

Other checks:

- **Action pin** — `bcd2ba49218906704ab6c1aa796996da409d3eb1` is exactly `actions/create-github-app-token` v3.2.0 (confirmed against the upstream tag object). `add-to-project` pin is untouched.
- **Secret names** — `RELEASE_TRAIN_APP_ID` / `RELEASE_TRAIN_APP_PRIVATE_KEY` match the four already-migrated board writers in `.github` (`kanban-columns`, `kanban-archive`, `kanban-reconcile`, `customer-priority-bump`), and the successful mint above proves they resolve in the Actions scope from this repo.
- **Step wiring** — `id: app-token` ↔ `steps.app-token.outputs.token`, correct; the mint precedes the consumer; no `permissions:` change needed since neither step uses `GITHUB_TOKEN` (the run shows it stayed read-only).
- **Fork PRs** — `client` is public and this fires on `pull_request`, so a fork PR has no secrets. Not a regression: previously `add-to-project` died on an empty `github-token`, now the mint step dies on an empty `app-id`. Same visible failure, one step earlier.
- **Dependabot** — `.github/dependabot.yml` is live here (security updates, `open-pull-requests-limit: 0`), so the separate Dependabot secret scope the comment calls out is a real requirement and not theoretical. I can't read org secret scopes to confirm both are set there; that's org config rather than anything in this diff.
- **Fleet consistency** — byte-compared this copy against `cli`, `docs`, `data-ingestors`, `model-zoo`, `start-training`, `backend`, `release-train`, `e2e-test-agent`. All nine sweep PRs are identical, trailing newline included, so the copies stay byte-identical to each other; the only open item is that `.github`'s own copy match when it lands last, which the PR body already sequences for.
- **No local guard is affected** — this repo's `Source-of-truth drift` check covers Helm/`CLIENT_ENV`, not workflow copies; workflow-copy drift is the weekly `caller-drift.yml` audit in `.github`, and that one deliberately keeps the PAT (documented in-file), so nothing here contradicts it.

One non-blocking observation, deliberately **not** something to change in this PR: the run logs `Input 'app-id' has been deprecated with message: Use 'client-id' instead.` Switching it here would break the byte-identical constraint this family depends on — if it's worth doing, it's a fleet-wide sweep of its own, not a per-repo edit.

LukasWodka:
/fr-pass

PR #746, opened 2026-08-18
release-train: staging -> main

LukasWodka:
Automated promotion by the release train (RFC-0008 D14). Head is the train-managed `release-train/to-main` branch (a mirror of `staging`), so it never collides with a human PR. Merged only when the fr-gate is green.

---

> [!NOTE]
> **Medium Risk**
> Changes default training-job resource wiring and cluster-wide RBAC for the resource-monitor; installer fail-closed kubecontext checks can block installs that previously proceeded against the wrong cluster.
> 
> **Overview**
> **Release train promotion (1.9.45)** bundles chart, security, and installer changes from staging.
> 
> **Helm / jobs-manager (backend#664):** When neither `env.RESOURCE_REQUESTS` nor `env.RESOURCE_LIMITS` is set, the chart **stops injecting** those env vars on jobs-manager (api + pods-monitor). jobs-manager can then size spawned training jobs from **node allocatable** (75% with `cpu=2` / `memory=8Gi` floors); explicit operator values still render the paired vars as before. Values schema defaults and docs were updated accordingly.
> 
> **Security (backend#950):** Resource-monitor **ClusterRole** drops unused `pods/log` and `namespaces` — cluster-wide log read was an exfiltration path the DaemonSet image never uses. Tests pin the allowed resource list.
> 
> **Installers:** Bash and PowerShell k3d flows now **fail closed** if `k3d kubeconfig merge` fails or `kubectl` current context is not `k3d-<cluster>` (client#732). Recreate guidance is centralized to **`tracebloc delete --keep-data` before `k3d cluster delete`** so dashboard secure environments are not stranded (backend#2077). Preflight improves **FUSE / network filesystem** detection (opaque `fuseblk` warns instead of “local”; cloud FUSE twins). Input sanitization adds **SS3 escape** handling (cli#516). Windows bootstrap uses **amd64 cosign** with a **runs-before-verify** check (client#734). `Invoke-BoundedProcess` swallows broken-pipe on stdin write.
> 
> **Docs:** INSTALL (MySQL local disk / FUSE preflight), SECURITY (resource limit sourcing).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b78f6955b224d5c1a420e4d4fb5c55a694f3608. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LukasWodka:
bugbot run

scripts/install-k8s.ps1
```diff
@@ -1814,7 +1814,25 @@ function Invoke-BoundedProcess {
...
+  #
+  # Swallowing is right here, and it is not a fail-open: a child that closed stdin has
+  # already decided something, and its exit code and output are read below and
+  # returned unchanged. The error we would raise is about OUR pipe, not about the
+  # command -- reporting it would replace the child's real verdict with plumbing.
+  #
+  # Found on 2026-08-16: it took down `Pester (ubuntu-latest)` on client's `main` tip
+  # right after a prod promotion, on a test whose assertion never ran.
+  if ($Stdin) {
+    try { $proc.StandardInput.Write($Stdin); $proc.StandardInput.Close() } catch { }
```
saadqbal: nit: `Close()` is inside the try, so a `Write()` that throws part-way leaves the child's stdin open and it never sees EOF. For the `docker exec -i $node sh` call site that turns into waiting out the full 60s and returning 124 instead of the child's own verdict. `finally { try { $proc.StandardInput.Close() } catch {} }` keeps the swallow and still closes.

saadqbal (dismissed):
Clean. Checked the two claims that actually carry risk rather than the prose around them: the resource-monitor image (Node-deploy/resource_monitor.py + resource_utils.py + constants.py + proxy_config.py) never calls a pod-log or namespace API, so the ClusterRole trim is exact; and jobs_manager.py's DEFAULT_JOB_RESOURCES is still cpu=2,memory=8Gi while both installers keep writing RESOURCE_LIMITS/REQUESTS explicitly, so dropping the chart injection is a no-op for installer installs and safe on pre-#664 images either way. manifest.sha256 matches all 18 listed files at this head. Everything green except gate/gate, which is the FR gate waiting on kanban status, not code.

One doc nuance for whenever that file is next touched, not for this promotion: values.yaml and SECURITY.md 8.8 present node-allocatable sizing as the unset default, but jobs-manager's `nodes: list` only exists in the clusterScope: true branch of templates/rbac.yaml. With clusterScope: false, list_node() 403s and you silently get the literal instead. Identical to today's behaviour, so nothing breaks here.

LukasWodka:
bugbot run

LukasWodka:
bugbot run

LukasWodka:
Promoted with 1 open Medium/Low Bugbot finding(s), per the prod-hop policy in release-train's README (High stops the line; Medium/Low are recorded and ship):

GNU xargs flag breaks macOS lint

This is a third look at twice-reviewed code -- it passed per-feature review and functional review on `staging`. Fix forward on `develop` if any is real.

**What the train did with each:**

- **GNU xargs flag breaks macOS lint** - filed as https://github.com/tracebloc/backend/issues/2241, thread resolved by the train.

LukasWodka:
bugbot run

LukasWodka:
Promoted with 1 open Medium/Low Bugbot finding(s), per the prod-hop policy in release-train's README (High stops the line; Medium/Low are recorded and ship):

Stdin close skipped after write error

This is a third look at twice-reviewed code -- it passed per-feature review and functional review on `staging`. Fix forward on `develop` if any is real.

**What the train did with each:**

- **Stdin close skipped after write error** - filed as https://github.com/tracebloc/backend/issues/2246, thread resolved by the train.

LukasWodka:
bugbot run

saadqbal (approved):

PR #747, opened 2026-08-18
feat(telemetry): one outcome event per install, with nowhere to put a path (backend#1907)

LukasWodka:
Parent epic: tracebloc/backend#1872 · Ticket: **tracebloc/backend#1907** · Contract: [the telemetry contract](https://github.com/tracebloc/rfcs/blob/main/specs/backend-1872-telemetry-contract.md) · Sibling: tracebloc/cli#527

The installer is the highest-variance, least-observed step in the product: it
runs on machines we have never seen, under package managers, proxies and shells
we do not control, and it reports to nobody. Each of the backend#736 failures
was invisible until a customer happened to mention it.

`scripts/lib/telemetry.sh` emits **one contract-shaped event per install**, from
`install_cleanup` — the EXIT trap — so it fires on every path including the
interrupted one and the one that dies under `errexit`. That is where §6.5's
"terminal event on every path" is actually honoured; anywhere else and the
failure rate is uncomputable.

## What is emitted

`install.run.succeeded` / `.failed` / `.cancelled`, with:

| layer | attribute | notes |
|---|---|---|
| record | `tracebloc.install.phase` | `preflight`…`connect`, from the closed a–f set |
| record | `tracebloc.install.phase_<name>_ms` | six fixed keys, one per phase |
| record | `tracebloc.install.duration_ms` / `.exit_code` | ints |
| record | `tracebloc.install.client_state` | summary.sh's closed diagnosis set |
| record | `tracebloc.install.cli_on_path` | **the #736 PATH case, as a 1/0** |
| record | `error.type` | closed vocabulary, failures only (§8.4) |
| record | `tracebloc.install.source` / `.source_line` | basename + line, never the path |
| resource | `service.name` / `tracebloc.component` | `installer` / `install` (§10.1) |
| resource | `service.version`, `os.type`, `host.arch`, `deployment.environment`, `service.instance.id` | |

Ctrl-C (130/143) is `cancelled` and carries no `error.type`.

## "No arguments, no paths, no data" is a shape, not a rule

Every value goes through one writer, `_telemetry_attr`, which admits a string
only if it matches `^[A-Za-z0-9._-]{1,64}$` and an integer only if it is one. A
filesystem path contains `/`. A proxy credential contains `:` and `@`. A token
is longer than 64 characters. A person's name contains a space. **Values that
fail are dropped, never trimmed or escaped** — a value that had to be repaired
to be safe is a value we did not understand, and shipping our guess about it is
how a redactor leaks.

The phase and the client state are *additionally* checked against their closed
sets at the render boundary. That line exists because of a test failure: a
canary assigned straight to `TB_TELEMETRY_PHASE` reached the record, and it was
shaped exactly like a legal value, so the token regex waved it through.
`telemetry_phase_begin` is the only writer and already closes the set — but
"the only writer is careful" is a property that stops being true silently.

`TB_ERR_CMD` is deliberately never emitted (unexpanded command text is still
free text), and `service.instance.id` is a fresh per-run value rather than the
hostname — field hostnames here are overwhelmingly `<firstname>-macbook`, which
§7.3 forbids outright.

## The vocabularies are derived, and a required check proves it

`scripts/tests/telemetry-vocabulary-agreement.sh` holds **no copy of any
vocabulary**. It parses:

1. `install-k8s.sh`'s `step_header` calls → the phase letters
2. `summary.sh`'s `CLIENT_STATE` writers → the client states
3. `gen-manifest.sh`'s `FILES` array → the source basenames
4. `install.sh`'s immutable-release-tag regex → the `service.version` shape
5. `common.sh`'s `print_help` → the documented opt-out variables

and for the error classes — which have no second declaration to parse — it
**exercises `telemetry_error_class` over the full cross-product** of its two
closed input sets, checking both that every answer is registered *and* that
every registered class is reachable. Comparing a declaration to itself would be
self-consistent and therefore blind.

It runs in `drift-checks.yaml`'s **`Source-of-truth drift`** job, which is a
required check — a guard in a job nobody must wait for is advice, not a gate.

**It found something on its first run:** `summary.sh`'s own `CLIENT_STATE`
docstring listed five states while `_diagnose_not_ready` produces six
(`image_pull_ca`, added by #424, never made it into the comment). Fixed here.

## Two real bugs the tests caught — both fatal, both invisible to unit tests

Same shape twice: **`grep -q` closes the pipe at its first hit, so under the
installer's `set -o pipefail` a producer takes SIGPIPE and the pipeline returns
141 — on a *successful* match.** This is the backend#1778 defect, already
documented in `summary.sh`'s own comments.

1. `printf '%s' "$key" | grep -qE …` in the shape checks.
2. `tr -dc 'a-f0-9' < /dev/urandom | head -c 16` for the instance id — this one
   **at source time**, so it killed the entire installer under `set -e` before
   it printed a line.

Every unit-level test passed throughout both. The only thing that went red was
the test that drives `install_cleanup` for real, under the installer's own shell
options. That test stays, and a second one now exercises the whole surface under
`set -euo pipefail` explicitly.

## Mutation proofs

Every guard broken, watched go red, restored — **and every anchor asserted to
have applied**, because an inert mutation and real coverage are identical in a
log.

| mutation | reddened |
|---|---|
| the phase is trusted at the render boundary | 2 tests |
| `CLIENT_STATE` passed through unchecked | 2 tests |
| `_telemetry_attr` sanitises instead of dropping | 1 test |
| the source attribute emits the whole location | 1 test |
| the version shape loosens to the generic token | 1 test |
| the `urandom \| head` SIGPIPE pipeline restored | 2 tests |
| the opt-out is ignored | 1 test |
| the emit-once latch removed | 1 test |
| cancel folded into failed | 3 tests |
| an unknown `CLIENT_ENV` exported under a guess | 1 test |
| the spool made unbounded | 1 test |
| `step_header` stops driving the phase clock | 2 tests |
| a 7th install step added with no phase name | agreement guard |
| `summary.sh` gains an unknown client state | agreement guard |
| a script dropped from the source vocabulary | agreement guard |
| the classifier returns an unregistered answer | agreement guard |
| a class registered that nothing produces | agreement guard |
| the version regex drifts from `install.sh` | agreement guard |
| the phase parse made inert | **exit 2, fail closed** |
| opt-out doc/code drift, both directions | agreement guard |
| the bootstrap fixture's FILES parse made inert | **setup fails closed** |

Two mutations were rerun after strengthening a test that had stayed green:
`CLIENT_STATE` and the phase were being tested with a value containing `/`,
which `_telemetry_attr`'s shape refuses regardless — so the test proved the
shape guard and said nothing about the vocabulary guard. It now uses
`degraded` / `verifying`, which are shape-legal and exactly what a future edit
to `summary.sh` would produce.

One mutation came back **inert** and is reported as such rather than counted:
putting `_telemetry_attr`'s key check back through a pipe reddens nothing,
because for a short key `printf` completes before `grep -q` exits and no SIGPIPE
occurs. The here-string is still the correct idiom (and is what fixes the value
checks, where the input can be long), but this PR does not claim coverage it
does not have.

## What remains to be connected

**The transport, and only the transport.** The ticket said "rides the gateway
and token"; the 17 Aug decision replaced the Collector gateway with an ingest
endpoint on the backend ([rfcs#28](https://github.com/tracebloc/rfcs/pull/28)),
which is **tracebloc/backend#1905, in flight and not yet available**.

`_telemetry_deliver` therefore writes the install log and a **bounded, 0600
local spool** at `$HOST_DATA_DIR/telemetry/pending.jsonl` (last 50 events, in a
0700 directory) — which tracebloc/backend#1906's forwarder can drain — and posts
nothing. That one function is the whole change when #1905 lands.

The ticket body has been corrected so it no longer says "blocked by the gateway".

## Would the three named field failures now be visible?

| #736 case | visible? |
|---|---|
| CLI on `~/.local/bin`, PATH advice only printed | **Yes.** `tracebloc.install.cli_on_path=0` rides every event. `install-cli.sh` has always computed this — it asks whether a *fresh login shell* resolves `tracebloc` — and only ever printed advice about it. It becomes a rate. |
| `apt-get` hung on the dpkg lock | **Yes.** `tracebloc.install.phase_prerequisites_ms` rides the **success** event too — that failure produces no non-zero exit at all, it just takes twenty minutes. It also rides the **cancel and failure** events, for the phase that was *still running* when the process ended, which is the likelier real shape: stuck twenty minutes and then killed. (That second half was broken in the first two commits and is Bugbot finding 3 — see the comments; the phase durations now sum exactly to `duration_ms`, which is asserted.) |
| `cluster info` on the wrong kubeconfig context | **Not here — that is the CLI's,** and it is covered in tracebloc/cli#527 (with an honest limit stated there: a run that succeeds against a wrong-but-valid context still looks like a success). |

Two of the three are this PR's; the third is the sibling's.

## Also here

* `install-bootstrap.bats` held a hand-written **second copy** of `install.sh`'s
  `FILES` array, in two places — so adding one lib turned ten unrelated
  supply-chain tests red. It now derives the list from the array (the same parse
  `gen-manifest.sh` uses) and fails closed on an inert parse. That is a landmine
  removed, not a workaround for this PR.
* `scripts/manifest.sha256` regenerated; `telemetry.sh` added to both `FILES`
  arrays, which `gen-manifest.sh` cross-checks.
* `--help` gains a "Usage reporting" section; the copy golden is regenerated.

## Review history

Bugbot found **5 issues across 4 rounds**: 4 fixed, 1 demoted with measured
evidence (the mechanism does not exist on bash 3.2 or 5.x). In order of what they
would have cost:

1. **the spool's `mkdir` disarmed the NFS guard** — `_telemetry_deliver` created
   `HOST_DATA_DIR` from the EXIT trap on the very path where
   `early_data_dir_guard` had just refused it, and that guard skips an existing
   directory by design. Run 1 refused and created it; run 2 installed MySQL onto
   NFS. Customer database corruption, reintroducing client#432.
2. **`--help` emitted `install.run.succeeded`** — a free success in the denominator
   of the failure rate this ticket exists to produce.
3. **the still-running phase had no duration** — the dpkg-lock case invisible in
   its likeliest form, and `phase_connect_ms` absent on every success. This
   undercut the DoD claim in the table above.
4. **pre-log failures were reported nowhere** — fallout from fixing (1): `log` is a
   no-op until `setup_log_file`, which runs *after* the guards, so the NFS refusal
   itself became permanently invisible. Events now fall back to a `mktemp`'d file
   in `$TMPDIR`. **My own NFS test had been masking this** by setting
   `LOG_FILE=/dev/null`.
5. **source lookup aborts the whole event** — demoted; a command substitution in an
   argument position does not propagate its status, measured on bash 3.2 and 5.x.

**All four real ones were invisible to my own tests, for the same reason:** they
were about *which code paths reach the emitter*, not about what the emitter does
with its inputs — and one of my tests was actively masking a bug. No unit test of `telemetry_render_event` can see any of them.
Every replacement test now drives the real `install-k8s.sh`, `install_cleanup`,
`_assess_handoff`, or `early_data_dir_guard`. Full reproductions and mutation
tables are in the PR comments.

Two of my own mutations came back **inert** and were fixed by moving the tests end
to end rather than by counting them; a third is reported as inert rather than
counted. Details in the comments.

## Test plan

`make check` green (lint, drift incl. the new guard, helm-lint, helm-vocab);
`bats scripts/tests/*.bats` green (31 new tests); `shellcheck -S
warning -x` clean on every new file; `bash -n` clean; `gen-manifest.sh --check`
current; the mutations above plus 10 more from the review rounds (see comments).

---

> [!NOTE]
> **Medium Risk**
> Changes installer EXIT-trap behavior and data-dir/spool logic on failure paths (including NFS refusal); mistakes could skew failure metrics or recreate dirs the guard rejects, though transport is local-only and heavily tested.
> 
> **Overview**
> Adds **`scripts/lib/telemetry.sh`** and wires it through the bash installer so **`install_cleanup`** emits **one contract-shaped outcome** (`install.run.succeeded` / `.failed` / `.cancelled` / `.skipped`) per committed run, with phase timings, exit code, closed vocabularies for errors/client state, and **`cli_on_path`** — values pass a strict shape gate so paths and credentials cannot be recorded. Delivery is **log + bounded `pending.jsonl` spool** only until backend ingest exists; opt-out via **`TRACEBLOC_NO_TELEMETRY`** / **`DO_NOT_TRACK`**.
> 
> **Integration:** `telemetry.sh` is fetched and manifest-signed; **`step_header`** drives the phase clock; **`telemetry_run_started`** skips `--help`/terminal exits; **`telemetry_run_skipped`** / **`telemetry_rerun_handoff`** distinguish healthy handoffs from real failures vs stray exit `2`; GPU reboot **`exit 2`** declares the handoff. **`--help`** documents usage reporting.
> 
> **Guards & tests:** **`telemetry-vocabulary-agreement.sh`** (required drift job, `make drift`, shellcheck) derives phases, client states, sources, version regex, and event names from producers; **`telemetry.bats`** exercises trap paths, NFS/pre-log/bootstrap TMPDIR fallbacks, and privacy. **`install-bootstrap.bats`** parses **`install.sh`’s `FILES`** instead of duplicating the list; **`summary.sh`** comment aligns with **`image_pull_ca`**; **`.gitleaks.toml`** allowlists telemetry canary fixtures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9a8d62cb553dfee62c401122c7be4415c0e1f22c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LukasWodka:
bugbot run

LukasWodka:
Sibling PR for the CLI half of backend#1907: tracebloc/cli#527. Independent — neither depends on the other, and the transport for both is tracebloc/backend#1905.

LukasWodka:
bugbot run

LukasWodka:
## ⚠️ `quality / gitleaks` is red, and it needs one command from a human

**What it found:** the `curl-auth-user` rule, on a test fixture in
`scripts/tests/telemetry.bats` that spelled out `curl -u $CANARY:hunter2 …`. It
was a canary — the whole point of that test is to prove `TB_ERR_CMD` never
reaches the record — but the rule is right: a source file containing that
spelling is a finding whatever the surrounding test claims, and a reviewer
scanning a diff cannot tell a canary from the real thing at a glance.

**Already fixed in the code** (f495186): the fixture now carries a path instead
of a credential, and the credential half of the same assertion is covered by
`HTTPS_PROXY`, which encodes `user:pass` in a proxy URL — what a hospital
network actually configures, and what gitleaks does not flag.

**Why the check is still red:** gitleaks scans the *commit range*
(`3a618e2..f495186`, "2 commits scanned"), so the line is still found in the
first commit's diff even though it no longer exists in the tree.

**The fix, and why I have not applied it.** The remedy is to squash the branch so
the blob never appears in any commit:

```bash
git rebase -i --autosquash origin/develop   # or: git reset --soft origin/develop && git commit
git push --force-with-lease
```

I do not force-push without being asked to — that is an explicit rule for AI
sessions in this repo, and it applies to my own branches too.

**What I deliberately did NOT do:** add this to `.gitleaks-baseline.json`. That
file is for findings that predate the gate; putting a brand-new one in it to
make my own PR green would be weakening a security check to avoid a rebase.

Everything else on this PR is green.

LukasWodka:
bugbot run

LukasWodka:
## Bugbot round 1 — one fixed, one demoted with evidence

### ✅ "Non-install exits report success" (High) — real, fixed

**Reproduced exactly as described:**

```
$ HOST_DATA_DIR=$T bash scripts/install-k8s.sh --help
$ cat $T/telemetry/pending.jsonl
{"resource":{…},"attributes":{"event.name":"install.run.succeeded",
 "tracebloc.install.phase":"bootstrap","tracebloc.install.exit_code":0,…}}
```

This is the worst possible bug in this PR and the severity is right: `--help` is
the command people run *most* while a real install is broken, and a free success
in the denominator moves the failure **rate** — the one number the whole ticket
exists to produce — in the direction that hides the problem.

**Fix:** a run-started latch, not a phase test. `main()` calls
`telemetry_run_started` once the terminal commands (`--help`, `--diagnose`,
`prepare-host`) have had their chance to dispatch; `telemetry_emit_outcome`
returns early without it. A latch rather than "phase != bootstrap" because a
genuine failure *in* the bootstrap phase — the leftover-data guard,
`validate_config` — **is** an install attempt and must still be reported.

**Second, related thing found while fixing it:** the assess handoff exits 0
having run no step, so counting it as `install.run.succeeded` would make the
success count grow with re-runs on machines nothing happened to. It now emits
`install.run.skipped` — a registered outcome verb (contract §6.4), so no new
vocabulary — which also makes "how often do people re-run an installer that was
already done" a question we can ask.

**Not covered, deliberately:** `prepare-host`. It is a different command with its
own registry component (§10.1 gives `installer` the components
install/preflight/upgrade), and reporting it as `tracebloc.component=install`
would be mislabelling it rather than measuring it.

### ⚠️ "Source lookup aborts whole event" (Medium) — mechanism does not exist

The predicted mechanism is that the unprotected command substitutions feeding
`tracebloc.install.source` abort `telemetry_render_event` under `set -e` and take
the whole outcome with them. **A command substitution in an *argument* position
does not propagate its status to the enclosing command** — only an assignment
(`x=$(false)`) does. Constructed and measured on **bash 3.2.57** (the system
bash on macOS, the version that actually matters here) and on 5.x:

```
$ TB_ERR_LOC='/home/someone/evil.sh:9'   # deliberately not one of our scripts
$ telemetry_emit_outcome 1               # under set -euo pipefail
REACHED_END   rc=0   spooled: 1
```

The event is rendered, the process survives, one line is spooled. Reported as
unreachable rather than fixed — a fix for a path that cannot be taken costs more
than the finding.

**But the invariant is worth having**, so it is now pinned: *an unrecognised
source location drops the field, never the event* — asserted through a subshell
under `set -euo pipefail`, checking the event still renders, still carries
`error.type`, and carries neither `tracebloc.install.source` nor the rejected
filename.

### Mutation proofs for the new code

Every new guard broken, watched go red, restored, **anchor asserted to have
applied**:

| mutation | reddened |
|---|---|
| the run-started latch removed | 2 tests |
| the latch set at source time (always on) | 1 test |
| `main` stops setting the latch (**feature dead**) | 1 test |
| the skipped handoff reported as succeeded | 1 test |
| `assess.sh` stops marking the handoff skipped | 1 test |
| a bootstrap failure classified as something else | 1 test |

**Two of those came back inert on the first pass** — "`main` stops setting the
latch" and "`assess.sh` stops marking the handoff" — because the tests set the
flags themselves and so could not see the *wiring* disappear. That is the same
class as the bug you found: a unit test of the emitter cannot observe which exits
reach the trap. Both are now driven end to end through the real entrypoint (a
`validate_config` rejection, which lands after the latch and before step a) and
through the real `_assess_handoff` with `tracebloc` mocked. Both mutations redden
now.

26 tests in `telemetry.bats`, `make check` green.

LukasWodka:
bugbot run

LukasWodka:
## Bugbot round 2 — the third finding was real too, and it hit the DoD claim

I missed this one in my first reply: it came from Bugbot's **earlier** review pass
and my query only fetched the later one. It is the most consequential of the three.

### ✅ "Final phase duration never recorded" (Medium) — real, fixed

`telemetry_phase_begin` only closes a phase when the *next* one starts, and
nothing closed the active phase before render. Reproduced — a run cancelled after
22 minutes stuck in `prerequisites`:

```
"tracebloc.install.phase":"prerequisites"
"tracebloc.install.duration_ms":1330000
"tracebloc.install.phase_preflight_ms":0
        ← no prerequisites duration at all
```

**This undercut the DoD claim in the PR body.** I wrote that per-phase durations
make the dpkg-lock case visible. They did — but only in the version where the
install goes on to *succeed*. In its likeliest real form the user is stuck twenty
minutes in `prerequisites` and then kills it or gives up, never reaching step c,
and the number was silently absent. Likewise `phase_connect_ms` was missing on
**every** successful install: the readiness wait, up to `READY_TIMEOUT` (600s),
the single longest phase in the run.

And as you noted, the remainder trick could not recover it, because `bootstrap`
had no key at all — the loop iterated the letter map, and `bootstrap` has no step
letter. So download + verify + leftover-guard + assess time was an unnamed
remainder.

**My own test passed throughout.** "A slow phase is visible on a run that
SUCCEEDED" only ever measured a phase that a *later* `step_header` had closed —
the exact shape of a test that proves the easy half of its own claim.

**Fix:** the live delta is added at **read** time, not by a "close the phase" call
in the emit path, so `telemetry_render_event` stays idempotent (the tests call it
repeatedly; a render that mutated the accumulators would report different numbers
each time). The loop now iterates `_telemetry_phase_names`, so `bootstrap` and
`unknown` get keys too. The clock is read **once per event**, so the per-phase
numbers and the total are exactly consistent.

**New invariant, asserted:** *the phase durations sum EXACTLY to `duration_ms`* —
every millisecond of the run attributed to exactly one phase, with the individual
attributions also pinned so a compensating pair of errors cannot pass, and a
`counted == 4` anchor so an inert loop cannot.

That test needed a **fake clock**, and the reason is worth recording: the obvious
fixture is wrong. Winding `_TB_TELEMETRY_PHASE_STARTED_MS` backwards *after* the
`step_header`s have already attributed that time invents milliseconds that never
elapsed. The first version of the test failed for exactly that reason — sum
1200000 against a total of 900000 — and the fixture was the bug, not the code.

### Mutation proofs

| mutation | reddened |
|---|---|
| the active phase's live delta dropped (the original bug) | 3 tests |
| the loop reverts to the letter map (bootstrap unattributed) | 2 tests |
| the live delta counted for every phase (double counting) | 1 test |
| the clock read per attribute instead of once | **inert — see below** |

The fourth is **reported as inert rather than counted**: `_telemetry_now_ms` has
second resolution (BSD `date` has no `%N`, and half the install base is macOS),
so two reads inside the same second are identical and the guard only matters
across a second boundary. It is kept because it is free, not because a test
proves it.

---

## Running total across both rounds

**3 findings, 2 fixed, 1 demoted with measured evidence.** Two of my own
mutations came back inert along the way and were fixed by moving the tests end to
end rather than by counting them.

29 tests in `telemetry.bats` (up from 20), `make check` green, the required
`Source-of-truth drift` guard green.

`quality / gitleaks` remains red for the history reason documented above and
needs one force-push from a human.

LukasWodka:
bugbot run

LukasWodka:
## Bugbot round 3 — the most serious one yet, fixed

### ✅ "Spool mkdir undoes NFS guard" (High) — real, and worse than the first two

The first two findings corrupted a *metric*. This one corrupted a customer's
*database*. Reproduced exactly as described:

```
guard exit=1
HOST_DATA_DIR created on the REJECTED volume? YES
./nfs-volume/.tracebloc/telemetry
```

`_telemetry_deliver` ran from the EXIT trap and `mkdir -p`'d
`$HOST_DATA_DIR/telemetry` unconditionally — including on the path where
`early_data_dir_guard` had just refused that directory for being on a network
filesystem and called `error`. That guard skips an existing directory *on
purpose* (`[[ -d "$target" ]] && return 0` — "an EXISTING data dir has no at-risk
mkdir here", client#441), so:

```
run 1: guard refuses (dir absent) → error → EXIT trap → telemetry creates it
run 2: guard sees the dir, returns 0 → MySQL installs onto NFS
```

which is the InnoDB corruption client#432 exists to prevent, reintroduced by the
telemetry that was only supposed to watch.

**The principle I got wrong:** an observer that changes the install's own
preconditions is not an observer. Adding a write to the EXIT trap put telemetry
*upstream* of a safety gate, and nothing in the design made me look for that.

**Fix:** the spool only writes **into a data dir that already exists**. A run that
dies before then still reports through the install log — which is what a support
bundle collects, and which `_choose_log_file` has already placed somewhere safe
(falling back to `$TMPDIR`). The forwarder (#1906) picks up every later run.

**Both directions mutation-proven**, because "don't write" is trivially satisfied
by "never write":

| mutation | reddened |
|---|---|
| the `HOST_DATA_DIR` existence check removed (guard disarmed again) | the new test |
| delivery disabled outright (a "fix" that kills the feature) | 4 other tests |

The regression test carries an anchor asserting the guard **actually refused**
(non-zero exit + the "network filesystem" message), so "no directory" cannot
silently mean "the fixture never ran the guard".

---

## Final tally across three rounds

**4 findings, 3 fixed, 1 demoted with measured evidence.** In order of what they
would have cost:

1. **spool mkdir disarms the NFS guard** → customer database corruption. Fixed.
2. **`--help` emits `install.run.succeeded`** → the failure rate the whole ticket
   exists to produce, moved in the direction that hides problems. Fixed.
3. **the still-running phase had no duration** → the dpkg-lock case invisible in
   its likeliest real form; `phase_connect_ms` absent on every success. Fixed.
4. **source lookup aborts the whole event** → mechanism does not exist on bash
   3.2 or 5.x; demoted with a reproduction attempt, invariant pinned anyway.

Three of those were invisible to my own tests, and the pattern is consistent: all
three were about **which code paths reach the emitter**, not about what the
emitter does with its inputs. Unit tests of `telemetry_render_event` cannot see
any of them. Every replacement test now drives the real `install-k8s.sh`, the real
`install_cleanup`, the real `_assess_handoff`, or the real
`early_data_dir_guard`.

30 tests in `telemetry.bats` (up from 20). `make check` green, `Source-of-truth
drift` green, full `bats scripts/tests/*.bats` green locally.

`quality / gitleaks` remains red for the commit-history reason documented earlier
and needs one force-push from a human.

LukasWodka:
bugbot run

LukasWodka:
## Bugbot round 4 — right again, and it was fallout from my own round-3 fix

### ✅ "Early failures drop outcome events" (High) — real, fixed

This is the sharpest of the four, because it is a **consequence of the previous
fix**: stopping telemetry from disarming the NFS guard also made the NFS refusal
itself invisible.

Reproduced:

```
$ early_data_dir_guard        # target reads as nfs, LOG_FILE unset as on the real path
guard exit=1
any telemetry written anywhere? 0
```

`_telemetry_deliver`'s comment claimed *"the install log always gets it."* It does
not. `log` is a no-op until `setup_log_file` sets `LOG_FILE`, and `setup_log_file`
runs **after** `validate_config` and `early_data_dir_guard` — deliberately,
because client#432 refuses a network data dir *before* logging starts. So on
exactly those paths there was no log **and** (correctly, since the last commit) no
data dir, and the rendered event was dropped on the floor.

Those pre-log failures are also precisely the class the run-started latch was
built to preserve, so losing them quietly undid that work too.

**And you were right about the test.** My NFS test set `LOG_FILE=/dev/null`, which
made `log` appear to succeed and hid the whole thing. It now leaves `LOG_FILE`
unset, as the real path does, and asserts the refusal **is** reported. A second
test covers the general pre-log case.

**Fix:** when there is no data dir, spool to a `mktemp`'d file in `$TMPDIR`.
`mktemp` and not a fixed name — `/tmp` is world-writable on Linux and the
installer runs privileged steps, so a predictable path is a symlink target for an
append that may be running under `sudo`; `mktemp` creates with `O_EXCL`. This
mirrors `_choose_log_file`'s own fallback, so an early-failure run leaves one
small file beside the install log it already leaves there. #1906's forwarder
reads both locations.

The false comment is replaced with what actually holds and why — a comment that
claims something untrue is itself the defect.

| mutation | reddened |
|---|---|
| the fallback removed (early failures dropped again) | 2 tests |
| the fallback given a predictable shared path | 1 test |
| the fallback's 0600 mode dropped | 1 test |
| the data-dir existence test dropped (NFS guard disarmed again) | 2 tests |

---

## Final tally — 5 findings across 4 rounds

**4 fixed, 1 demoted with measured evidence.** Every single real one was about
**which code paths reach the emitter**, never about what the emitter does with its
inputs:

| # | finding | would have cost |
|---|---|---|
| 1 | spool `mkdir` disarms the NFS guard | customer database corruption (client#432 reintroduced) |
| 2 | `--help` emits `install.run.succeeded` | the failure rate, moved the way that hides problems |
| 3 | the still-running phase has no duration | the dpkg-lock case, invisible in its likeliest form |
| 4 | pre-log failures reported nowhere | the NFS refusal, permanently invisible |
| 5 | source lookup aborts the event | *demoted — mechanism absent on bash 3.2 and 5.x* |

No unit test of `telemetry_render_event` can see any of 1–4. Every replacement
test now drives the real `install-k8s.sh`, `install_cleanup`, `_assess_handoff`,
or `early_data_dir_guard` — and one of my own tests had been actively masking a
bug, which is the thing worth remembering out of all of this.

31 tests in `telemetry.bats` (up from 20). `make check` green, `Source-of-truth
drift` green.

`quality / gitleaks` remains red for the commit-history reason documented earlier
and needs one force-push from a human.

LukasWodka:
bugbot run

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,616 @@
...
+#
+# Everything before this function is finished. This function is the change.
+# _telemetry_fallback_spool — where an event goes when there is no data dir yet.
+#
+# mktemp, never a fixed name: /tmp is world-writable on Linux, so a predictable
+# path is a symlink target for an append that may be running under sudo. mktemp
+# creates with O_EXCL. This mirrors _choose_log_file's own fallback exactly, so an
+# early-failure run leaves one small file beside the install log it already
+# leaves there — not a new class of litter.
+_telemetry_fallback_spool() {
```
saadqbal: This puts the fallback inside the bootstrap's own scratch dir, so on `curl | bash` it's deleted before anyone can read it.

`install.sh:238` does `TMPDIR="$(mktemp -d)"`, `:239` sets `trap 'rm -rf "$TMPDIR"' EXIT`, and `:571` runs `bash "$TMPDIR/install-k8s.sh"`. When TMPDIR is already in the environment — always on macOS, and plenty of Linux sessions — that assignment keeps the export attribute, so install-k8s.sh inherits the bootstrap's temp dir and `mktemp "${TMPDIR:-/tmp}/…"` lands inside it. The bootstrap then wipes it the moment install-k8s.sh returns:

```
$ cat p.sh
TMPDIR="$(mktemp -d)"
trap 'rm -rf "$TMPDIR"; echo "cleaned $TMPDIR"' EXIT
bash -c 'mktemp "${TMPDIR:-/tmp}/tb-XXXXXX"'
$ env TMPDIR=/var/folders/fake bash p.sh
/var/folders/…/tmp.lhlMIBO6ee/tb-X8Gzni
cleaned /var/folders/…/tmp.lhlMIBO6ee
$ env -u TMPDIR bash p.sh
/tmp/tb-qRQedU                      # survives
```

So the NFS refusal and every other pre-`setup_log_file` failure still produces no record anywhere on macOS. `_choose_log_file` has the same shape, but nobody promised to drain that one. Somewhere the bootstrap doesn't own would fix it — `$HOME`, or install.sh exporting its original TMPDIR under another name.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,616 @@
...
+  local key="$1" value="$2" kind="${3:-str}"
+  grep -qE "$TB_TELEMETRY_KEY_RE" <<<"$key" || return 0
+  [ -n "$value" ] || return 0
+  case "$kind" in
+    int)
+      grep -qE "$TB_TELEMETRY_INT_RE" <<<"$value" || return 0
+      _TB_TELEMETRY_BUF="${_TB_TELEMETRY_BUF:+${_TB_TELEMETRY_BUF},}\"${key}\":${value}"
+      ;;
+    *)
+      grep -qE "$TB_TELEMETRY_TOKEN_RE" <<<"$value" || return 0
```
saadqbal: `grep` matches line by line, so a value containing a newline passes the shape check and lands verbatim — the one input shape the "nowhere for a path to go" argument doesn't cover.

```
$ TB_VERSION=$'v1.9.3\n","tracebloc.install.injected":"yes'
$ telemetry_render_event 0
{"resource":{…,"service.version":"v1.9.3
","tracebloc.install.injected":"yes","deployment.environment":"prod",…
```

A forged attribute, plus one record split across two lines of a `.jsonl` spool — #1906's forwarder reads two malformed events. `_telemetry_attr` has it too: `_telemetry_attr tracebloc.install.phase $'helm\n/etc/passwd'` buffers the path.

`[[ $value =~ $RE ]]` is whole-string and fixes both here and at `_telemetry_version` (l.465). Worth noting install.sh's release-tag gate already uses `[[ =~ ]]`, so the agreement check proves the two regexes are byte-identical while they behave differently.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,616 @@
...
+  # §6.1 — three segments, a registered domain, a registered outcome verb. The
+  # name is assembled from literals only; no runtime value appears in it.
+  case "$code" in
+    0)       if [ -n "$_TB_TELEMETRY_SKIPPED" ]; then
+               event="install.run.skipped"
+             else
+               event="install.run.succeeded"
+             fi ;;
+    130|143) event="install.run.cancelled" ;;
+    *)       event="install.run.failed" ;;
```
saadqbal: exit 2 is the "complete this step and re-run" handoff, not a failure, and this files it as `install.run.failed`.

`gpu-nvidia.sh:55` exits 2 after `install_nvidia_drivers` **succeeded**, to ask for a reboot — called from `install_linux`, i.e. step b — so it renders `error.type=prerequisites_failed`. Every GPU host's first install books a prerequisites failure in the rate this ticket exists to produce. Same shape as the `--help` bug, opposite direction.

`install_cleanup` already treats 2 as its own outcome ("Re-run required"), so it reads like it wants its own verb rather than being folded into failed.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,616 @@
...
+  #
+  # So: spool INTO the data dir only when it already exists — and when it does
+  # not, into a temp file rather than nowhere. #1906's forwarder reads both.
+  if [ -n "${HOST_DATA_DIR:-}" ] && [ -d "$HOST_DATA_DIR" ]; then
+    spool="$(_telemetry_spool_path)"
+    dir="${spool%/*}"
+    mkdir -p "$dir" 2>/dev/null || return 0
+    chmod 700 "$dir" 2>/dev/null || true
+    printf '%s\n' "$json" >> "$spool" 2>/dev/null || return 0
+    chmod 600 "$spool" 2>/dev/null || true
```
saadqbal: The chmod runs before the trim replaces the file, so the 0600 doesn't survive. `tail > "${spool}.tmp"` creates the tmp file under the process umask and `mv` renames it over the spool, keeping the tmp file's mode.

`common.sh`'s `umask 077` normally covers it — but `_install_userspace_tools` (setup-linux.sh:893) and its macOS twin set `umask 022` around install_kubectl/k3d/helm and only restore it afterwards, so an install that dies in one of those emits from the EXIT trap with umask 022:

```
$ umask; telemetry_emit_outcome 0
022
spool mode: 644    dir mode: 700
```

Nothing sensitive is in there by construction, so it's defence-in-depth — but the test that pins 600 can't see it, because `load_lib` sources common.sh first. Moving the chmod after the `mv` (or a local `umask 077`) closes it.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,616 @@
...
+
+# ── Phase vocabulary ─────────────────────────────────────────────────────────
+# The letters are install-k8s.sh's own `step_header a..f` labels; the names are
+# what a dashboard shows. step_header calls telemetry_phase_begin, so this map
+# is keyed on the thing that actually runs the steps rather than on a parallel
+# list somebody has to remember to extend — and
+# scripts/tests/telemetry-vocabulary-agreement.sh parses install-k8s.sh to prove
+# the two sets are identical.
+#
+# `bootstrap` is the phase before step a: download + verify + the leftover-data
```
saadqbal: nit: download + verify aren't in this process. They happen in `install.sh`, which never sources telemetry.sh and whose EXIT trap is `rm -rf "$TMPDIR"`, not `install_cleanup` — so a fetch, manifest or cosign failure emits nothing at all, and `bootstrap` really means "install-k8s.sh before step a".

Same reason `install.sh` in `TB_TELEMETRY_SOURCES` is unreachable: `TB_ERR_LOC` is only ever set by common.sh's ERR trap, inside install-k8s.sh. Not asking for bootstrap telemetry here, just that the comment shouldn't claim the coverage.

saadqbal (changes requested):
Four findings, all about which runs reach the emitter or what gets past the shape guard — nothing wrong with the render logic itself. The `$TMPDIR` one is the significant one: it quietly undoes the round-4 fix on the primary macOS path.

Two things that don't fit on a line in the diff:

- `install.run.skipped` will almost never fire on the `curl | bash` path. `install.sh:144` `exec tracebloc`s on a healthy machine before install-k8s.sh is fetched at all, so assess.sh's gate only sees direct `./install-k8s.sh` re-runs. Fine, just much narrower than "how often do people re-run an installer that was already done".
- `quality / gitleaks` is still red. Your comment explains it and what it needs; noting it's the one thing blocking merge.

Vocabulary guard and the 31 bats tests are green here on 46a33de.

scripts/lib/telemetry.sh
LukasWodka: Reproduced, and it is worse than "the fallback is in the wrong place" — your `p.sh` is exactly it, and I ran the same shape through the real code:

```
$ bash repro.sh          # TMPDIR exported, as it always is on macOS
spooled to: /var/folders/…/T/tmp.whIP6voQPg/tracebloc-telemetry-fhZYDi
$ # after the bootstrap stand-in returns:
GONE: /var/folders/…/T/tmp.MTeFUxD3hM/tracebloc-telemetry-iBmlOh
```

So round 4 closed the hole on Linux-with-no-TMPDIR and nowhere else. Fixed in `_telemetry_fallback_dir`.

I took neither of your two suggestions exactly, for a reason worth stating: install.sh exporting its original TMPDIR under another name only helps when the bootstrap is new, and install.sh is served from a URL a user may have curl'd months ago — a fix that depends on the peer being current does not work on the machines this feature exists for. So it derives the answer instead: **is the running installer inside this directory?** That is true precisely when TMPDIR is the bootstrap's scratch dir, because install.sh unpacks install-k8s.sh + lib/ into it and runs it from there. When it is, the record goes to `$HOME` (your other suggestion) — never `$HOME/.tracebloc`, which is HOST_DATA_DIR and which telemetry must not create — and `/tmp` if there is no usable HOME.

One thing that only showed up because I re-ran the reproduction against the fix rather than reading it: the first cut compared the two paths as written and fired on nothing, because `pwd -P` resolves the installer's dir to `/private/var/folders/…` while `$TMPDIR` keeps the `/var/folders/…` spelling it was exported with. It missed every Mac — the one platform with the bug. Both sides are physical now.

New test `the pre-log record survives the BOOTSTRAP, not just the process` reproduces the bootstrap rather than describing it: it unpacks the libs into an mktemp'd dir, traps `rm -rf` on it, sources from there, and asserts the record is readable in `$HOME` afterwards and that the scratch dir really is gone (otherwise the test proves nothing). The existing pre-log test stays, with a note that it hands the fallback a safe TMPDIR and therefore says nothing about this path — that is why it was green throughout.

Mutation: revert the disqualification to a no-op → the new test reddens. Anchor asserted applied.

You are right that `_choose_log_file` has the same shape. Left alone deliberately — nothing drains it, and widening this PR into common.sh's log path is a separate change.

scripts/lib/telemetry.sh
LukasWodka: Reproduced, and it is all four checks, not two:

```
--- str:          BUF=["tracebloc.install.phase":"helm
                       /etc/passwd"]
--- int:          BUF=["tracebloc.install.duration_ms":12
                       ,"x":"y]                              # unquoted, so it forges structure
--- key:          BUF=["event.name
                       BAD KEY":"ok"]
--- source_line:  88
                  /etc/shadow
```

Every one now `[[ $v =~ $RE ]]`. The regexes are unchanged — only the operator — and the same inputs are refused. A bonus I did not expect: it also removes the external process, so the backend#1778 SIGPIPE class the here-strings were working around cannot recur here at all, and the comment explaining the here-string is now the comment explaining why there is no longer anything to work around.

**On the sting — yes, the guard was proving byte-identity and calling it behaviour.** Worth being precise about what was wrong, because the check itself was not:

- `[ "$boot_re" != "$TB_TELEMETRY_VERSION_RE" ]` is a true and useful check. It just answers "are these the same string".
- The line it printed was `ok: the service.version shape is install.sh's own release-tag regex` — a claim about what the emitter *admits*. Byte-identity does not support that claim, because the two sides were never evaluated by the same operator. Both halves of the sentence were true and the conjunction was false.
- That is #1729's class exactly, in the form that is hardest to see: not a check wired to nothing, but a check wired to a *neighbouring* property, reporting the one you wanted.

Check 5 now does both, in order: (a) byte-identity, kept because it localises a failure to "somebody edited one of them"; (b) verdict agreement over a corpus, with each side evaluated the way its own file evaluates it — `[[ =~ ]]` for install.sh, and **the real `_telemetry_version`** for telemetry.sh rather than a re-spelling of the match, so a mutation of the function has to redden it.

Two things about the corpus, both deliberate. It is written down independently of either matcher rather than derived — for a vocabulary, the input domain has to come from somewhere the matcher cannot reach, or a typo in the needle plants itself in the detector. And it **fails closed if it contains no embedded-newline input**, because without one (b) computes the same answer as (a) and would report clean forever.

Mutations, each with the anchor asserted:
- str / int / key / source_line back to `grep -qE` → the new bats test `a value with an embedded newline is refused, not split across two records` reddens on each. That test also asserts the record stays one line, and anchors on the legal spellings still being admitted.
- `_telemetry_version` back to `grep -qE` → the guard's behavioural check reddens (the byte check stays green, which is the whole point).
- Strip the newline inputs from the corpus → the guard exits 2 rather than passing. Worth flagging that my first attempt at this one came back green and looked like a vacuous guard: the regex I mutated with rewrote the *detector* next to the corpus as well, so it matched its own mutated needle. Vacuous mutation, not vacuous guard — retargeted at the array literal only.

scripts/lib/telemetry.sh
LukasWodka: Reproduced — traced the chain and rendered it:

```
step_header b → install_linux → dispatch_gpu_setup → install_nvidia_drivers → exit 2
{"event.name":"install.run.failed","tracebloc.install.phase":"prerequisites",
 "tracebloc.install.exit_code":2,…,"error.type":"prerequisites_failed"}
```

A fabricated prerequisite failure booked immediately after the prerequisite succeeded. Fixed: exit 2 now renders `install.run.cancelled` and carries no error.type.

**One correction to the scope, since you asked for evidence rather than agreement.** It is not *every* GPU host's first install — `exit 2` at gpu-nvidia.sh:55 sits inside `if [[ -n "${TRACEBLOC_SKIP_REBOOT_PROMPT:-}" ]]`. With a tty the installer prompts and continues; without the variable and without a tty it warns and continues too. So it is every **unattended** GPU host's first install. That is CI, orchestrated rollouts and anything piped — not a small population, and the defect is identical — but the comment and the commit say "unattended" rather than "every", because a rate we are fixing should not be described by a number that is itself wrong.

**On "its own verb".** I agree it wants one and did not add one: §6.4's outcome verbs are a closed list (`started succeeded failed skipped rejected retried timed_out expired cancelled completed`) and adding a row is a PR against `rfcs/specs/backend-1872-telemetry-contract.md`, which is not a call an emitter should take unilaterally — that is how the closed set stops being closed. Of the registered verbs, `cancelled` is the only terminal one that is true here: the run stopped before completing, deliberately, without an error. The verb does not say by whom, and `tracebloc.install.exit_code` already separates the two causes (2 = handoff, 130/143 = the user's Ctrl-C), which is precisely why the exit code is an attribute instead of something the name carries.

If you would rather have a dedicated verb — `deferred` is the honest one — say so and I will open the spec PR against rfcs and follow it here. I did not want to ship a name this repo had invented for itself.

**Closed vocabulary + guard**, as you asked. `TB_TELEMETRY_EVENT_NAMES` now declares the four names; `telemetry_render_event` checks its own answer against it and **drops** an unregistered one rather than filing it (same rule §3.2 applies to an unrecognised environment, same reason). The agreement guard derives the emitted set two independent ways — the literals in the case statement, and what the function actually renders over `0 1 2 42 130 143` — rather than reading the declaration twice, and checks §6.1's three-segment grammar. What it cannot check from this repo is the §6.4 half: the verb registry is in rfcs, not checked out here, and hand-copying the verbs into this file would be the fifth-declaration defect this script exists to prevent. That limit is written down in the declaration rather than left implied.

The guard also now derives its exit-2 premise from `install_cleanup`'s own branch: if common.sh stops treating 2 specially, the check fail-closes and retires itself loudly instead of guarding a rule that no longer exists.

Mutations, anchors asserted: `2|130|143` → `130|143` reddens both the new bats test and the guard's exit-2 check; deleting the event-name membership check reddens `every event name the renderer produces is one it declares`.

scripts/lib/telemetry.sh
LukasWodka: Reproduced, your numbers exactly:

```
umask: 0022
spool mode: 644
dir   mode: 700
```

Fixed — but not by moving the chmod after the `mv`, which was my first cut and which the mutation caught as unobservable. Removing that trailing chmod left the test green, because the chmod I had also added on `${spool}.tmp` already did the work. Two chmods where one is reachable is belt and braces that no test can redden, which is the thing this repo keeps finding in its own guards. So there is now **one** chmod, on `${spool}.tmp` before the `mv` — the inode that survives — which also closes the window in which the file is briefly world-readable that a post-`mv` chmod would leave open.

And yes, the test was blind, for the reason you name: `load_lib` sources common.sh, so every test in the file has been running under `umask 077` and getting the right answer for the wrong reason. New test `the spool is 0600 under the umask the installer can actually be holding` sets 022 explicitly and — because a umask that silently did not take looks exactly like a pass — **asserts `umask` reads 0022 before it measures anything**, restores the caller's umask, and checks the trim actually ran (2 lines from 3 emits), so a spool that was never rewritten cannot keep its 600 vacuously.

Mutation: drop the `chmod 600 "${spool}.tmp"` → the new test reddens with `spool is 644 under umask 022, not 600`. The old test stays green through that mutation, which is the evidence that it was measuring the umask and not the code.

Agreed it is defence in depth — nothing sensitive is in there by construction. The reason I did not just note it and move on is the one you imply: which mode the file ends up with should not depend on which line the installer happened to die on.

scripts/lib/telemetry.sh
LukasWodka: Correct on every point, and checked rather than taken:

- `install.sh`'s only mention of telemetry.sh is line 268 — the FILES list it downloads. It never sources it.
- its only EXIT trap is `trap 'rm -rf "$TMPDIR"' EXIT` (:239). `install_cleanup` is nowhere in it.
- `TB_ERR_LOC` has exactly one writer in the tree, `common.sh:988` (`_record_err`), and common.sh is only ever sourced inside install-k8s.sh's process — so nothing in the bootstrap can name itself, and `install.sh` in `TB_TELEMETRY_SOURCES` is unreachable, exactly as you say.

Both comments now say so. `bootstrap` is documented as "install-k8s.sh before step a", not "everything before step a", with download/verify explicitly named as *not* covered and a line saying extending coverage over the fetch is separate work. I kept `install.sh` in `TB_TELEMETRY_SOURCES` — removing it would desync the derivation from gen-manifest.sh's FILES + the bootstrap for no gain — but the comment now says it is unreachable today and not to read its presence as coverage.

No bootstrap telemetry added. Scope unchanged.

One more from your review body, not on a line — `install.run.skipped`. Also correct, and also now written down where someone reading the counter will see it: install.sh:132 reaches a healthy machine and `exec tracebloc`s at :144, before install-k8s.sh is fetched at all, so on curl|bash assess.sh's gate is never reached. What the event actually counts is direct `./install-k8s.sh` re-runs, plus the curl|bash runs where the bootstrap's own health check did not bail (`--force`, `TRACEBLOC_FORCE_REINSTALL`, a pinned `REF`/`BRANCH`, or no `tracebloc` on PATH). Left as a comment rather than a rename — the name is right, the reading of it was not.

LukasWodka:
bugbot run

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,790 @@
...
+  # separable because tracebloc.install.exit_code is already an attribute —
+  # exit_code=2 is the re-run handoff, 130/143 the user's own Ctrl-C — which is
+  # why the exit code is an attribute rather than something the name carries.
+  case "$code" in
+    0)       if [ -n "$_TB_TELEMETRY_SKIPPED" ]; then
+               event="install.run.skipped"
+             else
+               event="install.run.succeeded"
+             fi ;;
+    2|130|143) event="install.run.cancelled" ;;
```
saadqbal: Not blocking, and this is a consequence of what I asked for rather than something you got wrong — but worth writing down: `2` is a sentinel the installer chose, and it is also a status ordinary tools produce. `grep` exits 2 on a file error, `curl` on a failed init, `tar` on a fatal. Any bare one of those failing under `set -e` exits the script with 2, and that now renders `cancelled` with no `error.type` — a hard failure removed from the numerator instead of misfiled in it, which is the harder direction to notice.

`install_cleanup` has made the same assumption since client#681 (it already prints "Re-run required" for a stray 2), so it isn't new — this change just makes it load-bearing for the rate. If you want it airtight later, keying on a marker set at the `exit 2` site rather than on the number would do it, and the agreement check's derivation from common.sh's branch would carry over unchanged.

saadqbal (dismissed):
All five addressed, and the fixes are better than what I suggested in two places — deriving "is the installer running inside this directory?" instead of asking install.sh to export a second name is the right call for a bootstrap users may have curl'd months ago, and the `pwd -P` catch on both sides is the kind of thing only re-running the reproduction finds.

Verified on this head rather than taken: the agreement check is green including the new behavioural version corpus and the exit-2 branch; `bats scripts/tests/telemetry.bats` is 36/36; and I re-checked the four shape guards under `/bin/bash` 3.2.57 — `[[ =~ ]]` refuses every embedded-newline input and still admits the legal spellings, and the `{1,15}` interval works there.

One non-blocking note below on exit 2. `quality / gitleaks` is still the only red check and still needs the squash from a human.

scripts/lib/telemetry.sh
LukasWodka: Fixed — and you were right that this is the harder direction to notice.

`2` is now only a handoff when an `exit 2` site **declared** it. `telemetry_rerun_handoff` is called at the deliberate site (`gpu-nvidia.sh`), and a bare `2` with no marker renders `install.run.failed` with `error.type=unexpected_exit_2`.

That value is deliberately specific rather than folded into a generic bucket, for the reason your comment implies: the exit status is **not a value we chose**, so no other reading of it — the phase's bucket, a readiness diagnosis — is more trustworthy than saying exactly that. `grep`'s file error, `curl`'s failed init, `tar`'s fatal, or anything escaping under `set -e` now lands in the numerator with a label an operator can act on, instead of being removed from it.

Your framing is what made this worth doing: a hard failure **removed from the numerator** is worse than one misfiled inside it, because the second is visible and the first looks like health. Silence beats a wrong label only when nobody is computing a rate off it.

You noted `install_cleanup` has assumed this since client#681 — agreed, so this change is what makes the assumption load-bearing rather than what introduced it. Keying on the marker rather than the number is what closes it for good, and it stays derived from `common.sh`'s branch so the agreement guard carries over.

LukasWodka:
bugbot run

saadqbal (dismissed):
Re-approving — the marker is the right shape and goes further than the note asked for. Declaring at the `exit 2` site, clearing at source time so an inherited `_TB_TELEMETRY_RERUN_HANDOFF=1` can't pose as a handoff, and failing closed toward `failed` for an undeclared 2 all land the way round I'd want. `unexpected_exit_2` as its own row rather than the phase bucket is better than what I suggested: a stray 2 stays distinguishable from a real prerequisite failure instead of just being counted.

(d4) deriving the site list from `grep -n 'exit 2'` over the runtime, and failing closed on zero sites with the `TRACEBLOC_DOCKER_FIRST_RUN_EXIT` precedent named, is the part that will still be true in a year.

Verified on a00b1c9: agreement check green (13 checks, including both exit-2 directions and the site sweep), `bats scripts/tests/telemetry.bats` 40/40, `gpu-nvidia.bats` green, `gen-manifest.sh --check` current, `shellcheck -S warning -x` clean on the three touched files apart from the pre-existing SC2034 at gpu-nvidia.sh:253. `dispatch_gpu_setup` is a plain call chain from `install_linux`, no subshell, so the marker really does survive to the trap — test 30 pins it.

Two follow-ups, neither for this PR: `install_cleanup` still prints "Re-run required. Complete the step above" for an undeclared 2, so the user gets the message telemetry now correctly refuses to believe — it can key on the same marker. And the dead `TRACEBLOC_DOCKER_FIRST_RUN_EXIT` read at common.sh:1022 deserves its own ticket now that you've traced its producer to 8c3a3d4.

`quality / gitleaks` is still the only red check, still needs the squash.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,872 @@
...
+        # that matters more — a second chmod on the spool is unreachable belt and
+        # braces that no test can redden, which is how a guard nobody has watched
+        # fail gets shipped. One chmod, on the inode that survives.
+        chmod 600 "${spool}.tmp" 2>/dev/null || true
+        mv "${spool}.tmp" "$spool" 2>/dev/null || rm -f "${spool}.tmp" 2>/dev/null || true
+      else
+        rm -f "${spool}.tmp" 2>/dev/null || true
+      fi
+    fi
+    return 0
```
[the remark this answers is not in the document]
LukasWodka: Reproduced, and the framing is right — the fallback was unreachable in exactly the case it was added for. Fixed in 91ce4fa.

I reproduced both halves, each with an anchor so the fixture can't pass by accident:

**Case A — a 0500 data dir with no `telemetry/` yet.** Your point about `_choose_log_file` is the part that makes this bad, so I checked it by *calling* it rather than asserting it:

```
ANCHOR _choose_log_file -> /var/folders/.../a/tmp/tracebloc-install-QpKvrn
ANCHOR OK: the log fell back out of the data dir
ANCHOR OK: telemetry/ was NOT created (the mkdir -p failed)
data-dir spool  : <none>
fallback spool  : <NONE — THE EVENT IS LOST>
```

So on `curl | bash` the loss is total, not partial: `log` did write, but into the mktemp log inside the bootstrap's own doomed `TMPDIR` — the 46a33de hole reached by a second route.

**Case B — `pending.jsonl` replaced by a directory,** so the append cannot open it. Same outcome, and this one holds for root too:

```
/…/telemetry.sh: line 809: /…/telemetry/pending.jsonl: Is a directory
fallback spool  : <NONE — THE EVENT IS LOST>
```

That stderr line is a second, smaller bug on the same statement: `2>/dev/null` came *after* the `>>`, and a failing `>>` is reported by the **shell**, not by printf — so the installer printed the customer's path out of an EXIT trap. Redirections apply left to right, so the `2>/dev/null` now precedes the `>>` and the diagnostic goes with the rest.

**On not writing it twice** — I took control flow over a flag. The `return 0` now sits after a *successful* append and nowhere else, so the fallback is reachable only on a path that wrote nothing; there is no boolean to keep in step. Everything after the append (chmod, trim, mv) may fail freely, because the line is already on disk and re-filing it would turn one install into two rows — which would quietly double the denominator of the failure rate this feature exists to produce. I moved the trim out into `_telemetry_trim_spool` so the append's success is the last thing in that branch: inline, the trim sat between the append and the return, which is what made the original `|| return` easy to write.

Mutation-proved, anchors asserted first:

| | |
|---|---|
| both halves return instead of falling through | RED |
| only the mkdir half returns | RED |
| only the append half returns | RED |
| a successful append no longer returns (every install filed twice) | RED |

The last one is the guard against my own fix's opposite error, and it caught a defect in this commit's first draft. My trim-failure fixture broke `tail` with a PATH shim — and `common.sh:8` *prepends* the system directories to `PATH`, so the real `tail` ran and both assertions passed against a trim that had worked perfectly. Now a shell function (which beats `PATH` lookup outright) plus a behavioural anchor: `SPOOL_MAX=3` against 7 spooled lines must leave 7, or the trim ran after all. I also covered the trim's `mv`-failure cleanup, which was the one line in there no mutation could redden.

1168 bats tests green, `make lint` clean, manifest regenerated.

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,872 @@
...
+    class="$(telemetry_error_class "$code" "$phase" "$state" "$_TB_TELEMETRY_RERUN_HANDOFF")"
+    _telemetry_in_set "$class" "$TB_TELEMETRY_ERROR_CLASSES" || class="unclassified"
+    _telemetry_attr "error.type" "$class"
+    # Where the shell died, to the file and line — never the path that reached
+    # them. There is deliberately no exception.* set: bash has no stack trace,
+    # and TB_ERR_CMD is unexpanded command text, which is free text.
+    if [ -n "${TB_ERR_LOC:-}" ]; then
+      _telemetry_attr "tracebloc.install.source" "$(_telemetry_source_basename "$TB_ERR_LOC")"
+      _telemetry_attr "tracebloc.install.source_line" "$(_telemetry_source_line "$TB_ERR_LOC")" int
+    fi
```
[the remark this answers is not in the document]
LukasWodka: Reproduced and fixed in 91ce4fa — and it is reachable on a real installer path, not only on a foreign basename.

```
/home/someone/evil.sh:9   -> "tracebloc.install.source_line":9    (no source key)
?:118                     -> "tracebloc.install.source_line":118  (no source key)
```

The second one is the one that convinced me this is worth more than its severity: `?:118` is what the ERR trap at `install-k8s.sh:118` produces whenever `BASH_SOURCE` is empty, so this wasn't waiting on somebody sourcing a stranger's script — the installer can emit it about itself. A line number with no file reads as information, groups like information, and points at line 118 of nothing.

**Deriving the gate rather than restating it:** the condition is now `_telemetry_source_basename`'s own exit status, so `TB_TELEMETRY_SOURCES` stays the single declaration of what counts as one of our files. A second copy of that list here — or a hand-written "is it ours" test — would agree with itself and drift from the set the renderer actually uses.

**Not gated the other way, deliberately.** `TB_ERR_LOC` has exactly one writer in the tree, the ERR trap at `install-k8s.sh:118`, and it always appends `:${LINENO}` — so source-without-line is unreachable, and a branch for it would be belt and braces no test could redden. A file with no line is honest information anyway; only the line without a file lies.

The test writes its inputs down independently and derives each expectation by asking the vocabulary, so getting the pairing wrong in either direction reddens it. Mutations:

| | |
|---|---|
| source and line gated independently again | RED |
| the same mutation vs the **pre-existing** source test | RED — it never checked the line, which is why this shipped |
| `common.sh` dropped from `TB_TELEMETRY_SOURCES` | **GREEN — survivor** |
| `cluster.sh` dropped (the independent anchor names it) | RED |

Reporting the survivor rather than counting it as a pass: it is by design, but I checked that rather than assuming it. Because the test derives its expectation *from* the vocabulary, a consistent removal moves both sides — so I confirmed the gate that owns the vocabulary catches it instead, and `telemetry-vocabulary-agreement.sh` goes red with `source basenames disagrees with gen-manifest.sh's FILES array`. That is the right split: this test owns the pairing, that one owns the list. The independent anchor at the end also names `cluster.sh` in its own right, which is why the second removal reddens here too.

LukasWodka:
bugbot run

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,935 @@
...
+             else
+               event="install.run.succeeded"
+             fi ;;
+    2)       if [ -n "$_TB_TELEMETRY_RERUN_HANDOFF" ]; then
+               event="install.run.cancelled"
+             else
+               event="install.run.failed"
+             fi ;;
+    130|143) event="install.run.cancelled" ;;
+    *)       event="install.run.failed" ;;
```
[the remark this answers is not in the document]
LukasWodka: Reproduced and fixed. Good catch — and it is worth more than the label.

I reproduced it through the real INT and EXIT traps rather than just the render function, with the skipped latch set identically in both arms:

```
skipped run, ordinary exit 0     -> install.run.skipped
skipped run, then Ctrl-C (130)   -> install.run.cancelled     <- the defect
```

`cancelled` asserts that an install was cancelled. On this path there was no install to cancel — `_assess_handoff` has already marked the run skipped and is only holding the user on the home screen. So the row lands in the denominator of "how often do installs not complete", inflating it with runs that never attempted anything. That is the same class and the same direction as the `--help` bug fixed earlier on this branch: it makes the product look worse while telling nobody anything actionable. And Ctrl-C is the *most ordinary* thing a user does on that screen, so this was not a rare path.

The 130/143 branch now consults `_TB_TELEMETRY_SKIPPED`. No contract change: §6.4's verb list is untouched, nothing added, removed or redefined, and both verbs were already registered and already emitted by this file. Only which of the two a run books changes.

**The asymmetry is deliberate, and it is the part worth reviewing.** The flag is consulted only on the exits that mean nothing was installed — 0 and the two signals. A skipped run that then dies with a real non-zero stays `failed` and keeps its `error.type`. The shorter spelling — hoisting a blanket "skipped wins" ahead of the case — would have swallowed a genuine failure, so it is not used, and there is a test that reddens if anyone reaches for it later.

I also rewrote the comment that claimed 130/143 were "unconditional", since this change makes it false.

Mutations, anchors asserted first:

| | |
|---|---|
| 130/143 unconditional again (the defect) | RED |
| **130/143 ALWAYS skipped** | **RED** — the positive control |
| the flag test inverted | RED |
| only 130 consults the flag, 143 forgotten | RED |
| blanket skipped-wins hoisted ahead of the case | RED |
| only the `*)` failure branch consults the flag | RED |

The second one is the control that makes the rest mean anything: without it, "renders skipped" would be satisfied by a change that never renders `cancelled` at all, quietly deleting the interrupted-install signal. The last one exists because the blanket-hoist mutation tripped an earlier `exit_code` assertion before reaching the failure-swallowing one — it mutates only the `*)` branch so `exit_code` is untouched, and proves that assertion is live rather than decorative.

`telemetry-vocabulary-agreement.sh` passes unchanged: the case statement gains no new event-name literal, and its sweep already covered `130 x skipped=1`, so the input domain was already derived from the producer rather than hand-listed.

44 telemetry tests green, `make lint` clean, manifest regenerated.

LukasWodka:
bugbot run

scripts/lib/telemetry.sh
```diff
@@ -0,0 +1,962 @@
...
+# bootstrap's own: install.sh refuses to fetch from anything that is not an
+# immutable vX.Y.Z release tag, so a TB_VERSION that does not match that never
+# came from a release. The generic token shape was not enough on its own —
+# `v1.9.3-<64 arbitrary chars>` satisfies it, which makes the version column a
+# 64-byte free-text channel. telemetry-vocabulary-agreement.sh checks this
+# against install.sh's — byte-for-byte AND, since client#747, verdict-for-verdict
+# over a corpus, each side evaluated by the operator its own file uses. The byte
+# check alone was not enough and said it was: the two regexes were identical
+# while install.sh's `[[ =~ ]]` refused an input this file's `grep -qE` admitted.
+TB_TELEMETRY_VERSION_RE='^v[0-9]+\.[0-9]+\.[0-9]+([.-][A-Za-z0-9.]+)?$'
```
saadqbal: nit: this is the one field in the record with no length bound — `([.-][A-Za-z0-9.]+)?` is open-ended where the token shape is `{1,64}` — and `service.version` is also one of the seven resource fields that go straight into the `printf` rather than through `_telemetry_attr`, so the "one writer" argument doesn't cover it.

The "install.sh refuses anything that isn't a release tag" reasoning doesn't constrain it either. `common.sh:1128` is `TB_VERSION="${TB_VERSION:-${TRACEBLOC_INSTALL_REF:-}}"`, so the caller's own `TB_VERSION` wins over the ref the bootstrap exported — and on the direct `./install-k8s.sh` path (or under `TRACEBLOC_ALLOW_UNVERIFIED=1`) install.sh's gate never ran at all. `TB_VERSION=v1.0.0-alice.hospital.internal ./install-k8s.sh` lands verbatim in the version column.

Not a leak — the charset has no `/`, `:`, `@` or space, so a path or a proxy credential still can't fit — but it's an unbounded, environment-supplied string in a record where everything else is capped. A `{1,32}` on the tail would close it.

scripts/tests/telemetry-vocabulary-agreement.sh
```diff
@@ -0,0 +1,418 @@
...
+  [ "$rendered" != "0.0.0-unknown" ]
+}
+
+version_corpus=(
+  'v1.9.3'                       # a plain release tag: both must admit
+  'v1.9.3-rc.1'                  # the pre-release suffix the regex allows
+  'v1.9.3.post1'                 # the dotted suffix it allows
+  'main'                         # a branch name: both must refuse
+  'v1.9'                         # two segments: both must refuse
+  'v1.9.3-'"$(printf '%064d' 0)"  # 64 trailing chars: both must refuse
```
saadqbal: nit: this row doesn't test what the comment says. `([.-][A-Za-z0-9.]+)?` matches 64 zeros, so `v1.9.3-000…0` (71 chars) is *admitted* by both gates, not refused:

```
$ RE='^v[0-9]+\.[0-9]+\.[0-9]+([.-][A-Za-z0-9.]+)?$'
$ c="v1.9.3-$(printf '%064d' 0)"; [[ $c =~ $RE ]] && echo admitted
admitted
```

The check still passes, because it only asserts the two gates agree — but the row is proving "both admit" while the comment reads as a length bound being exercised. Same underlying point as the note on `TB_TELEMETRY_VERSION_RE`.

saadqbal (approved):
Post-hoc read (this is already on `develop` as c564a15), and a privacy-focused one: every field the record can carry, traced to its source.

The short version is that the shape gate holds. Nine attributes, all of them either a closed vocabulary checked at the render boundary (`event.name`, `phase`, `client_state`, `error.type`, `source`) or a bounded integer (`exit_code`, `duration_ms`, `cli_on_path`, `source_line`, the eight `phase_*_ms`). `_telemetry_attr` is the only writer for all of them, it drops rather than repairs, and `[[ =~ ]]` closes the newline hole a `grep` would have left. I could not find a free-text channel: `TB_ERR_CMD` is never read, `TB_ERR_LOC` is reduced to a basename that must be in the source vocabulary, and there is no `stderr` capture anywhere in the file. The bats "derived guard" walks the whole rendered object and requires every value to be int-or-token, which is a much better test than greping for the absence of a canary.

Two things about *enforcement* rather than shape, neither of them a defect here:

The seven `resource` fields are printf'd straight into the JSON and do not pass through `_telemetry_attr` — so the "one writer" argument in the header covers `attributes` only. Each of the seven is independently constrained today (`installer`/`install` constants, three closed `case` sets, the version regex, a `$RANDOM` id), so nothing gets past. But there is no allowlist of field *names*, and the token shape `^[A-Za-z0-9._-]{1,64}$` admits an identifier as readily as it refuses a path: `alice`, `db-01.internal`, `hospital-vm-03.example.com` all match it. What is enforced is "no path, no credential, no whitespace", which is what the title claims; "no identifier" is currently a property of the code's shape rather than of a gate. backend#2165's allowlist is the stronger form if this grows.

Opt-out is honoured on every path that can emit, because `telemetry_emit_outcome` is the only one and the check sits ahead of both render and deliver. It is documented in `--help` and the golden — but not in `install-k8s.sh`'s own "Environment variable overrides" header block (where `TRACEBLOC_ALLOW_NETWORK_FS`, `TRACEBLOC_SKIP_LEFTOVER_GUARD` etc. live), not in `docs/SECURITY.md`, and `install-k8s.ps1` has no equivalent. Worth a follow-up rather than a change here. The help copy also doesn't mention that a file appears at `$HOST_DATA_DIR/telemetry/pending.jsonl` and keeps 50 runs; a customer reading "records ONE outcome event per run" wouldn't expect that.

On the parts of the brief that turned out to be moot: there is no transport. No `curl`, no POST, no endpoint, no token, nothing with a timeout — delivery is `log` plus a local 0600 spool in a 0700 dir. So the installer cannot block, cannot hang on an air-gapped site, and cannot fail an install on a telemetry error, and there is no `-k`/`--insecure` question to ask. Nothing leaves the customer's machine in this PR at all; the review that actually decides "does customer data leave the environment" is #1905/#1906, not this one.

Exactly-once holds: the `_TB_TELEMETRY_EMITTED` latch is set before any other work in `telemetry_emit_outcome`, and the run-started latch keeps `--help` out. Zero-times cases are all deliberate and documented in the code — `install.sh`'s own fetch/manifest/cosign failures (different process, different EXIT trap), `--diagnose` (clears the trap), `prepare-host` (swaps it). The outcome is right on the paths that matter, including the two that were wrong earlier in the branch (declared exit 2, and Ctrl-C on the skipped handoff).

Manifest: regenerated correctly. I recomputed all 19 entries against the tree at this head rather than trusting the diff — every one matches, including the three files whose content changed and the new `telemetry.sh`.

Coordination note for backend#2123, not a defect here: its metric filters key on `service.name` and `deployment.environment`, and both live inside the nested `resource` object in this record, so a filter written as `$.service.name` won't match — it needs `$.resource.service.name`. Worth settling before the filters are cut, since nothing emits to CloudWatch until #1905 anyway.

Split is roughly 1050 lines of implementation and wiring against 1830 of tests and drift guards, plus 10 of help copy. The vocabulary-agreement script is the best thing in the PR — deriving the sets from their producers and failing closed on an inert parse is the right shape, and exercising the classifier over its cross-product instead of restating it is genuinely better than a second declaration.

Two nits inline.

(Process note: I did not run `/code-review` on this one — this session's working directory is the `tracebloc-engine` checkout, not `client`, and that skill resolves its target against the cwd, so it would have reviewed the wrong repository's diff. Everything above is from a direct read of `origin/pull/747/head`.)

LukasWodka:
/fr-pass

PR #748, opened 2026-08-18
release-train: develop -> staging

LukasWodka:
Automated promotion by the release train (RFC-0008 D14). Head is the train-managed `release-train/to-staging` branch (a mirror of `develop`), so it never collides with a human PR. Merged only when the fr-gate is green.

---

> [!NOTE]
> **Medium Risk**
> Touches installer exit/telemetry paths, required CI gates, and jobs-manager RBAC across namespaces; chart and docs changes are additive but egress lockdown misconfiguration could still false-pass if probe hosts are wrong.
> 
> **Overview**
> Automated **develop → staging** promotion carrying several independent feature/fix trains.
> 
> **Installer telemetry (backend#1907)** adds `scripts/lib/telemetry.sh` and wires it through `install-k8s.sh`, `common.sh` (`install_cleanup`, `step_header`, `error`), assess skip/handoff paths, and GPU reboot `exit 2` handoff. Each committed install emits one structured outcome event (phase timings, error class, opt-out via `TRACEBLOC_NO_TELEMETRY` / `DO_NOT_TRACK`); delivery is log + bounded spool until backend ingest exists. `telemetry-vocabulary-agreement.sh` runs in required **Source-of-truth drift**; `.gitleaks.toml` allowlists telemetry canary fixtures.
> 
> **CI (#753)** removes redundant shellcheck from `installer-tests` and `standard-checks` (runner image already has it); `make lint` derives parse + shellcheck from the tree/`git ls-files`. **bats** installs are bounded, retried, and job timeouts increased. Kanban workflow mints a **release-train GitHub App** org installation token instead of a PAT.
> 
> **Helm chart (1.9.47)** grants jobs-manager read access to **DaemonSets** (heartbeat version inventory), optional **Role/RoleBinding** in `nodeAgents.namespace` when split from release ns, and passes **`NODE_AGENTS_NAMESPACE`** (empty when `resourceMonitor` is off).
> 
> **Installer behavior**: shared **`_mysql_engine_decision`** so preflight and `_assert_engine_runs_on_this_arch` don’t refuse fresh arm64 on native 8.4; macOS **Docker Desktop nudge** before privileged tier; bounded **`_docker_answers`**; device sign-in **retry** in `provision.sh`; fail-closed **`parse`** in Makefile.
> 
> **Docs**: **SECURITY.md** / **SEAL-CHECK.md** expand §8.2 egress lockdown (CNI pre-flight, drain/flip/verify/rollback, **probe-host false-pass** on TCP :443).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 526fc49d88ad7822a20394ccc6b33c45d27fb7af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LukasWodka:
bugbot run

scripts/lib/install-client-helm.sh
```diff
@@ -870,15 +891,75 @@ _resolve_mysql_engine() {
...
+  # this host's data that may be false.
+  case "${TB_MYSQL_ENGINE_REASON:-}" in
+    explicit)
+      warn "TB_MYSQL_ENGINE=5.7 was requested, and the MySQL 5.7 image is amd64-only — it cannot run on ${ARCH}."
+      hint "Drop TB_MYSQL_ENGINE (or set TB_MYSQL_ENGINE=8.4) to use the native multi-arch 8.4 engine — fresh data directories only."
+      hint "To keep 5.7, enable amd64 emulation and re-run:"
+      hint "  docker run --privileged --rm tonistiigi/binfmt --install amd64"
+      error "MySQL 5.7 was requested explicitly and cannot run on ${ARCH} without amd64 emulation." ;;
+    existing-datadir)
+      warn "This host holds existing MySQL 5.7 data, so the install must keep the MySQL 5.7 engine — and that image is amd64-only."
```
saadqbal: This claims data that may not exist, and the remedy under it can't clear the trigger. `existing-datadir` also fires on `existing_id` alone, and that comes from `detect_installed_client` — `helm list -A` plus the release's clientId, nothing to do with the host datadir.

Worst case is `TB_STORAGE_MODE=node-local`: neither `_ensure_tracebloc_dirs` nor `_ensure_release_dirs` runs, so both `_mysql_dir_has_content` probes are guaranteed false and `existing_id` is the only trigger. arm64, no binfmt, existing release → "This host holds existing MySQL 5.7 data" and "install into an empty data directory: --data-dir=…". `--data-dir` doesn't change `helm list`, so the next run refuses identically. And the release could be an 8.4 one whose values.yaml sticky marker was lost with the old data dir — then we're telling them their data is 5.7-format when it isn't.

Worth splitting the reason (`existing-release` vs `existing-datadir`), or at minimum naming "uninstall the existing release" as the remedy. Same text at preflight.sh:453.

scripts/lib/install-client-helm.sh
```diff
@@ -870,15 +891,75 @@ _resolve_mysql_engine() {
...
+}
+
+# ── The arch gate, asked where the engine is actually known (backend#2047) ────
+# preflight's _pf_arch asks the same question early (that is the fast-fail), but
+# it runs before the Helm release and the per-release namespace are known — so an
+# existing edge can still look fresh there and only resolve to the amd64-only 5.7
+# image here. Without this second ask, letting the fresh case through preflight
+# would let that edge proceed to an exec-format CrashLoop with no earlier signal.
+# Same conditions as _pf_arch, same probe: Linux, non-amd64, no binfmt, no escape
+# hatch. macOS is excluded because assert_amd64_emulation (#433) owns it there.
```
saadqbal: `assert_amd64_emulation` doesn't ask the engine rule, so backend#2047 stays broken on Apple Silicon. `install_macos` calls it on both the tier-0 and normal paths (setup-macos.sh:604) and it hard-errors "amd64 emulation unavailable … the client images are amd64-only" whenever Rosetta amd64 is off — before `install_client_helm` runs. A fresh Mac that `_mysql_engine_decision` would serve natively on 8.4 is still refused and told to enable Rosetta.

That's the same fresh-arm64-refused-for-nothing symptom, on the arm64 host most of our users actually have. If macOS is out of scope for this PR, worth saying so here rather than "owns it there", which reads as covered.

saadqbal (dismissed):
Two things on the arm64 work. Also, one that isn't inline-able: preflight's datadir probe resolves TB_NAMESPACE to `tracebloc` because `provision_client` only exports the minted slug later, so on a real edge the per-release probe (`$HOST_DATA_DIR/<slug>/mysql`) misses and the refusal still lands after cluster setup + sign-in. The comment calls that the residual case, but with a minted namespace it's the common case, not the edge. Bugbot's two are still open. Manifest checksums check out. Fixes go on `develop`, not this branch.

LukasWodka:
bugbot run

LukasWodka:
bugbot run

LukasWodka:
bugbot run

LukasWodka:
bugbot run

scripts/lib/install-client-helm.sh
LukasWodka: Addressed in #752 (merged): the reason is now split — a live Helm release (`existing_id`) resolves to **`existing-release`**, and `existing-datadir` fires ONLY on real host datadir content (`_mysql_dir_has_content`). So this warn no longer claims host 5.7 data on an `existing_id`-only trigger; the `existing-release` path says the data format is unknown and does not offer `--data-dir`. The `node-local` worst case you named now lands on `existing-release`, not here. Resolving.

LukasWodka:
bugbot run

LukasWodka:
Resolving threads to unblock the staging promotion:
- `Makefile` macOS xargs (Medium, Bugbot) → tracebloc/backend#2200 (shipped-with-ticket, non-High policy)
- `scripts/lib/install-client-helm.sh` (@saadqbal) → **addressed** by client#756 (reviewed by @aptracebloc); thread is outdated.

LukasWodka:
/fr-pass

PR #749, opened 2026-08-18
docs(#248): the §8.2 egress lockdown needs an enforcing CNI — say which, and gate on it

saadqbal:
Closes #248 — all three folded follow-ups (#248, #257, #258).

## Why

The §8.2 training-pod egress lockdown only blocks anything on a CNI that enforces **egress** NetworkPolicy. Verified 2026-06 on `tb-client-dev-templates` (EKS, **self-managed** VPC CNI, NetworkPolicy disabled): a DNS-only egress policy did **not** block `https://example.com` — the probe returned `200`.

On such a fleet, flipping `allowExternalHttps=false` is cosmetic: the rule renders and nothing enforces it. The docs didn't say which CNIs actually enforce, and the rollout had no gate that would have caught it.

## What changed

**`docs/SECURITY.md`**
- **§5.1** — a concrete enforcing-CNI list, and an EKS row that distinguishes the `vpc-cni` **managed add-on** with `enableNetworkPolicy=true` (enforces, v1.14+) from a **self-managed DaemonSet** (does not — the flag alone is insufficient without the control-plane PolicyEndpoint controller). Records the dev-EKS finding.
- **§4.2** — the policy bullets are a *request* to the CNI, not a guarantee; the prerequisite is now linked from where the rules are stated.
- **§6.2** — an **egress pre-flight probe**: DNS-only policy in a throwaway namespace, the connect MUST fail. This is the gate the old rollout lacked.
- **§8.2** — rollout is now **Gate 0 pre-flight → drain → flip → `helm test` the egress-enforcement check → real experiment**, with an explicit rollback. Drain is called out because the policy change applies to **running** pods too.

**`docs/SEAL-CHECK.md`**
- A per-fleet **production flip/rollback runbook** alongside the existing local k3d one, on `--reset-then-reuse-values` — a plain `--reuse-values` re-applies the stored `false` and silently defeats the rollback.
- The **probe-host false pass** (folded #257): the check reads a refused connect (curl exit 7) as "blocked", and a host with *nothing listening on :443* refuses identically — so a wrong `enforcementProbeHost` passes without testing anything.

**`client/values.yaml`** — the same false-pass caveat where the operator actually sets the host.

## Scope

Docs plus one values comment. No template renders differently.

`Chart.yaml` 1.9.45 → 1.9.46: `values.yaml` is packaged chart content, so `chart-version-guard` requires the bump (same as #717, a values-comment-only change).

## Verification

- `helm lint ./client` — clean
[...]

PR #750, opened 2026-08-18
feat(rbac): let jobs-manager read daemonsets for the heartbeat version inventory

saadqbal:
Companion to **tracebloc/client-runtime#348** (client-runtime#152). Refs client-runtime#152.

## Why

client-runtime#152 adds a version inventory to the edge-device heartbeat, assembled from the configured images on the chart's own workloads. The jobs-manager Role already grants read on `deployments`, which covers jobs-manager, pods-monitor, requests-proxy, mysql-client and egress-proxy.

**resource-monitor is a DaemonSet** — the one workload that isn't a Deployment — and `daemonsets` was granted nowhere in this Role.

Without the grant the list 403s. The runtime degrades rather than failing (it lists the two kinds separately for exactly this reason), so the symptom isn't an error anyone would chase: resource-monitor is reported as `null` forever, which reads as "not deployed".

## What

`apps: ["deployments"]` → `apps: ["deployments", "daemonsets"]`, read-only (`get, list, watch`), in **both** RBAC branches — an edge on `clusterScope: false` needs it just as much. Namespace-scoped in the Role branch.

Two helm unit tests assert both branches. The two rule lists are maintained separately in this template and have drifted before, so the assertion is per-branch rather than shared.

## Verification

- `helm lint ./client` — clean
- `helm unittest ./client` — **469 passed, 31 suites** (467 + the 2 new)
- `scripts/chart-version-guard.sh` — `client chart content changed and client/Chart.yaml 'version:' was bumped. ✓`

## Note on the version bump

Bumped 1.9.45 → 1.9.46. [#749](https://github.com/tracebloc/client/pull/749) (docs, also open) bumps to the same version — whichever merges second needs a trivial bump to 1.9.47.

---

> [!NOTE]
> **Medium Risk**
> Adds namespace-scoped RBAC and a load-bearing env contract for heartbeat inventory; misconfiguration could cause 403s or silent null resource-monitor reporting, but grants are read-only and gated on layout flags.
> 
> **Overview**
> Enables jobs-manager to include **resource-monitor** in the edge heartbeat version inventory (companion to client-runtime changes). The main jobs-manager **ClusterRole/Role** now grants read-only access to **`daemonsets`** as well as **`deployments`**.
> 
> Because resource-monitor is a **DaemonSet in `nodeAgents.namespace`**, split-namespace installs (`clusterScope: false`) get an extra **Role/RoleBinding** in that namespace (via `tracebloc.rbacNodeAgentsName`), mirroring image-refresh RBAC; the pair is skipped when the node-agents namespace is the release namespace to avoid name collisions.
> 
> The jobs-manager **api** container receives **`NODE_AGENTS_NAMESPACE`** (the configured namespace when resource monitor is on, **empty** when `resourceMonitor: false` so the runtime does not list a disabled workload). Chart version **1.9.47**; helm unit tests cover RBAC branches and the env wiring.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bf85a23678608d045cdd4bc2692ab4d1e58df739. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/templates/rbac.yaml
+++ b/client/templates/rbac.yaml
@@ -27,8 +27,17 @@ rules:
   - apiGroups: [""]
     resources: ["pods", "pods/log", "events"]
     verbs: ["get", "list", "watch"]
+  # client-runtime#152: jobs-manager reports a version inventory on the
+  # heartbeat, assembled from the configured images on the chart's own
+  # workloads. `deployments` covers jobs-manager, pods-monitor,
+  # requests-proxy, mysql-client and egress-proxy; `daemonsets` is needed
+  # for resource-monitor, the one workload that is not a Deployment.
+  # Read-only, and scoped to the release namespace in the Role branch.
+  # Without daemonsets the list 403s and resource-monitor is reported as
+  # null forever — the runtime degrades rather than failing, so the symptom
+  # is a permanently blank field, not an error anyone would chase.
   - apiGroups: ["apps"]
-    resources: ["deployments"]
+    resources: ["deployments", "daemonsets"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["metrics.k8s.io"]
     resources: ["pods"]
@@ -126,8 +135,17 @@ rules:
   - apiGroups: [""]
     resources: ["pods", "pods/log", "events"]
     verbs: ["get", "list", "watch"]
+  # client-runtime#152: jobs-manager reports a version inventory on the
+  # heartbeat, assembled from the configured images on the chart's own
+  # workloads. `deployments` covers jobs-manager, pods-monitor,
+  # requests-proxy, mysql-client and egress-proxy; `daemonsets` is needed
+  # for resource-monitor, the one workload that is not a Deployment.
+  # Read-only, and scoped to the release namespace in the Role branch.
+  # Without daemonsets the list 403s and resource-monitor is reported as
+  # null forever — the runtime degrades rather than failing, so the symptom
+  # is a permanently blank field, not an error anyone would chase.
   - apiGroups: ["apps"]
-    resources: ["deployments"]
+    resources: ["deployments", "daemonsets"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["metrics.k8s.io"]
     resources: ["pods"]
--- a/client/templates/rbac.yaml
+++ b/client/templates/rbac.yaml
@@ -27,15 +27,8 @@ rules:
   - apiGroups: [""]
     resources: ["pods", "pods/log", "events"]
     verbs: ["get", "list", "watch"]
-  # client-runtime#152: jobs-manager reports a version inventory on the
-  # heartbeat, assembled from the configured images on the chart's own
-  # workloads. `deployments` covers jobs-manager, pods-monitor,
-  # requests-proxy, mysql-client and egress-proxy; `daemonsets` is needed
-  # for resource-monitor, the one workload that is not a Deployment.
-  # Read-only, and scoped to the release namespace in the Role branch.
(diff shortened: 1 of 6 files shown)


client/templates/rbac.yaml
```diff
@@ -126,8 +135,17 @@ rules:
...
+  # workloads. `deployments` covers jobs-manager, pods-monitor,
+  # requests-proxy, mysql-client and egress-proxy; `daemonsets` is needed
+  # for resource-monitor, the one workload that is not a Deployment.
+  # Read-only, and scoped to the release namespace in the Role branch.
+  # Without daemonsets the list 403s and resource-monitor is reported as
+  # null forever — the runtime degrades rather than failing, so the symptom
+  # is a permanently blank field, not an error anyone would chase.
   - apiGroups: ["apps"]
-    resources: ["deployments"]
+    resources: ["deployments", "daemonsets"]
```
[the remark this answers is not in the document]
saadqbal: Right, and it's the same root cause a reviewer hit on client-runtime#348. Added a second Role + RoleBinding in `nodeAgents.namespace.name`, following the image-refresh split you pointed at. Also wired `NODE_AGENTS_NAMESPACE` into the jobs-manager `api` container — RBAC alone wouldn't have helped, since the runtime was listing the wrong namespace (fixed in client-runtime#348 too). The pair renders only when resourceMonitor is on and that namespace is actually distinct from the release namespace, and the lookup is `dig`-guarded so a `--reuse-values` upgrade predating `nodeAgents` can't nil-pointer this always-rendered template. 472 helm unit tests pass. a76d6f2.

saadqbal:
Bumped to **1.9.47** (was 1.9.46) so this no longer collides with #749.

⚠️ **This creates a merge-order dependency: #749 must merge first.** If #750 lands first, #749 would carry 1.9.46 against a `develop` already on 1.9.47 — a version *regression*, which `chart-version-guard` would not catch: it only checks that a bump happened, not that it moves forward.

client/tests/rbac_test.yaml
```diff
@@ -77,6 +77,88 @@ tests:
...
+    documentIndex: 1
+    asserts:
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["apps"]
+            resources: ["deployments", "daemonsets"]
+            verbs: ["get", "list", "watch"]
+
+  # client-runtime#152 / review PR #750: resource-monitor is a DaemonSet in
```
aptracebloc: Maybe we can get rid of the comments referencing the issue for which the change is made. WDYT?

aptracebloc (commented):
Many long comments, and the issue references seem unnecessary to me. Can we tighten this?

client/tests/rbac_test.yaml
saadqbal: Done — all issue references removed from the templates and tests, including the `(client-runtime#152)` suffixes on the test names, and the comment blocks cut back hard: 84 lines of comment removed for 34 added. I kept only the two bits of "why" that aren't recoverable from the code — that the resource-monitor DaemonSet lives outside the release namespace, and that the empty `NODE_AGENTS_NAMESPACE` is deliberate rather than a no-op. 474 helm unit tests still pass. bf85a23.

One thing worth your call: the surrounding code in `rbac.yaml` does carry ticket refs (`backend#1876`, `client-runtime#217`, `#155`), so this block is now the odd one out in its own file. Happy either way — say the word if you'd rather the file stay internally consistent, or if the broader convention should change.

aptracebloc (approved):

LukasWodka:
/fr-pass

PR #751, opened 2026-08-18
fix(chart): inject the release-scoped pull-secret name for training pods (backend#2119)

LukasWodka:
Part of backend#2119. Pairs with `client-runtime` PR (branch `fix/2119-image-pull-secret`) — **that one is the behaviour change; this one supplies the name it needs.**

## The defect

`client-runtime`'s `job.yaml` hardcoded:

```yaml
imagePullSecrets:
- name: regcred
```

This chart has **never** created a Secret by that name. `tracebloc.registrySecretName` renders `<release>-regcred`, and `docker-registry-secret.yaml` renders only when `dockerRegistry` is set with `create: true`. So the reference could not resolve on **any** release: every training pod named a Secret the kubelet could not retrieve, logged `FailedToRetrieveImagePullSecret`, and fell back to an **anonymous pull of the ~2.7 GB training image**.

Anonymous pulls are rate-limited per source IP, so under concurrency this becomes intermittent image-pull failures that present as random training failures.

**On a mirrored install the consequence is sharper than slowness:** the training pod was the one pod in the release *not* using the credential the operator supplied.

## Why the name is injected rather than written in the runtime

It is release-dependent. A literal in `client-runtime` cannot be correct for every release — and the literal it carried was correct for none. `IMAGE_PULL_SECRET_NAME` is rendered from the same helper as every other `imagePullSecrets` block in this chart, and gated on the same condition that **creates** the Secret, so the three cannot disagree.

When no registry is configured the variable is **absent**, and the runtime reads that as "pull publicly" and sets no `imagePullSecrets` at all — the honest declaration of a public pull, rather than a broken reference to a private one.

Skew-safe in both directions: a runtime older than backend#2119 ignores the variable; a newer one treats its absence as public.

## Verified by rendering, not by reading

| values | result |
|---|---|
| `dockerRegistry.create=true` + server/user/pass/email | `IMAGE_PULL_SECRET_NAME: "myrel-regcred"` — **and** `Secret myrel-regcred` present in the same render |
| no registry configured | variable **absent** (0 occurrences) |

Rendering it was worth doing: my first three attempts failed on chart schema validation (`server` must be an absolute URI, `email` required, `clientId`/`clientPassword` non-empty, `storageClass.provisioner` required when `storageClass.create`), so an unrendered "looks right" review would have proved nothing.

## Manifest

`scripts/manifest.sha256` covers installer **scripts** only — 0 template entries — so a chart-template change needs no regeneration. Confirmed against `scripts/gen-manifest.sh`'s `FILES` array rather than assumed.

---

> [!NOTE]
> **Medium Risk**
> Changes registry authentication wiring for spawned training pods and RBAC on mirrored pull Secrets in kube-system; mitigated by layered schema/template guards and extensive render tests, but wrong names would still break private-registry training pulls.
> 
> **Overview**
> Pairs with **client-runtime** backend#2119: jobs-manager now gets **`IMAGE_PULL_SECRET_NAME`** from the same **`tracebloc.registrySecretName`** helper as every **`imagePullSecrets`** block, only when a registry pull Secret is configured. Training pods can use the real Secret name instead of a hardcoded **`regcred`** that this chart never creates.
> 
> **`dockerRegistry.existingSecret`** is added for bring-your-own **`kubernetes.io/dockerconfigjson`** Secrets. Helpers split **“use a pull secret”** vs **“chart creates the Secret”** so the chart does not overwrite operator Secrets; **`create: true`** and **`existingSecret`** together **fail** at schema and template. Auto-upgrade GPU-namespace RBAC names only the chart-created **`<release>-regcred`** via **`tracebloc.createdRegistrySecretName`**, not the operator’s Secret name.
> 
> **`env.IMAGE_PULL_SECRET_NAME`** is excluded from the generic **`env`** passthrough so it cannot shadow the computed value. Chart **1.9.48**; **helm unittest** and **`chart-pull-secret.bats`** (real **`helm template`**) lock the behavior. **Unit tests** CI installs **Helm v3.15.4** so bats cases cannot skip when helm is missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d0cf5197e81891ca13a4077cee925ccd990dc3b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aptracebloc:
Findings from a review pass. No verdict attached.

### 1. Version collision, and both gates are blind to it

`develop` is already at **1.9.47** — #750 merged today (`8de5d64a`) with the byte-identical bump. Because both sides made the same two-line edit, git auto-merges with no conflict, so nothing surfaces it. Neither gate catches it either:

- `chart-version-guard.sh` diffs `BASE_SHA...HEAD` with `BASE_SHA` = this PR's base (`175ef94e`, pre-#750), so it just sees a `+version:` line.
- `version-bump-gate` only fails when the version is already *released*, and the newest tag is `v1.9.45-rc.1`.

If the release train tags v1.9.47 off develop before this merges, this content lands at an already-published version and either ships dark or overwrites the published tarball — exactly the mode `chart-version-guard.sh`'s own header documents. Rebase and bump to **1.9.48**.

(#750 also touched `jobs-manager-deployment.yaml`: its `NODE_AGENTS_NAMESPACE` inserts *after* `CLIENT_ENV` while this inserts *before* it, so the 3-way merge is clean and semantically fine. The version number is the only real casualty — but the combination has never been rendered by CI.)

### 2. Nothing asserts the new env var

No test anywhere references `IMAGE_PULL_SECRET_NAME` — checked all 32 suites. The Secret side is fully locked at `client/tests/secrets_test.yaml:34-79` (create:true → `RELEASE-NAME-regcred`; create omitted → 0 docs; dockerRegistry omitted → 0 docs). The env-var side rests on a one-off manual render.

A future edit to the `{{- if }}` gate or to the helper drops the var silently with the Helm unit suite still green — the same "gate stopped matching what it guards" class as the bug being fixed. Two mirror cases in `jobs_manager_test.yaml` would close it.

### 3. Nit — comment density

18 lines of comment for 2 lines of YAML, restating the PR body, and it renders into the shipped manifest. The *why-injected-rather-than-literal* sentence earns its place inline; the incident history belongs in the ticket.

### Verified and clean

The gating claim holds exactly as stated, and it's worth recording why: `tracebloc.useImagePullSecrets` is a **helper**, not a values key — `_helpers.tpl:408-412` defines it as `and .Values.dockerRegistry (default false .Values.dockerRegistry.create)`, and there is no `.Values.tracebloc.useImagePullSecrets` anywhere in `values.yaml`. So an operator cannot set it independently of `create`, and the new gate is the identical string to `docker-registry-secret.yaml`'s own condition. Neither divergence is reachable: no path where the env var names a Secret the chart doesn't create, and none where the Secret exists but the var is missing. All three sites call the same name helper, so they can't disagree. Empty-name render is impossible (`{{ .Release.Name }}-regcred`).

Merge order vs the paired `client-runtime` PR is safe in both directions — runtime-first is the interesting one: the runtime finds the var absent, sets no `imagePullSecrets`, and that strictly improves on today's dangling `regcred` reference.

— drafted with Claude Code

client/templates/jobs-manager-deployment.yaml
```diff
@@ -415,6 +415,28 @@ spec:
         # --reset-then-reuse-values upgrades.
         - name: JOB_IMAGE_HOST
           value: {{ printf "%s/" (dig "imageRegistry" "docker.io" (.Values.global | default dict) | default "docker.io") | quote }}
+        {{- if include "tracebloc.useImagePullSecrets" . }}
+        # IMAGE_PULL_SECRET_NAME is the pull Secret jobs-manager puts on the
+        # TRAINING pods it spawns (backend#2119). It has to be injected rather
+        # than written into client-runtime's job.yaml, because the name is
+        # release-dependent: a literal in that repo cannot be right for every
+        # release, and the one it carried ("regcred", unqualified) was right for
+        # none. Every training pod named a Secret the kubelet could not retrieve
```
saadqbal: There is one config where the old literal did work: `dockerRegistry.create: false` plus a Secret hand-made in the namespace and literally named `regcred`. Undocumented, but it is the obvious thing an operator does after reading `FailedToRetrieveImagePullSecret (regcred)` in the events. Once the runtime drops the literal, those installs go from an authenticated pull to an anonymous one with nothing louder than an INFO log — the same silent downgrade this PR exists to remove, just moved. A note next to `dockerRegistry` in values.yaml (or a supported `existingSecret` name the helper can return) would cover them.

client/templates/jobs-manager-deployment.yaml
```diff
@@ -415,6 +415,28 @@ spec:
...
+        #
+        # Rendered from the same helper as every imagePullSecrets block in this
+        # chart, and gated on the same condition that creates the Secret, so the
+        # three cannot disagree. Absent when no registry is configured: the
+        # runtime reads that as "pull publicly" and sets no imagePullSecrets at
+        # all, which is the honest declaration of a public pull.
+        #
+        # Skew-safe both ways: a client-runtime older than backend#2119 ignores
+        # this var, and a newer one treats its absence as "public".
+        - name: IMAGE_PULL_SECRET_NAME
```
saadqbal: nit: `IMAGE_PULL_SECRET_NAME` isn't in the `.Values.env` passthrough exclusion lists (489/568), so an operator who sets `env.IMAGE_PULL_SECRET_NAME` renders it twice and the passthrough copy wins. Same is already true of `JOB_IMAGE_HOST`, so it's not a regression — but this one is different in consequence: the runtime treats a name that doesn't resolve as permanent and dead-letters the experiment, so a shadowed value costs the run rather than a slow pull.

saadqbal (changes requested):
Rendered both cases and ran the suite on the branch — the mechanism is right. Gate, Secret name and env var all come out of the same two helpers, the Secret lands in `.Release.Namespace`, and that is where jobs-manager spawns training Jobs (`get_current_namespace()` reads its own SA namespace), so the name cannot dangle. `rbac.yaml` already grants `secrets: get` in both the clusterScope true and false branches, so the runtime's fail-closed 403 branch isn't reachable by upgrading one side. 467 helm-unittest tests pass with the change in place.

Requesting changes on the version only: `develop` is already at 1.9.47 (#750, 8de5d64), and since both sides made the identical two-line edit git merges this with no conflict and no CI signal — two different chart contents under one version. Merge `develop` and bump to 1.9.48.

aptracebloc already raised that and the missing helm-unittest coverage for the new var, so I'm not repeating either inline. I do think the test is worth adding — the whole fix is one `{{- if }}`, and `global_image_registry_test.yaml` already has the exact two-case shape to copy.

aptracebloc:
Status update on the version collision from my earlier comment — re-checked, and @saadqbal reached the same conclusion independently.

**Still colliding, still latent.** `client/Chart.yaml` is `1.9.47` on both `develop` and this head, byte-identical. `develop` has advanced four commits past #750 (#752, #741, #743, #747), none touching chart content, so 1.9.47 is still the live develop version.

**v1.9.47 is not tagged and not released.** Newest tag and release is `v1.9.45-rc.1` (`4b78f695`, 2026-08-18); there is no v1.9.46 or v1.9.47 ref, and nothing unreleased hiding behind the release list. So this is not urgent — the release train hasn't reached the colliding version. It just needs fixing before the next tag lands.

Both gates remain blind, as expected: `version-bump-gate / version-check` still passes (consistent with 1.9.47 being unreleased) and `chart content ⇒ Chart.yaml version bump` still passes. `mergeable: MERGEABLE` with `mergeStateStatus: BLOCKED` — blocked by review, not by a conflict, so git still auto-merges the identical two-line edit silently.

Still no test referencing `IMAGE_PULL_SECRET_NAME` — repo-wide code search returns 0 hits, and the 32-file `client/tests/` listing is unchanged on this branch.

Nothing new needed from me; @saadqbal's two inline findings (the hand-made `regcred` case at line 424, and the `.Values.env` passthrough exclusion at 437) are sharper than mine and both look right. The 424 one is worth carrying to the paired `client-runtime` PR too, since that's where the hardcoded name is being removed.

— drafted with Claude Code

client/templates/jobs-manager-deployment.yaml
LukasWodka: Right, and worth naming what that config actually was: those installs were **half** authenticated. Every chart-rendered pod got no `imagePullSecrets` at all — `useImagePullSecrets` is false under `create: false` — and only the *training* pod authenticated, because the literal was in `job.yaml`. So `regcred` was doing work for exactly one pod in the release, by accident.

Which is why I took your second suggestion rather than the note. The note documents the breakage without giving them anywhere to go: the reason to hand-make the Secret is usually that credentials must not live in values, so "set `create: true`" is not an answer for them. `dockerRegistry.existingSecret` is, and it also authenticates their control-plane pods — which never were.

Fixed in `8b68807`.

**The trap, and why there is now a second helper.** "A pull Secret exists" and "the chart renders one" had the same answer until this change, and `docker-registry-secret.yaml` was gated on the former. Left that way, the chart would overwrite the operator's hand-made Secret using credentials it does not have — a supported config turned into a broken pull on the first upgrade. `tracebloc.createRegistrySecret` is the create-gate; `useImagePullSecrets` stays the reference-gate. Mutation: revert the template to the old gate → `existingSecret: the chart renders NO Secret` reddens.

`create` + `existingSecret` together is refused at **both** layers, and I checked both fire:

```
$ helm template ... --set create=true --set existingSecret=regcred
Error: ... at '/dockerRegistry/create': value must be false     # the schema
$ ... --skip-schema-validation
Error: ... Those contradict: one asks the chart to build the Secret   # the template
```

The schema gives the better message and fires first; the template is the gate, because `--skip-schema-validation` exists.

**The limitation I could not close, documented rather than hidden.** With `existingSecret` the chart cannot mirror into `nodeAgents.namespace.name` or `gpu.devicePlugin.namespace` — it has no contents to mirror. `values.yaml` names both namespaces and the conditions under which they apply. I chose not to `fail` on it: `resourceMonitor` defaults true with a differing namespace, so failing would make the option unusable by default, and the symptom if an operator misses one is ImagePullBackOff — loud and diagnosable, unlike the failure the option prevents.

Rendered all five shapes:

| values | `IMAGE_PULL_SECRET_NAME` | Secrets the chart renders | refs |
|---|---|---|---|
| `create: true` | `myrel-regcred` | `myrel-regcred` in `tracebloc` **and** `tracebloc-node-agents` | `myrel-regcred` |
| `existingSecret: regcred` | `regcred` | **0** | `regcred` |
| neither | *absent* | 0 | none |
| `create: false` | *absent* | 0 | none |
| both | refused | — | — |

client/templates/jobs-manager-deployment.yaml
LukasWodka: Correct, and your point about the *consequence* is what decided the shape of the fix rather than just whether to make it. A slow pull is recoverable; a pull-secret name that does not resolve is treated as permanent and dead-letters the experiment, so a shadowed value costs the run. Fixed in `8b68807`.

Excluded from **both** containers' lists, not only the one where it could do harm — so a stray `env.IMAGE_PULL_SECRET_NAME` is inert rather than sitting unread in pods-monitor's environment. Nobody can depend on that: the variable does not exist before this PR. (`jobs_manager.py` is what reads `job.yaml`, and that runs in the api container, so pods-monitor never needed it.)

`JOB_IMAGE_HOST` I deliberately left, for the reason you gave — not a regression — plus one you could not have known: removing it would take away an override somebody may be relying on, which is a separate decision from adding a new var's exclusion. It now earns its keep as the **canary** for the general test below.

Because this class will recur, the fix is not just two `ne` clauses:

* `no container renders any env key twice` — over every Deployment / StatefulSet / DaemonSet / Job / CronJob / Pod in the render, not just jobs-manager. The next computed var added without an exclusion is caught here instead of in a review.
* `the duplicate-env check can actually see a duplicate` — asserts `JOB_IMAGE_HOST` still renders twice. Without a live duplicate, a parser that found no containers at all and a clean render look identical in a log, and the check above would be inert. If `JOB_IMAGE_HOST` is ever excluded, that test fails and tells the next person to pick a new canary.

Measured per container by parsing the rendered YAML, not grepping — the two containers have separate env lists and a grep cannot tell them apart, which is exactly why my first count of this variable read 2 when the api container had 1:

```
api                    IMAGE_PULL_SECRET_NAME x1 = ['regcred']   # computed value survives
pods-monitor-container IMAGE_PULL_SECRET_NAME x0 = []
```

Mutation: remove both `ne` clauses and the env-shadow test plus the general duplicate test both redden.

LukasWodka:
bugbot run

client/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.46
-appVersion: "1.9.46"
+version: 1.9.47
+appVersion: "1.9.47"
```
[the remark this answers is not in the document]
LukasWodka: Confirmed by measurement, and the gate was green the whole time. Fixed in `b99287b` — now `1.9.48`.

```
merge-base            1.9.46
this PR               1.9.47
origin/develop        1.9.47   <- #750, unrelated content
newest published tag  v1.9.44
```

`1.9.48` is unused on develop and untagged, checked both.

Your point about the guard is the more important half, so I filed it rather than leaving it in a thread: **backend#2193**. `chart-version-guard.sh` asks *"did chart content change without a bump against this PR's base"*, and by that rule it was right to pass. A collision is a different question — *"is this version already claimed by something that will be in the same branch"* — and a base-diff cannot express it, because both sides are self-consistent and the version did increase. Two PRs bumping `1.9.46 → 1.9.47` in parallel is the ordinary case; whichever merges second silently reuses a version.

The consequence you named is what makes it worth a ticket: auto-upgrade keys on the chart version, so the second PR's content can land in the repo under a version Helm already considers applied and never reach a cluster. Here that content was the training-pod credential fix itself.

The ticket asks for the comparison against the merge target's *current* version plus the published-tag list, failing closed on an unreadable Chart.yaml either side, and requires the mutation to be proved on this exact input — base 1.9.46, PR 1.9.47, target 1.9.47 — since today's guard is green on it.

client/templates/_helpers.tpl
```diff
@@ -137,8 +137,22 @@ mysql-pvc
...
+  jobs-manager stamps onto training pods both read THIS, so the name cannot
+  disagree between the pod that creates the Secret and the pod that uses it.
+*/}}
 {{- define "tracebloc.registrySecretName" -}}
+{{- $reg := .Values.dockerRegistry | default dict -}}
+{{- if $reg.existingSecret -}}
+{{ $reg.existingSecret }}
+{{- else -}}
 {{ .Release.Name }}-regcred
+{{- end -}}
```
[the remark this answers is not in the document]
LukasWodka: Correct, and the second half is the direction I would not have found. Fixed in `b99287b`.

You are right that `registrySecretName` reached a place it should not: that Role is in `kube-system` by default, and pinning `resourceNames` through the release-wide helper handed the auto-upgrade ServiceAccount **get + delete on the operator's own dockerconfigjson** — rights over a Secret this chart does not own, granted purely as a side effect of naming one.

`tracebloc.createdRegistrySecretName` returns `<release>-regcred` regardless of `existingSecret`, and RBAC uses that. It is one more consumer of the split the Secret template already forced: *"does the chart own it"* is a different question from *"what does the release reference"*, and I had only applied it to the Secret template.

The migration case you flagged is why naming the created secret is better than dropping the rule: switching a live release from `create: true` to `existingSecret` orphans the mirrored `<release>-regcred`, and Helm must still be able to delete it. Gating the rule off entirely would have traded your finding for a stalled auto-upgrade. Both directions are now tests:

| test | asserts |
|---|---|
| `no Role grants rights on the operator's own Secret` | no `resourceNames` entry equals `regcred`, with the device plugin enabled |
| `the chart's own former mirror stays deletable` | some rule still has `delete` on `myrel-regcred` |
| `the RBAC name still matches the Secret that is rendered` | on `create: true`, unchanged — the split must not desynchronise the ordinary path |

The parser asserts it found `resourceNames` at all before judging them; without that, a Role-walker reaching nothing is indistinguishable from a chart with no offending grant.

Mutation: point RBAC back at `registrySecretName` and tests 12 and 13 both redden.

**One thing I deliberately did not change**, since it is pre-existing rather than introduced here: the same Role holds `create` on secrets in `kube-system` with no `resourceNames` — RBAC cannot scope `create` by name, so that grant is unrestricted today regardless of this PR. Say the word and I will file it; I did not want to widen a review round into an RBAC audit without asking.

LukasWodka:
bugbot run

scripts/tests/chart-pull-secret.bats
```diff
@@ -0,0 +1,323 @@
...
+NS="tracebloc"
+
+setup() {
+  # Set here rather than in setup_file: exports from setup_file do not reach the
+  # test body in every bats version, and an empty $CHART makes helm fail with
+  # "non-absolute URLs" — eleven confusing reds instead of one clear skip.
+  CHART="${BATS_TEST_DIRNAME}/../../client"
+  VALUES="${CHART}/ci/bm-values.yaml"
+  command -v helm >/dev/null || skip "helm not installed"
+  command -v python3 >/dev/null || skip "python3 not installed"
```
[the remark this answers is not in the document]
LukasWodka: Confirmed, and this is the sharpest finding on the PR — the fourteen tests I added to close an inert-verification gap were themselves inert. Fixed in `f03be28`.

Measured rather than assumed: no `setup-helm` anywhere in `standard-checks.yml`, and `scripts/tests/*.bats` is run by exactly one job. So every case hit `skip "helm not installed"` and a wrong `IMAGE_PULL_SECRET_NAME` or Secret name would have shipped green. That job — `Unit tests` — **is** required on develop, which makes it worse rather than better: a skip in a required gate is indistinguishable from a pass.

Two fixes, because arming the job alone would leave the trap for the next test:

1. **helm in the `Unit tests` job** — same sha-pinned `azure/setup-helm` and same `v3.15.4` pin as `helm-ci.yaml` and `drift-checks.yaml`, for the reason helm-ci already documents: helm is absent from the runner image and an unpinned one silently changes which assertions run. I also re-did the job's timeout arithmetic in its own comment rather than leaving it stale.
2. **a missing tool now FAILS in CI and skips only locally.** `require_tool` reads `CI`, which Actions sets to `true`. Proved both directions with a PATH that hides helm but keeps bats:

   | environment | result |
   |---|---|
   | `CI=true`, no helm | **14 reds** |
   | unset `CI`, no helm | 14 skips, 0 reds |
   | helm present | 14 passes |

You are right about the version gate too: `--skip-schema-validation` landed in **3.16** and CI pins **v3.15.4**. That case now self-skips below 3.16 and names the version it saw — the same treatment and the same reason as the helper-backstop cases in `chart-env-vocabulary.sh`. Worth noting the pair is not left unguarded: the **schema** half runs on every version, so the contradiction is always covered; only the second layer is version-bound, and bumping the pin turns it on with no change to the test.

client/templates/jobs-manager-deployment.yaml
LukasWodka: Addressed in `8b68807` — `dockerRegistry.existingSecret`, your second option rather than the note. Full reasoning in the reply above; the short version is that those installs were **half** authenticated (only the training pod, off the literal), so the note would have documented a breakage with nowhere to go, since the reason to hand-make the Secret is usually that credentials must not live in values.

Resolving as fixed. The follow-up findings it produced — the `kube-system` RBAC name and the version collision — are answered on their own threads.

client/templates/jobs-manager-deployment.yaml
LukasWodka: Addressed in `8b68807` — excluded from both containers' lists, plus a general `no container renders any env key twice` test over every workload with `JOB_IMAGE_HOST` as the live canary so that test cannot go inert. Full reasoning in the reply above.

Resolving as fixed.

LukasWodka:
bugbot run

scripts/tests/chart-pull-secret.bats
```diff
@@ -0,0 +1,354 @@
...
+@test "create plus existingSecret is refused by the values schema" {
+  run helm template myrel "$CHART" -f "$VALUES" --namespace "$NS" \
+    --set dockerRegistry.create=true \
+    --set dockerRegistry.server=https://index.docker.io/v1/ \
+    --set dockerRegistry.username=u --set dockerRegistry.password=p \
+    --set dockerRegistry.email=e@x.io \
+    --set dockerRegistry.existingSecret=regcred
+  [ "$status" -ne 0 ] || return 1
+  [[ "$output" == *"must be false"* ]] || return 1
+}
```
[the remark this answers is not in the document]
LukasWodka: Right about the consequence, wrong about the cause — and CI proved the consequence before I could argue either way. Fixed in `4484ff9`.

I installed the pinned 3.15.4 and read both outputs rather than trusting the diagnosis or my own guess:

```
3.15.4  - dockerRegistry.create: dockerRegistry.create does not match: false
4.1.1   - at '/dockerRegistry/create': value must be false
```

So 3.15.4 does not print `Must be false` — it prints `does not match: false`. This was never a case-sensitivity bug; the per-rule wording changes wholesale between schema-validator generations, which is a stronger reason not to assert on it.

Anchored on the header line **both** print — `meet the specifications of the schema` — which is also the better assertion: it proves the **schema** layer refused rather than the template, and that distinction is the whole point of this test, since its sibling covers the template layer. `create` and `false` are then matched case-insensitively to pin which rule fired, so an unrelated schema failure (a missing required value, say) cannot satisfy it.

Verified on both:

| helm | ok | not ok | skipped |
|---|---|---|---|
| v4.1.1 | 14 | 0 | 0 |
| v3.15.4 | 14 | 0 | 1 — `--skip-schema-validation`, as designed |

Worth noting what this sequence actually demonstrates: your previous finding was that these fourteen tests never ran. One commit after arming them, the arming caught a real defect in my own assertion. That is the difference between a test suite and a test suite that is connected to anything.

LukasWodka:
bugbot run

saadqbal (dismissed):

LukasWodka:
@saadqbal — you asked for the helm-unittest coverage in your review **body**, and I had only acted on your two line threads. That is squarely the failure mode of treating "0 unresolved threads" as "review addressed", so thank you for putting it in writing where re-reading caught it. Added in `d0cf519`.

Eight cases in `client/tests/image_pull_secret_test.yaml`, following `global_image_registry_test.yaml`'s two-case shape as you suggested. The two invariants are the ones the original defect broke in each direction:

1. **a pull Secret exists** → the var is injected **and names the Secret the pod actually pulls with**. Asserted against the pod's own `imagePullSecrets` and against the rendered Secret's `metadata.name`, so the two cannot drift apart — the broken state was a plausible string that resolved to nothing, which an equality-against-a-literal test would have happily accepted.
2. **no registry configured** → the var is **absent, not empty**. client-runtime reads absence as "pull publicly", so an empty value would make it set a pull secret nothing created: the same defect inverted.

`existingSecret` is covered too, since that is the path where the name is not derivable from the release name at all — the case a literal in `job.yaml` could never have got right — plus the assertion that the chart renders **no** Secret there, so it cannot overwrite yours.

Mutation-proved, because a green suite that cannot fail is the thing this coverage is for:

| mutation | result |
|---|---|
| delete the injecting `{{- if }}` block | 2 failed |
| render the var unconditionally (empty on a public pull) | 2 failed |
| drop the release prefix from the name helper | 3 failed |
| gate the Secret template on `useImagePullSecrets` | 1 failed |

Full suite **482 passed / 32 suites** — 467 before this file, matching your count. helm lint clean, `chart-version-guard` 23/23, `chart-pull-secret` 14/14.

Everything else from your review is in: the version is `1.9.48` with `develop` merged (your finding — and the conflict that produced is itself the evidence, since the identical 1.9.47 on both sides merged silently), and `IMAGE_PULL_SECRET_NAME` is excluded from both passthrough lists. The guard blind spot you identified is filed as backend#2193.

Arturo's two follow-ups on the same change are also addressed: the `kube-system` RBAC no longer names a bring-your-own Secret, and Bugbot's catch that the render tests were **silently skipped** in the required `Unit tests` job (no helm installed) is fixed — helm is now installed there, and a missing tool fails in CI rather than skipping.

LukasWodka:
bugbot run

aptracebloc (approved):

PR #752, opened 2026-08-19
fix(installer): split existing-release from existing-datadir in the arch gate

LukasWodka:
Resolves the three unresolved threads on the \`client\` develop→staging mirror (#748) — 1 Bugbot **Medium** + 2 from @saadqbal — on \`scripts/lib/install-client-helm.sh\`.

## The bug
`existing_id` (a **live Helm release**, from `detect_installed_client` → `helm list -A`) was folded into the `existing-datadir` reason. On arm64 the arch gate then:
- claimed **"This host holds existing MySQL 5.7 data"** — which can be false: on `TB_STORAGE_MODE=node-local` both `_mysql_dir_has_content` probes are empty, so `existing_id` is the only trigger and there may be no host datadir at all; and the release could even be an 8.4 one whose sticky marker was lost;
- offered **`--data-dir`** as a fresh start — but that cannot clear a release `helm list` reports, so the next run refuses identically.

## The fix
Split the trigger into two reasons, same 5.7 engine choice, **different remedy**:
- `existing-release` (from `existing_id`): 5.7 kept as a data-safety default, **no** host-data claim, remedy is emulation **or uninstalling the release** — not `--data-dir`.
- `existing-datadir` (real host files): unchanged; `--data-dir` is a valid fresh start there.

Also corrected the header comment that implied `assert_amd64_emulation` fully "owns" macOS — it refuses whenever Rosetta amd64 is off **without** consulting `_mysql_engine_decision`, so a fresh arm64 Mac that 8.4 would serve natively is still turned away. Named as a separate **backend#2047 follow-up** rather than silently implied covered (@saadqbal’s second point).

## Tests
- `existing-release` distinguished in the reason-distinguishability and purity suites.
- New arch-gate test pins the corrected remedy: no false host-data claim, no `--data-dir`.
- Both halves **mutation-proved**; **157/157** bats green; `shellcheck -x` clean.

On `develop` so the staging hop re-prepare picks it up.

---

> [!NOTE]
> **Low Risk**
> Installer messaging and engine-reason ordering only; behavior still pins 5.7 for safety, with expanded tests and manifest hash update.
> 
> **Overview**
> Fixes misleading **arm64 Linux** installer errors when MySQL stays on **5.7** only because a **Helm release** exists, not because host datadir files were found.
> 
> **`_mysql_engine_decision`** now returns **`5.7 existing-release`** when `existing_id` is set and host mysql dirs are empty, and **`5.7 existing-datadir`** only when `_mysql_dir_has_content` fires; **datadir is checked first** so release + files still report datadir.
> 
> **`_assert_engine_runs_on_this_arch`** adds an **`existing-release`** branch: it does **not** claim the host holds 5.7 data or suggest **`--data-dir`**; fresh-start guidance covers **uninstall plus retained MySQL PVC** (`helm.sh/resource-policy: keep`). Comments note **macOS** Rosetta gating is still a separate follow-up.
> 
> Bats cover the new reason, precedence, and arch-gate messaging; **`scripts/manifest.sha256`** is updated for `install-client-helm.sh`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d84f90f51bcbb663bd458f34e3fab969075e513f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/lib/install-client-helm.sh
```diff
@@ -862,7 +862,8 @@ _client_default_namespace() { _sanitize_workspace_name "${TB_NAMESPACE:-traceblo
 #   5.7|8.4 explicit      an explicit TB_MYSQL_ENGINE=<value>
 #   invalid <value>        TB_MYSQL_ENGINE is not auto|5.7|8.4
 #   8.4 sticky             this machine already opted into 8.4
-#   5.7 existing-datadir   an existing release, or real mysql datadir content
+#   5.7 existing-release   a live Helm release (helm list), data format unknown
+#   5.7 existing-datadir   real 5.7-format mysql datadir content on this host
```
saadqbal: nit: `real 5.7-format` overstates the probe — `_mysql_dir_has_content` only tests non-emptiness, it never looks at format. That's the same unbacked claim this PR is removing from `existing_id`. The old wording ("real mysql datadir content") was accurate; if you do want the format claim, `mysql.ibd` is the discriminator the chart's format guard already uses.

saadqbal (changes requested):
The four-quadrant split itself reads correctly: datadir-first, then `existing_id`, keeps the engine at 5.7 in all three "something exists" quadrants exactly as the old OR did, so nothing new becomes reachable engine-wise — only the reason string and the remedy change. Both probes fail closed (`INSTALLED_CLIENT_UNKNOWN` hard-errors before `_resolve_mysql_engine`; unlistable datadir counts as content), and `TB_NAMESPACE` is switched to `existing_ns` before the decision runs, so the per-release probe looks at the right path. `manifest.sha256` matches the new blob (`d0a72ba7…`). No values.yaml or template change, so no `--reuse-values` nil-guard exposure.

Holding this on Bugbot's still-open thread rather than restating it inline: the `existing-release` "just uninstall the release" remedy is wrong in the one case the reason exists for. `node-local` is exactly where both host probes are empty, and the mysql PVC carries `"helm.sh/resource-policy": keep` (`client/templates/mysql-storage-pvc.yaml`, outside the `hostPath.enabled` block), so `helm uninstall` keeps the 5.7 datadir. The next run then sees no release and empty host dirs, resolves `8.4 fresh`, and the format guard in `mysql-deployment.yaml` refuses to start — after the operator has torn down a working install on the installer's own advice. What actually clears in-node data is `k3d cluster delete` or deleting the PVC. The comment at :903 asserting "no files remain to re-trigger the gate" needs the same correction.

Two smaller ones, both outside this diff so no inline: `_pf_arch`'s `existing-datadir` hint in `preflight.sh` never got this PR's "uninstall any existing release" correction, and preflight hard-fails before the corrected copy here can print — so the datadir case still sees the old `--data-dir`-only advice. And `_resolve_mysql_engine`'s `case "$reason"` still narrates only explicit/sticky/fresh, so `existing-release` (like `existing-datadir`) pins 5.7 with nothing in the log saying why — invisible on amd64, where the gate never fires.

Test coverage of the four quadrants is good, and adding `existing-release` to the mutation loop is the right instinct.

LukasWodka:
Fixed — you're right, `_mysql_dir_has_content` only tests non-emptiness, so "5.7-format datadir content" was the same unbacked claim this PR strips from the `existing_id` path. Reverted the datadir reason to format-neutral wording ("real mysql datadir content, non-empty"); the engine still keeps 5.7 precisely because the format is *unknown* and 8.4 cannot open a 5.7 dir if that is what it is. Also softened the one "that 5.7-format data" in the existing-release remedy to "may be". Pushed.

scripts/lib/install-client-helm.sh
```diff
@@ -885,15 +886,34 @@ _mysql_engine_decision() {
...
+  #     there IS data here and a clean start means clearing the data dir (a
+  #     release-only "uninstall" remedy would leave those files to re-pin 5.7 on
+  #     the next run). An UNLISTABLE dir counts as content (fail closed; see the
+  #     helper).
+  #
+  #   existing_id — a live Helm release from detect_installed_client (helm list
+  #     -A + clientId) with NO host datadir files, e.g. TB_STORAGE_MODE=node-local
+  #     where both probes are empty. Reason 'existing-release': 5.7 stays the safe
+  #     default, but the gate must not claim host 5.7 data — and "uninstall the
+  #     release" is a COMPLETE fresh-start remedy here precisely because no files
```
saadqbal: nit: this rationale now says the opposite of the hint you just fixed — "uninstall the release" is a COMPLETE fresh-start remedy here precisely because no files remain. :980 now correctly says `helm uninstall` alone is not enough because the PVC is kept. Worth correcting here too: this comment is the argument a future maintainer would reach for to undo the fix.

scripts/lib/install-client-helm.sh
```diff
@@ -929,30 +949,42 @@ _resolve_mysql_engine() {
...
       error "MySQL 5.7 was requested explicitly and cannot run on ${ARCH} without amd64 emulation." ;;
+    existing-release)
+      warn "An existing tracebloc release is installed here, so the install keeps the MySQL 5.7 engine as a data-safety default — and that image is amd64-only."
+      hint "Whether this release's data is actually 5.7-format cannot be told from here: the release is detected via 'helm list', not the data directory. 5.7 is kept because 8.4 cannot open a 5.7-format datadir if one exists."
+      hint "Keep the existing release — enable amd64 emulation, then re-run:"
+      hint "  docker run --privileged --rm tonistiigi/binfmt --install amd64"
+      hint "Or start fresh on the native 8.4 engine — this needs BOTH the release AND its retained data removed. 'helm uninstall' alone is not enough: the MySQL volume is annotated 'helm.sh/resource-policy: keep', so it survives the uninstall, and 8.4 cannot open a 5.7-format datadir, which that retained volume may be (the next run resolves to 8.4 and fails the format guard after cluster setup). Delete the retained MySQL PVC (and any host data directory) as well before re-running."
```
saadqbal: nit: every other hint in this function ends in something copy-pasteable; this one leaves the operator to work out the PVC name and namespace themselves. The one-client guard at :1284 already offers `k3d cluster delete ${CLUSTER_NAME:-tracebloc}` for exactly this ("wipes this client + its local data") — reusing it would make the remedy actionable rather than descriptive.

saadqbal (approved):
Verified the manifest by hand rather than trusting the resolved thread: recomputed all 18 digests in `scripts/manifest.sha256` against the `d84f90f` blobs — every one matches, and no fetched lib under `scripts/lib` is missing from the manifest. The signed fetch is intact at this ref.

The remedy now covers both storage modes. `node-local` is named explicitly (retained PVC under `resource-policy: keep`), and `hostpath` cannot reach `existing-release` at all — `HOST_DATA_DIR` is bind-mounted at `/tracebloc`, so the per-release probe sees those files and the reason is `existing-datadir`, whose remedy already says to uninstall the release and point at an empty data dir. The annotation is outside the `hostPath.enabled` conditional in `mysql-storage-pvc.yaml`, so the claim in the new hint holds in both modes.

The invariant from my last pass still holds: the decision block itself (:908-920) is untouched by these two commits, so the engine is 5.7 in all three "something exists" quadrants exactly as the old OR was, and 8.4 is still never selected where the old condition selected 5.7. `_mysql_dir_has_content` still fails closed on unlistable dirs, `INSTALLED_CLIENT_UNKNOWN` still hard-errors before `_resolve_mysql_engine`, and the PR still touches no values.yaml or template, so the `--reuse-values` nil-guard rule stays N/A.

Tests: 137 of 157 ran green (0 failures) before I capped the run locally — the whole engine and arch-gate block passes, including the updated `existing-release` assertions.

Two nits inline, neither blocking. The `preflight.sh` `existing-datadir` hint and the missing `existing-release` log line from my last pass are both still outstanding and still outside this diff — follow-up material, not this PR.

scripts/lib/install-client-helm.sh
saadqbal: Fixed in d84f90f — now reads "real mysql datadir content on this host (non-empty)", and the block comment says the probe tests content, not format. Resolving.

LukasWodka:
/fr-pass

Issue #753, opened 2026-08-19
ci: package installs have no timeout or retry — an apt stall burns the job budget and reports as 'cancelled'


LukasWodka:
## What happened

On 2026-08-19 from ~07:35 UTC, every CI job in this repo that installs a package stalled indefinitely:

| workflow → job | stuck on | duration |
|---|---|---|
| Standard checks → Lint | `Install shellcheck` | 64 min+, still running |
| Standard checks → Unit tests | `Install bats` | 64 min+, still running |
| Installer tests → Static analysis | `sudo apt-get install -y -qq shellcheck` | hit `timeout-minutes: 10`, **twice** |

Affected `develop`, `feat/1907-installer-outcome-telemetry` and `fix/runtime-down-start-docker` alike, so it is not branch content. githubstatus.com reported all components operational throughout — this is the Ubuntu/Azure apt path, which their status page does not cover.

## The durable defect, independent of today

The stall is transient. **The way our steps handle it is not.**

`.github/workflows/installer-tests.yaml`'s Static analysis step opens with:

```
sudo apt-get update -qq && sudo apt-get install -y -qq shellcheck
```

No timeout, no retry, and `-qq` means it emits nothing while it hangs. So a slow mirror:

1. burns the entire `timeout-minutes: 10` job budget before a single test runs;
2. reports as **`cancelled`**, which reads like a human pressed cancel rather than "the toolchain never installed";
3. produces a log with zero output, so diagnosing it means comparing step timestamps to work out that `shellcheck --version` never printed.

I lost two cycles to (2) and (3) today, and re-ran once on a wrong diagnosis of "transient mirror" before three simultaneous hangs made the real cause obvious.

This is the same class as #744 — an unbounded external call that hangs instead of failing — one layer out, in CI rather than the installer.

## Proposal

- **Bound and retry the installs.** `timeout 120 sudo apt-get …` with 2-3 attempts and a backoff. A failed install should fail *fast and loudly*, not consume the job budget.
- **Drop `-qq` on the install**, or echo a marker before and after. A step that hangs silently is a step nobody can diagnose from the log alone.
- **Prefer not installing at all where possible.** `shellcheck` and `bats` are stable, pinned tools; a cached binary (`actions/cache`) or a pinned release download removes the apt dependency from the hot path entirely. Note `Installer tests` and `Standard checks` each install shellcheck separately, so the fix wants to be shared rather than pasted twice.
[...]

PR #754, opened 2026-08-19
ci: one shellcheck gate, and it is the one that does not apt-install (#753)

LukasWodka:
Closes the shellcheck half of #753.

## The measurement

shellcheck runs in **three** places in this repo. On 2026-08-19 two of them hung 35+ minutes on `apt-get install shellcheck`; the third passed green in ~5 seconds.

| # | Mechanism | Required? | Installs shellcheck? | On 2026-08-19 |
|---|---|---|---|---|
| 1 | `quality / shellcheck` (org `code-quality.yml`) | ✅ yes | no — preinstalled | green, ~5 s |
| 2 | `Lint` (`standard-checks.yml` → `make lint`) | ✅ yes | `apt-get` | hung 35+ min, **no `timeout-minutes` at all** |
| 3 | `Static analysis` (`installer-tests.yaml`) | ❌ no | `apt-get` | killed at `timeout-minutes: 10`, twice, reported `cancelled` |

The comment above #2's install said *"shellcheck itself is not on the runner image, so it is installed here."* That is false. Run `32232682350` prints `version: 0.9.0` 3.9 s into the job with no install step.

## The file sets, measured before deleting anything

The ticket asks whether #2 and #3 add coverage over #1. They do not — they subtract it.

- #2's `SHELLCHECK_FILES` and #3's inline list are **byte-identical to each other**: 19 entries, expanding to **34 files**.
- #1 **derives** its set (extension, else shebang): **42 files** at the time of measurement, **44** after merging develop's #747 — which had to hand-edit the enumerated list in three places to add one script, while the derivation picked it up with no edit at all.
- The enumeration is a **strict subset**. `comm` reports zero files it catches that the derivation misses.

The eight it had drifted past:

```
docker/k3s-cuda/build.sh          docker/k3s-cuda/k3d-entrypoint-tracebloc-cdi.sh
docs/migration-tools/generate.sh  docs/migration-tools/migrate-tenant.sh
scripts/chart-version-guard.sh    scripts/check-digest-drift.sh
scripts/index-invariants.sh       scripts/tests/test_helper.bash
```

Mutation-proven, anchor confirmed applied:

```
planted a parse error in scripts/index-invariants.sh
  enumerated set (34) -> exit 0   <- invisible to today's gate
  derived set    (42) -> exit 1   SC1073 / SC1072
restored
  derived set    (42) -> exit 0   <- anchor applied
```

Same severity (`error`) and same `--exclude=SC1091` on all three, so no verdict changes for the files they share.

## Scope, stated plainly (corrected after review)

The two claims below are separate, and the first draft of this PR conflated them:

- **Set-wise**, the enumerated list is a strict subset of the derived set. Still true.
- **Scope-wise**, the deleted jobs swept their whole list on *every* run, while `quality / shellcheck` reads only the **PR diff** — its caller passes `all-files: false` and declares no `schedule:`. Demonstrated on this very PR: that job is green in 8s with `Shell files to check: 0`.

As first written, this PR would therefore have left **no** CI job doing a whole-tree error-severity sweep. Fixed in 80671e8 rather than merely documented: **`Lint` runs `make lint`** (parse + derived whole-tree shellcheck), which needs no install because shellcheck is preinstalled. The result is two complementary required gates — whole-tree in `Lint`, diff-scoped in `quality / shellcheck` — neither installing anything and neither holding a file list. Verified green on the 44-file set under the runner's own shellcheck **0.9.0**, not just a newer local build.

## The change

- **`installer-tests.yaml`** — `static` loses its ShellCheck step **only**. R8 manifest, check-facts and check-style are untouched. This matters: `Static analysis` is **not** a required context and R8 has no other home, so this change must not be read as thinning that job. Making it required is the companion ticket, filed separately.
- **`standard-checks.yml`** — `Lint` loses the apt install, calls `make parse`. Still a required context reporting unconditionally, as its own header demands. `timeout-minutes` added to **both** jobs.
- **`Makefile`** — keeps shellcheck for the pre-push tier but **stops enumerating**, deriving the set with the org job's rule and flags instead. backend#1606 found this list drifted once already and fixed it by re-copying it, which reset the clock rather than stopping it.
- **`helm-ci.yaml`** — a comment citing the shellcheck apt install as its precedent; that precedent no longer exists and was never sound.

## Fail-closed properties, both mutation-proven

The old one-liner had neither:

1. **Classifying zero files now fails.** A broken derivation and a clean tree look identical otherwise. Proven against a fixture repo with no shell files → exit 2, not green.
2. **The classifier's exit status is not swallowed by a pipe.** Recipes run under `/bin/sh` = dash on Debian, which has no `pipefail` — the exact #630 hazard the Makefile already warns about. The list is materialised first.

## Verification

- `make check` green under **bash and dash**, 9.9 s against a 60 s budget.
- `make parse`, `make shellcheck` green under both shells; derived set byte-identical to an independently computed one (42 files).
- `actionlint` clean on all three touched files. (It reports pre-existing info-level findings in `release-helm-chart.yaml` / `windows-e2e.yaml`, untouched here — that is backend#1776.)
- No `scripts/` file changed, so `scripts/manifest.sha256` needs no regeneration; `gen-manifest.sh --check` passes as part of `make check`.

## Deliberately not in scope

- **The bats install.** Unlike shellcheck it genuinely has to be installed, so bounding and retrying package installs stays #753's own scope. The new `timeout-minutes: 15` at least turns a stall into a bounded failure instead of a 6-hour one.
- **`quality / gitleaks`**, tracked separately.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to CI workflows and Makefile lint orchestration; behavior shifts toward broader shellcheck coverage with no application runtime impact.
> 
> **Overview**
> **Consolidates shellcheck** so CI stops `apt-get install shellcheck` (preinstalled on runners; those installs hung jobs for 35+ minutes). The duplicate, non-required shellcheck step is removed from installer **Static analysis**; required **Lint** runs `make lint` only.
> 
> **`make lint`** now means **parse + derived whole-tree shellcheck**: the old enumerated `SHELLCHECK_FILES` list is gone in favor of the same classification rule as org `quality / shellcheck` (extensions/shebangs, error severity, SC1091 excluded). That pairs with the diff-scoped org check so every push still gets a full-tree sweep without maintaining three file lists.
> 
> **Fail-closed lint**: `parse` and `shellcheck` materialize file lists, refuse zero files, and propagate real exit codes under dash (fixes fail-open `find | xargs` behavior).
> 
> **CI resilience**: job `timeout-minutes` on standard checks; **bats** install is bounded, retried, and uses `sudo timeout` in standard-checks and installer **unit-bash** (15-minute job budget). **helm-ci** comment updated to drop the invalid shellcheck-install precedent.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d811b3939689d5d8c0dc9b8f05b18890cd6241f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


LukasWodka:
Companion ticket for the sibling finding is now filed: tracebloc/backend#2182 — the R8 manifest check, `check-facts.sh` and `check-style.sh` live **only** in `Static analysis`, which is not a required context, so they can be red and still merge.

This PR deliberately leaves that job's steps intact so the gap is not widened. Confirmed while writing it that the required `Source-of-truth drift` job does **not** cover them — it runs `check-drift.sh` and the CLIENT_ENV vocabulary guard only.

LukasWodka:
bugbot run

Makefile
```diff
@@ -157,12 +187,52 @@ install-hooks:
...
+#   * the classifier's exit code is not swallowed by the pipe. Recipes run
+#     under /bin/sh -- dash on Debian, no pipefail -- so the file list is
+#     materialised first and shellcheck's own status is what propagates.
+# The `; :` inside the classifier is load-bearing: `grep -q ... && printf`
+# exits 1 on every non-shell file, which would otherwise make xargs return 123.
+.PHONY: shellcheck
+shellcheck:
+	@files=$$(mktemp); \
+	git ls-files -z \
+	  | xargs -0 -r -n1 sh -c 'case "$$1" in \
```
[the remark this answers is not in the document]
LukasWodka: Refuted — measured on the machine this PR was developed on (Darwin 25.5.0, stock `/usr/bin/xargs`, no brew findutils on PATH).

BSD `xargs` supports `-r`, and its own man page says so explicitly:

```
-r, --no-run-if-empty
        Compatibility with GNU xargs.  The GNU version of xargs runs the
        utility argument at least once, even if xargs input is empty, and
        it supports a -r option to inhibit this behavior.
```

Direct check that it is BSD and that it accepts the flag:

```
$ which xargs
/usr/bin/xargs
$ xargs --version
xargs: unrecognized option `--version'      <- BSD, not GNU
$ printf '' | /usr/bin/xargs -0 -r echo NOPE
$ echo $?
0
```

And `make check` / `make shellcheck` ran green on this Mac repeatedly while developing the change (`shellcheck: 42 file(s), severity=error`, ~5 s), which it could not have done if the flag were rejected.

On the second half — "the zero-file check already makes `-r` unnecessary" — `-r` is not redundant, it is what keeps the *good* failure mode. Drop it and, on GNU `xargs`, an empty list invokes `shellcheck` with no file arguments:

```
$ shellcheck --severity=error
No files specified.
```

so the target would fail with shellcheck's generic usage error instead of the explicit "classified ZERO shell files -- the derivation above is broken" message the guard is there to print. Both layers are deliberate: `-r` stops the empty invocation, the counter explains *why* it was empty.

Keeping as-is. Resolving.

LukasWodka:
Note for review: this branch is 3 commits behind `develop` (#752, #741, #743). No rebase needed and none intended —

- **No overlap.** Those three touch `scripts/lib/*.sh`, `scripts/tests/*.bats|*.ps1` and `scripts/manifest.sha256`; this PR touches only three workflow files and the `Makefile`. Disjoint file sets, so no textual conflict.
- **Verified against the post-merge tree, not just this branch.** Checked out `develop`'s tip, dropped this PR's `Makefile` on top, and ran the widened derived set: `shellcheck: 42 file(s), severity=error` → exit 0, and `make parse` → exit 0, both under dash. So the eight newly-covered scripts stay green against the shell code those three PRs just changed.
- `strict: false` is deliberate fleet-wide (backend#1276), so being behind is not a merge blocker here.

Re-verifying rather than assuming was the point: #741 and #752 both edited `scripts/lib/*.sh`, which is exactly the surface this change widens coverage over.

aptracebloc:
Findings from a review pass. No verdict attached. The main one is about this PR's interaction with #755.

## #754 × #755 — the dangerous breakage merges *without* a conflict marker

Verified by running a real 3-way `git merge-file` from the merge base (`8de5d64`) against both heads. There are visible conflicts in all three shared files:

| File | Conflicts | Where |
|---|---|---|
| `Makefile` | 1 | the `check` timing comment only (`~10 s` vs `~5 s`) |
| `standard-checks.yml` | 1 | the `Lint` install-step block |
| `installer-tests.yaml` | 2 | the header bullet list, and the `static` job body — #754 edits it, #755 deletes it |

The problem is what merges **cleanly**. #754 deletes `SHELLCHECK_FILES`; #755 adds a `lint-warnings` target that *uses* `$(SHELLCHECK_FILES)`. Those edits sit at different offsets, so git merges them with no marker at all.

I built the merged Makefile, resolved only the visible conflict (a comment about seconds), and ran it: `make lint-warnings` invokes `shellcheck --severity=warning --shell=bash` with **zero file operands** and the target still exits **0**, because `|| true` swallows shellcheck's "No files specified." The advisory SC2034 sweep #755 went to the trouble of preserving becomes a permanent no-op that reports green — and the only warning of it is a conflict about a benchmark number. This is precisely the "classifying ZERO files is a FAILURE, not a green run" mode #754's own new `shellcheck` target is written to prevent.

The wrong order (#755 first) is worse still: #754's diff replaces the whole `Lint` step region, and with `static` already deleted, a resolution taking #754's side silently removes the **PSScriptAnalyzer** step — un-gating the PowerShell lint of `scripts/install.ps1`, the R8 Windows trust root.

**They're also contradictory in principle, not merely overlapping.** #754's thesis is "shellcheck runs in exactly one place, `quality / shellcheck`". #755 re-homes shellcheck *into* `Lint` via `make lint` + `make lint-warnings`, keeps the enumerated `SHELLCHECK_FILES` list this PR's entire argument identifies as the defect, keeps `--shell=bash`, and adds a conditional apt install back. #754 files the sibling gap as backend#2182 ("make `Static analysis` required"); #755 is the opposite remedy for the same gap. Someone has to pick one.

**Suggestion:** land #754 first, then rebase #755 with `lint-warnings` rewritten onto the derived set, `make lint` → `make parse`, and `Ensure shellcheck` dropped. In either order the second PR must be rebased and re-reviewed, never resolved through GitHub's conflict editor — the real breakage is outside the markers.

## Two findings on #754 itself

### 1. `Makefile:198-200` — `make parse` fails open

`@find scripts -type f -name '*.sh' -print0 | xargs -0 -n1 bash -n` has no `-r`, no count guard, and `find`'s status is lost across the pipe (no `pipefail` under dash). Verified: in a tree with no `scripts/` dir, `make parse` prints `find: scripts: No such file or directory`, then `all shell scripts parse`, and exits **0** — under both dash and bash.

This PR promotes `parse` to be the *entire* content of the required `Lint` check. Previously the enumerated shellcheck line right after it would have exited non-zero on the same breakage. It's the same fail-open the new `shellcheck` target is hardened against three lines below.

### 2. The "strict subset" argument is set-wise but not scope-wise

The org job (`tracebloc/.github` `code-quality.yml:499-503`) checks `git diff --name-only --diff-filter=ACMR "$BASE_SHA...HEAD"` unless `all-files: true`, and `code-quality-caller.yml` triggers only on `pull_request` with `all-files: false` — no schedule. The removed sweeps ran their 34 files on **every** PR; the survivor runs only on the diff and exits 0 on an empty set.

Demonstrated on this very PR: `quality / shellcheck` is green in 8s and its log reads `Shell files to check: 0`. Net effect — after this lands, no CI job performs a whole-tree error-severity shellcheck; it survives only in the pre-push hook and manual `workflow_dispatch`. The Makefile comment admits the diff-scoping honestly; the PR body doesn't.

## Verified and clean

The recipes really are dash-clean — ran them under `SHELL=/bin/dash` in throwaway git repos: no `[[`, arrays, `pipefail`, or process substitution, `$$` escaping and the backslash-continued `sh -c` parse correctly, and classification is right (picks `.sh`, `lib/*.sh`, shebang-only extensionless; skips `.bats`, `.ps1`, `.md`). Exit-status propagation is real, not just claimed: stub `shellcheck` exiting 1 → `make: *** Error 1`; no shellcheck on PATH → `Error 127`. The zero-classified-files guard genuinely exits non-zero. `timeout-minutes` added to both jobs (10 / 15). Job names `Lint` / `Unit tests` / `Static analysis` are unchanged, so no required-status-check rename hazard, and `Lint` has no `if:` so it still reports unconditionally. `helm-ci.yaml:54-70` is comment-only — every added and removed line begins with `#`.

Bugbot's one thread ("macOS xargs rejects GNU `-r`") was wrong; re-verified on macOS that `printf '' | /usr/bin/xargs -0 -r echo NOPE` returns 0 with the flag accepted. The resolution is correct.

— drafted with Claude Code

LukasWodka:
Both findings on #754 are correct and both are fixed in 80671e8. I reproduced each before touching anything.

## 1. `make parse` failed open — confirmed, fixed

Reproduced exactly as described, under both shells:

```
$ make parse            # tree with no scripts/ dir
find: scripts: No such file or directory
all shell scripts parse
$ echo $?
0
```

You're right that this got worse, not better, with #753: the enumerated shellcheck line that used to run straight after would have failed on the same breakage, and #753 made `parse` the *entire* content of a required check. It now materialises the list, refuses a zero count, and prints the derived count.

Four proofs, all under dash and bash: missing dir → non-zero, zero `.sh` → non-zero, real parse error → non-zero, clean tree → `all 39 shell scripts parse`.

Worth recording that my **first attempt at the third proof was inert**: I planted `if [ -z "$x" ; then :; fi`, which is valid bash — the missing `]` is a runtime error, not a parse error — so `bash -n` passed it and I briefly read `exit=0` as a broken fix. Redone with an unterminated `if`, after checking `bash -n` rejects it standalone. An inert mutation and a real one look identical in a log, which is the whole reason to check the anchor.

## 2. "Strict subset" was set-wise true, scope-wise false — confirmed, fixed

Verified: `code-quality-caller.yml` triggers on `pull_request` + `workflow_dispatch` only, no `schedule:`, and passes `all-files: ${{ inputs.all-files || false }}` → `false` on every PR. So #753 as written left **no** CI job doing a whole-tree error-severity sweep, and this PR's own green `quality / shellcheck` reading `Shell files to check: 0` is the demonstration.

Rather than document the regression, I removed it: **`Lint` now runs `make lint`** (parse + derived whole-tree shellcheck) instead of `make parse`. This costs nothing to install — shellcheck being preinstalled is the finding this PR is built on — and it restores `make lint` ≡ CI `Lint` parity (backend#1850). Two complementary gates: whole-tree here, diff-scoped in the org job, neither installing anything, neither holding a file list. That makes the superset claim actually true rather than sharpened.

Before arming it on a required check I verified the 44-file set under the **runner's** shellcheck, not just my newer local one:

```
$ docker run --rm -v "$PWD:/mnt" -w /mnt koalaman/shellcheck:v0.9.0 \
    --severity=error --exclude=SC1091 $(cat files.txt)
$ echo $?
0
```

## 3. #754 × #755 — reproduced end to end, and it bit me an hour later

Your merge analysis is right, including the part that matters. Real 3-way `git merge-file` from `8de5d64`: **one** conflict in the Makefile, and it is the seconds comment. Resolving only that yields a Makefile with no `SHELLCHECK_FILES` definition and line 253 still reading `$(SHELLCHECK_FILES)` — merged clean, no marker. Running it:

```
$ make lint-warnings
No files specified.
Usage: shellcheck [OPTIONS...] FILES...
$ echo $?
0
```

The advisory sweep becomes a permanent no-op reporting green, exactly as you said, and `|| true` is what hides it.

Then the same class of thing caught me. Develop moved (#747) and I resolved the Makefile conflict with `git checkout --ours`, which takes the whole file — silently dropping develop's *non-conflicting* addition of `scripts/tests/telemetry-vocabulary-agreement.sh` to the `drift` target. Nothing would have failed; the guard would just have stopped running. A concurrent session's merge (b4ab27b) kept it, so I discarded my resolution and rebuilt on theirs. I'd written the warning about damage-outside-the-markers an hour earlier and still walked into it, which is a decent argument for your "never resolve this through GitHub's conflict editor".

Agreed on the sequencing: **#754 first, then #755 rebased** with `lint-warnings` rewritten onto the derived set, `make lint` kept (it is now what `Lint` runs), and `Ensure shellcheck` dropped. I've left that note on #755 rather than touching its branch.

One correction to your framing: #755 is not the opposite remedy for backend#2182/#2183 — those are about making the **R8 manifest step** a gate, which this PR deliberately leaves untouched. The contradiction is narrower than "someone has to pick one": it is only about where shellcheck lives and whether the list is enumerated.

Thanks — finding 1 in particular was a real fail-open in a required check, introduced by the PR that was arguing against fail-open guards.

LukasWodka:
bugbot run

Makefile
```diff
@@ -148,7 +183,9 @@ install-hooks:
 
 # ---- individual targets ------------------------------------------
 
-# lint: standard-checks.yml `Lint` + installer-tests.yaml `static`.
+# lint: both halves, so `make check` and the pre-push hook are unchanged by
+# #753 splitting them. In CI the halves now live in different places: `parse`
```
saadqbal: Stale after the fixup — `Standard checks / Lint` now runs `make lint`, i.e. both halves, so the halves do not live in different places. Line 18 of this same file says so correctly. Given this PR's whole case rests on a comment that had gone false, worth not shipping a fresh one.

Makefile
```diff
@@ -158,12 +195,84 @@ install-hooks:
...
+# a child failure as a non-zero exit. (Bugbot, #630.) The same hazard is why
+# the `shellcheck` target materialises its file list instead of piping it.
 .PHONY: lint
-lint:
-	@find scripts -type f -name '*.sh' -print0 | xargs -0 -n1 bash -n
-	@echo "all shell scripts parse"
-	shellcheck --severity=error --shell=bash $(SHELLCHECK_FILES)
+lint: parse shellcheck
+
+# parse: the half CI still runs, as `Standard checks / Lint` -> `make parse`.
```
saadqbal: Same thing: `Lint` runs `make lint`, not `make parse`. Also, while you're in here, line 105 reads "the 42 the derivation actually finds" in the present tense — it's 44 now (CI on this head prints it).

saadqbal (dismissed):
Measured the coverage delta myself instead of taking the body's word for it. Exported the head tree, ran both sets: the enumerated list (36 files now that #747 landed) is a strict subset of the derived 44, `comm -23` empty, and both are green — nothing silenced, nothing imported. All 17 shell entries in `scripts/manifest.sha256` are in the derived set, so every file the installer fetches is linted. Dropping `--shell=bash` is a no-op here: everything in the set is `#!/usr/bin/env bash` except `docker/k3s-cuda/k3d-entrypoint-tracebloc-cdi.sh`, which the old list never covered. No new `# shellcheck disable=` anywhere. Ran `make lint` under `/bin/sh` and `/bin/dash` in throwaway repos — 39 parse / 44 shellcheck green, zero-file guards on both halves exit 1, missing `scripts/` exits 1, a planted parse error propagates. And the `Lint` job log on this head settles the load-bearing claim empirically: `all 39 shell scripts parse` / `shellcheck: 44 file(s)`, 16 s, no install step. Keeping the whole-tree half in `Lint` after the scope review is the right call — approving.

Two comments in the Makefile went stale in the fixup; inline. And two things outside the diff: the title still says "one shellcheck gate" when the design is now deliberately two (whole-tree in `Lint`, diff in `quality / shellcheck`) — worth fixing since the body is what people will read this by. `.pre-commit-config.yaml` also still claims *"Same gate as CI: error severity, bash dialect, installer scripts only"* and scopes itself `^scripts/.*\.sh$` with `--shell=bash`; after this PR that's wrong on two of three counts, and the regex reaches 37 of the 44 — missing `docker/k3s-cuda/*.sh`, `docs/migration-tools/*.sh` and `test_helper.bash`. It is the same enumerated-scope defect the new Makefile comment argues against, so it belongs in this PR.

nit for the record: the body still says "same `--exclude=SC1091` on all three" — neither removed invocation passed it. No behaviour change (SC1091 is info-level, both ran at `--severity=error`), but it isn't accurate.

One pre-merge check I can't do: branch protection isn't readable with my token, so someone with admin should confirm `quality / shellcheck` is genuinely a required context on develop before this merges.

Makefile
LukasWodka: Correct, and fixed in 9d811b3. This comment described the *previous* commit: the halves did live in different places until `Lint` went from `make parse` to `make lint`, and I updated line 18 without updating this one. Now:

> lint: both halves, and `Standard checks / Lint` runs exactly this target, so the pre-push tier and the merge gate cannot disagree about what linting means (backend#1850).

Your framing is the right one — a PR whose entire case is "a comment had gone false" has no business shipping a fresh false comment. I swept the four files for every claim of that shape rather than just this line, and found one more you didn't flag (a bare "yields 42" in the header block).

Makefile
LukasWodka: Both fixed in 9d811b3.

On the count: rather than change 42 to 44, I removed the counts from the prose entirely. It moved twice in one day — 42 when #753 was measured, 44 once #747 landed — and both targets already print their live count when they run (`shellcheck: 44 file(s), severity=error`). That is exactly the `BATS_TEST_COUNT` reasoning three declarations above, which says a number that is wrong is worse than no number and re-hardcoding today's value only resets the drift clock. I'd been restating a count in prose while deleting a restated file list two screens below.

The one figure left is pinned to a commit — "measured on develop at 8de5d64" — so it reads as a dated observation rather than current state.

I also found a fourth instance you didn't flag, same shape, in the header block.

LukasWodka:
Pushed 9d811b3. Both of Asad's threads addressed and resolved, plus the CI failure this PR was sitting on.

## The `bats` cancellation was #753 happening again, to me

`bats (bash unit, mocked)` reported `cancelled` — which, as #753 documents, is what a `timeout-minutes` kill looks like:

```
started   11:58:12Z
completed 12:08:28Z          <- 10m16s against timeout-minutes: 10
  success    Set up job
  success    Run actions/checkout
  cancelled  Install bats     <- empty log, no `bats --version`
  skipped    Run bats
```

The first draft of this PR deliberately left the two `apt-get install bats` calls alone, reasoning that bounding installs was #753's own scope. That reasoning was wrong for a *required* check, and it took about two hours to be proven so. Both are now bounded and retried, using the shape `Pull <distro> (bounded, retried)` already uses in this repo.

**The sizing is measured, and my first attempt at it was wrong.** 120 s bounds × 3 attempts + backoff = 780 s worst case, which overruns a 600 s budget and would report `cancelled` again — reintroducing the exact bug. Healthy numbers from run history are install **9 s** and bats **155 s**, so: 60 s bounds, 3 attempts, backoff only *between* attempts (10 + 20, no trailing sleep), worst case 390 s + ~155 s. `unit-bash`'s budget goes 10 → 15 to match its sibling.

Three details worth flagging for review:

- **`sudo timeout`, not `timeout sudo`** — timeout must run as root so it signals `apt-get` directly, rather than signalling `sudo` and trusting TERM to be forwarded.
- **`if [ "$i" -lt 3 ]; then sleep; fi`, not `[ ... ] && sleep`** — the default Actions shell is `bash -e`, where a bare test returning false on the final iteration aborts the script *before* the `::error::` line prints. That would turn a diagnosed failure back into a silent one, which is the whole thing #753 is about.
- **`-qq` dropped** from the install, `bats --version` kept as the proof-of-install marker — its absence is what identified the stall.

Verified by extracting the **live** `run:` block out of the YAML with a yaml parser and executing it under `bash -e` against a stubbed `apt-get`, rather than testing a retyped copy: always-fails → 3 warnings, the `::error::` line, exit 1, 30 s; fails-once → warning then success; healthy → marker, exit 0, no sleep.

## `E2E auth-proxy (squid)`

Not this PR, and I checked rather than assumed, since it first appeared on my commit. My `installer-tests.yaml` change at that commit was **comment-only** and confined to the `static` job region; the squid job runs `bash scripts/tests/e2e-proxy.sh`, which never invokes `make`. 7 of the 8 most recent runs of that job are green, including two on `develop`. It is the known squid Service-readiness race, and it is not a required context. It should clear on this run; if it doesn't I'll rerun that job alone.

## Not merged

Approved, but I'm holding until the run is green with no unresolved threads. Two things I won't decide unilaterally: the sequencing against #755 (note posted there), and the backend#2182/#2183 duplicate — both are on Lukas.

LukasWodka:
bugbot run

LukasWodka:
**Green at 9d811b3 — 43 pass, 0 fail, 0 unresolved.** Re-requesting review from you both: the approval at `80671e8` was auto-dismissed by the two fix commits (stale-review dismissal on new commits), and each of you left findings I'd rather have you confirm than take my word on.

Verified the changed jobs specifically rather than trusting the rollup:

| Check | Result | Evidence |
|---|---|---|
| `Lint` | success | log reads `all 39 shell scripts parse` then `shellcheck: 44 file(s), severity=error` — the whole-tree sweep is really running, which was @aptracebloc's finding 2 |
| `bats (bash unit, mocked)` | success | `Install bats (bounded, retried)` succeeded; this is the step that reported `cancelled` at 12:08 |
| `Unit tests` | success | same bounded install |
| `Static analysis` | success | R8 / check-facts / check-style / PSScriptAnalyzer all intact, shellcheck step removed |
| `quality / shellcheck` | success | diff-scoped, as designed |
| `E2E auth-proxy (squid)` | success | cleared without a rerun — the known readiness race, not this diff |

What changed since the approval:
- `80671e8` — `make parse` fails closed; `Lint` runs `make lint` (whole-tree). @aptracebloc's findings 1 and 2.
- `9d811b3` — four stale comments (@saadqbal's two threads, plus two more of the same shape I found sweeping); bats install bounded and retried in both workflows after it stalled on this very PR.

Still open and deliberately not decided here: the **#755 sequencing** (note on that PR — its `lint-warnings` becomes a silent no-op if it merges after this) and the **backend#2182 / #2183** duplicate. Both are Lukas's calls.

aptracebloc (approved):
Re-reviewed at `9d811b39`. Both of my earlier findings are fixed, and fixed with reproductions rather than prose. Approving.

**`make parse` no longer fails open.** It now materialises the list, counts it, refuses zero, and captures `rc` after the pipe. You reproduced my repro and posted four mutation proofs — including flagging that your first anchor was inert, which is the part most people leave out. CI on this head: `all 39 shell scripts parse`.

**The scope-wise gap is closed by removing the regression, not documenting it.** `Lint` runs `make lint` = `parse` + a derived whole-tree `shellcheck`; CI logs `shellcheck: 44 file(s), severity=error`. I independently re-confirmed the org job reads `git diff --name-only --diff-filter=ACMR "$BASE_SHA...HEAD"` unless `all-files: true` and exits 0 on an empty set, so the two gates are genuinely complementary now rather than one silently replacing the other with less.

**The bats hardening is a real self-caught defect** with measured sizing. I tested the retry loop under `bash -e` with stubs: the `if`/`fi` and `bats --version && exit 0` shapes are safe, and the `::error::` line does print. 390 s worst case against the raised 15-minute budget is right.

All 8 required contexts green. The only reason this is blocked is that the two fix commits auto-dismissed the earlier approval.

## Two nits, neither blocking

**`installer-tests.yaml:94`** — "same SC1091 exclusion as the step deleted here" isn't true of that step. The deleted step ran `shellcheck --severity=error --shell=bash <files>` with **no** `--exclude` (verified on `c564a15d`). The claim holds for the *org* job, which does use `--exclude=SC1091` — just not for the one being compared. A reader who trusts "so no verdict changes" won't notice the new whole-tree gate is strictly more permissive on SC1091 than the sweep it replaced.

**The `shellcheck:` target omits the org job's `[ -f "$f" ] || continue`.** Tested: with `scripts/check-facts.sh` deleted from disk but still in the index, it stays in the 44-operand argv. Mid-refactor `rm scripts/foo.sh` then fails the pre-push hook with a shellcheck "does not exist" error instead of something useful — and it's a divergence from the rule the comment claims to mirror exactly ("Same rule, same flags").

Also informational: the `E2E auth-proxy (squid)` failure at `80671e8` was a flake, not this diff — the #119 assertion failed once at 11:59, passes on develop's last three runs, passed on #755's identical develop base, and is green here. This PR touches no installer script.

## On #755 — two things to know before either merges

**The PSScriptAnalyzer hazard I flagged earlier is gone**, and this PR is what killed it. Changing `run: make parse` → `run: make lint` made the `make lint` step **byte-identical on both sides**, so it's now context rather than a change. The `standard-checks.yml` conflict shrank to the `Install shellcheck` region only, and #755's advisory-shellcheck and PSScriptAnalyzer steps both merge in cleanly below it. I verified the merged file contains both intact.

**But the `$(SHELLCHECK_FILES)` no-op is still live.** Merged `9d811b39` × `74eeb059`, resolved only the one visible Makefile conflict (a timing comment), and ran it with an argv-logging stub:

```
$ make lint-warnings
STUB shellcheck argv: --severity=warning --shell=bash
No files specified.
EXIT=0
```

`SHELLCHECK_FILES` has no definition in the merged Makefile — only a mention inside a comment — while #755's `lint-warnings` still references it. Zero operands, `|| true` swallows the usage error, exit 0. For contrast `make lint` on the same merged tree is healthy: 39 parsed, 44 shellcheck operands.

One resolution hazard worth naming: in the merged `installer-tests.yaml` the conflict ends at line 153 and `unit-bash:` starts at 155, with your new bounded bats install at 178. Resolving that file with `git checkout --theirs` would silently drop it — the same whole-file hazard that bit the Makefile an hour after you wrote the warning about it. Resolve region 48–153 only, hunk by hunk.

Also correcting myself from my earlier comment: I characterised #754 and #755 as opposite remedies for the same gap. That was wrong, and your correction is right — backend#2182/#2183 are about making the R8 step a *gate*, which this PR explicitly leaves alone. #755 is a different answer on where shellcheck lives, not the opposite of this one.

— drafted with Claude Code

LukasWodka:
/fr-pass

PR #759, opened 2026-08-19
test(installer): the bats host is an input — declare it, don't inherit it (backend#2208)

LukasWodka:
Fixes tracebloc/backend#2208

## The bug

Three `install_client_helm` flow tests in `scripts/tests/install-client-helm.bats` pass on CI and fail on every Apple Silicon Mac:

- `same client re-run is allowed (upgrade in place)`
- `same client in a different namespace -> upgrades in place, no duplicate`
- `different-namespace reconcile works WITHOUT jq (Bugbot #284)`

```
MySQL 5.7 is required here, but amd64 emulation could not be verified on this
Apple Silicon Mac — enable Rosetta and re-run (or set TRACEBLOC_ALLOW_ARM64=1 to override).
```

#756 made `_assert_engine_runs_on_this_arch` run on macOS as a fail-closed backstop, and it probes the **real host**: `uname -m`, plus a live Rosetta/Docker smoke on Darwin. `setup()` pinned neither `OS` nor `ARCH`, so these tests inherited the developer's CPU.

Only these three, because they are the only flow tests whose `helm list` mock reports an **existing release** — the one path that resolves the engine to `5.7` and therefore engages the gate. Every other flow test resolves `fresh` → `8.4` and returns early at the engine check.

**CI cannot see this by construction.** CI is Linux/amd64, so the gate returns early at the arch check and `Unit tests` stays green. The one platform the gate exists *for* is the one CI never runs — so it is a red local suite that CI calls green, for every Mac developer.

## The fix

Pin `OS=Linux` / `ARCH=x86_64` in that file's `setup()` — to what CI actually is — so a local run reproduces CI and no test's verdict depends on the machine it runs on.

This keeps the gate **armed**. `TRACEBLOC_ALLOW_ARM64=1` would also turn the suite green, and that is precisely why it is not used here: it switches off the behaviour #756 added. Every arch-sensitive test in the file already sets `OS`/`ARCH` for itself (`_arch_gate_ctx`, and the engine tests at lines ~1306-1567), so they override this and keep driving the real rule.

**No new gate tests.** #756 already covers both arms — emulation ok → proceeds, missing → refuses, helper absent → refuses fail-closed, 8.4 → skips, escape hatch → skips. More would duplicate, not strengthen.

## A comment this exposed

The block above `_assert_engine_runs_on_this_arch` still read:

> macOS is NOT covered here: this gate is Linux-only … That macOS engine-aware gate is a separate fix … deliberately out of scope here

sitting directly above the `Darwin)` arm #756 added. Anyone trusting the prose over the code concludes macOS is ungated. Rewritten to describe what the function does, and to say that it probes the real host so callers under test must declare `OS`/`ARCH`.

## Mutation proof (run on an arm64 host)

| Mutation | Result |
|---|---|
| drop the `setup()` pin | the same 3 tests redden again |
| break the gate's `Darwin` arm (`return 0` unconditionally) | 2 of the gate's own tests redden |

The second row is the one that matters: it shows the gate is **still armed** after this change, so the pin fixed the tests without disarming the thing they were tripping over. A fix that quietly disabled the gate would satisfy row 1 alone and look identical.

## Verification (arm64 macOS, no `TRACEBLOC_ALLOW_ARM64`)

- `bats scripts/tests/install-client-helm.bats` → **159/159**, was 156/159
- `bats scripts/tests/setup-macos-arch.bats scripts/tests/preflight.bats` → **157/157**
- `scripts/gen-manifest.sh --check`, `scripts/check-facts.sh --check`, `scripts/check-style.sh`, `scripts/tests/check-drift.sh` → clean
- `bash -n` + `shellcheck -S error` on the changed lib → clean

Reproduced beforehand on a clean `git archive origin/develop` checkout at `526fc49` and at tip `c238e5f`, so this was never branch-specific.

---

> [!NOTE]
> **Low Risk**
> Test harness and documentation only; installer behavior unchanged aside from comment accuracy and checksum manifest.
> 
> **Overview**
> Fixes **local-only failures** in three `install_client_helm` flow tests (same-client upgrade / cross-namespace reconcile) on Apple Silicon: those mocks report an **existing Helm release**, which resolves MySQL to **5.7** and runs **`_assert_engine_runs_on_this_arch`** against the **real host** (Rosetta/`uname`), not the test harness.
> 
> **`install-client-helm.bats` `setup()`** now sets **`OS=Linux`** and **`ARCH=x86_64`** (CI’s platform) so default flow tests don’t inherit the developer CPU. Arch-sensitive cases still override via `_arch_gate_ctx` and engine tests; this is **not** `TRACEBLOC_ALLOW_ARM64`, so the gate stays armed.
> 
> **Comment above `_assert_engine_runs_on_this_arch`** is updated to state that **macOS is gated** (client#756) and that callers under test must **declare `OS`/`ARCH`**.
> 
> **`scripts/manifest.sha256`** is refreshed for the touched lib file.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f2b9f668dadc686d3a7112ed2447fd84809646e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saadqbal (dismissed):
Verified the reasoning against the source rather than the description, and it holds:

- `_assert_engine_runs_on_this_arch` reads `${ARCH:-$(uname -m)}` and `${OS:-}`, so the pin returns at the `x86_64` arm before the OS probe is reached — that's why the three existing-release reconcile tests were the only ones affected (they're the only flow tests whose `helm list` mock resolves the engine to 5.7).
- The gate stays armed: `_arch_gate_ctx` sets its own `OS=Linux; ARCH=aarch64`, and the engine tests at ~1245-1584 each set `ARCH` explicitly, so they override `setup()` and keep driving the real rule. Using `TRACEBLOC_ALLOW_ARM64=1` instead would have disarmed it — good call not to.
- The comment fix is the real find. The old prose said "this gate is Linux-only … deliberately out of scope" while sitting directly above the `Darwin)` arm #756 added. That's the kind of stale comment that gets trusted over the code.

Pinning the host in `setup()` is the right fix — a unit test's verdict depending on the developer's CPU is the bug, not the symptom.

aptracebloc (approved):
Approving.

The diagnosis holds: `_assert_engine_runs_on_this_arch` (client#756) probes the real host, and `setup()` pinned neither `OS` nor `ARCH`, so the three flow tests whose helm mock reports an existing release — the only path that resolves the engine to 5.7 and engages the gate — inherited the developer's CPU and failed on Apple Silicon, invisibly to a Linux/amd64 CI that returns early at the arch check.

The fix is the right shape: pin `OS=Linux`/`ARCH=x86_64` so a local run reproduces CI, and pointedly **not** `TRACEBLOC_ALLOW_ARM64`, which would disarm the very gate #756 added — the arch-sensitive tests still set their own `OS`/`ARCH`, so they keep exercising the real rule. It also corrects the inverted comment that claimed the gate was Linux-only / macOS-uncovered when #756 had made macOS the one path where this is the sole late gate.

Test-only + a comment; CI green, no conflict, no open threads.

— drafted with Claude Code

PR #769, opened 2026-08-20
fix(installer): never write an envelope larger than the machine (backend#2220)

shujaatTracebloc:
> **Stacked on #768** — base is `feature/2220-envelope-provenance`. Merge #768 first, then retarget this to `develop` before merging.

## The bug

A machine below the training floor was handed `cpu=2,memory=8Gi`. On a box with ~4 GiB allocatable **that envelope is larger than the machine**, so every training pod stays Pending forever — and `--reset-then-reuse-values` re-applies that value on every upgrade, indefinitely.

It's the exact failure `install-k8s.ps1`'s own header says the machine sizing exists to remove — *"dead on arrival on nodes under 8 GiB (the WSL2 field case — nothing could ever schedule)"* — recreated by its own fallback.

## This is a permitted band, not a hypothetical

`preflight.sh` hard-fails below **5 GB** on Linux and only **warns** on macOS/Windows (`PF_MIN_MEM_GB=5`, `PF_WARN_MEM_GB=8`), while its own comment says *"16 GiB is needed to train locally (a job's limit is ~8 GiB+)"*.

So the installer **permits exactly the machines whose envelope cannot schedule**. The allowed band and the unschedulable band overlap.

## Root cause

Two different situations produced the same answer. *"I cannot read the machine"* and *"I read it and it's too small"* both returned empty, and both fell through to the literal. Separating them is the whole change:

| Situation | Before | After |
|---|---|---|
| Cluster unreadable | literal | **literal, unchanged** — we can't do better than history, and warning about machine size would be a fabrication |
| Below the floor | literal (unschedulable) | **the honest remainder** (≥1 core, ≥1 GiB) — it *fits*, so a run can be scheduled and fail for a **reason** instead of hanging, and the operator is warned |
| Below even that | literal | literal **plus a hard warning** — `cpu=0` isn't a training request, so there's no honest number to write |

On that last row: refusing the install would break the run-but-don't-train-locally case preflight deliberately allows (it permits 5 GB to *run* while recommending 16 GB to *train*). So it warns instead.

## The warning lives in the caller

Not in the resolver. `_training_resources` / `_training_provenance` are captured with `$(...)` and their tests compare the whole output, so **a warn emitted from the resolver would land inside `RESOURCE_LIMITS`**. Pinned by a test that captures the value and asserts it's clean.

`_machine_training_resources` keeps its exact old semantics (empty when non-viable), so its callers and its golden replay are untouched; the new viability-aware `_machine_training_ceiling` sits beside it.

## Two expectations changed rather than added — flagging deliberately

- bats: *"below-floor machine falls back to the static default"*
- Pester: *"below-floor machine falls back to the static default"*

Both asserted the **unschedulable literal as if it were the contract**. Neither loses coverage: the unreadable-cluster tests still pin the case where the literal is genuinely right, which is now a *distinct branch* rather than the same one. I'd rather call this out than have a reviewer discover a silently-flipped assertion.

## The golden replay got stronger, not just repaired

The Pester replay used to map **every** non-viable vector onto the literal — which is exactly what let this bug live inside a passing replay. It now checks them against the contract's own `render_gi`.

Those numbers already matched: `wsl2-too-small-4gi` renders **3c/1Gi**, which is precisely what the installer now writes. So the accessor and the installer agree on small machines too, and that agreement is now asserted.

*(The bats replay still exercises the viable-only wrapper by design — its flattened vector table carries no remainder for non-viable rows. The targeted bats tests cover those instead. Asymmetry noted rather than papered over.)*

## Verification

- **bats** 176 ok (+4), same 3 pre-existing failures clean `develop` has (#64–66, they need real helm)
- **Pester** 754 passed / 0 failed (+4)
- **PSScriptAnalyzer** 0 errors; `make lint` clean over 47 files
- `gen-envelope-embed.sh --check` clean; `manifest.sha256` regenerated
- No `client/**` chart files touched, so no Chart bump beyond the 1.9.49 this branch already carries

Refs: RFC-BACKEND-664 §P0, client-runtime#358, client#766, client#768, backend#2167

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 90999982240dea463f32c8ff0c1962681892d5c5. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


saadqbal (commented):
Strong finding, and I verified the premise verbatim rather than taking it — `preflight.sh:27-28`:

```bash
PF_MIN_MEM_GB="${PF_MIN_MEM_GB:-5}"   # hard-fail below this (Linux; warn on Mac/Win)
PF_WARN_MEM_GB="${PF_WARN_MEM_GB:-8}" # warn below this (comfortable to run)
```

So the installer really does permit the 5–8 GB band on Linux and *below* 5 on Mac/Win with only a warning, while a job's limit is ~8 GiB. The permitted band and the unschedulable band overlap, and `--reset-then-reuse-values` makes the bad value permanent. That's a genuine hole, not a hypothetical.

**The root cause is the right diagnosis.** "I cannot read the machine" and "I read it and it's too small" both returned empty and both fell to the literal — one answer for two situations. Separating them is the fix, and the three arms are each correct on their own terms: unreadable keeps history (warning about machine size would be fabrication), below-floor writes the honest remainder so a run *fails for a reason* instead of hanging, and not-even-requestable keeps the literal and flags it.

Tests are the full matrix, 11/11, each arm named — override wins / existing choice carried / historic default not carried / viable sized / below-floor remainder / below-floor flagged / unreadable keeps default / too-small-for-1c1Gi flagged / kubectl absent. Mutation-proof: reverting the `|undersized` arm to the literal reddens `below-floor machine gets the honest remainder, not an unschedulable literal` precisely.

**The case I'd single out is `the resolver itself never emits a warning (it would corrupt the value)`.** Keeping the resolver pure and letting the caller warn is the right separation, and having a test for it means the next person who reaches for a `warn` inside a function whose stdout *is* the value gets stopped. That's a subtle one to have thought of.

### Three things before this can land

**1. The `DIRTY` is an artifact of the stacking, not a develop conflict.** Diagnosed rather than guessed: this PR's base is `feature/2220-envelope-provenance` (#768's branch), and it does **not** contain #768's newest commit `516e30ca` — the one-shared-lookup fix. So it conflicts against its own moving base. `git merge-tree` against `develop` reports **zero** conflicting files, so once #768 lands and this retargets, it should be clean. Merging #768's branch in works too.

**2. The chart version will collide again.** Both #768 and this say `1.9.50`; `develop` is at `1.9.49`. Because this is stacked, it inherited the bump. When #768 merges, this will carry `1.9.50` *identically* to develop — so git shows no conflict, and `chart-version-guard.sh` will demand `1.9.51` once the base moves.

That's the **third** instance of this exact shape today (#764, #768, now this), which makes it structural rather than three mistakes. Worth a one-time fix in the guard: it currently asserts the version *changed* relative to the PR base, which parallel chart PRs each satisfy independently. Asserting it is also **greater than the target branch's** version would catch the collision at push time instead of relying on a reviewer noticing that two PRs picked the same number.

**3. The stacking itself.** Org standard is no stacked PRs on open ones, and this is the second in the 2220 chain to do it (cli#539 did, resolved by #538 landing first). It works out when the parent merges promptly — just noting the cost is exactly the two items above.

One minor, non-blocking: the fall-through arm re-probes `kubectl get nodes` to decide whether to set `_TB_TRAINING_UNSCHEDULABLE`, after `_machine_training_ceiling` already read the nodes. That's a second lookup that can disagree with the first — the same class as the finding on #768 — though here the disagreement direction is safe (a cluster that became unreadable produces no warning rather than a false one). Not worth a change on its own; mentioning it because the pattern is now recurring in this chain.

PR #777, opened 2026-08-20
fix(shell-gate): `||` is not a pipe, and `|&` is (backend#2264)

LukasWodka:
Part of tracebloc/backend#2264.

## The gate rejected the output of its own remediation

Converting the fleet for #2264 produces exactly this line:

```bash
cmd || grep -q needle <<<"$out"
```

There is no pipe in it. But the hazard regex matched the **second bar of the `||`**, so the scanner reported the very form this repo documents as the house idiom. It would have blocked every cleanup PR in the #2264 wave — and a gate that rejects its own fix is how a gate teaches people to switch it off.

Found by running the gate against my own converted `cli` scripts, not by reading it.

## Fix

Neutralise `||` into `\001\001` on a **probe copy** before the hazard test. `\001` cannot occur in a shell source line, and the original line is still what gets printed, so offender output is byte-identical.

**Second, smaller gap** found while writing that test: `|&` — bash's pipe-both-streams — *is* a pipe and carries the identical hazard, but the bar had to be followed by space-or-name, so `noisy |& head -1` was missed entirely. The bar now takes an optional `&`. A gate that misses a real pipe form is the "advice, not a gate" failure from the house rules, so I fixed it here rather than filing it.

## Mutation-proof

Baseline 33 green. Each mutation asserted its anchor matched **exactly one line** before running — an inert mutation and real coverage are indistinguishable in a log otherwise.

| mutation | reddens |
|---|---|
| remove the `\|\|` neutralisation | **3** — tests 27, 28, 29 |
| hazard test reads `line`, not `probe` | 1 — test 28 |
| drop `\|&` from the head arm | 1 — test 32 |
| drop `\|&` from the grep -q arm | 1 — test 33 |

Every mutation is caught by the test that *names* the property, not by one shared failure that any mutation would trip.

The load-bearing test is **"a real pipe is still flagged on a line that also contains `||`"**. Without it, the three new spare-cases would all pass if the hazard test were deleted outright — three tests asserting the right property and proving nothing. That's the shape client#776 is about, so it would have been a poor look to ship it here.

## Verification

- the gate reports **zero** over the whole tree
- `check-style.sh` clean, `gen-manifest.sh --check` up to date
- 23 other bats suites run, **0 failures** (the remainder are unaffected — this touches only the scanner and its own test file)

## Sequencing note

This wants to land **before** the remaining #2264 conversion PRs, or the gate will be wrong about them once it's armed. It does not block them from merging today, since the gate is not yet a required check anywhere.

---

> [!NOTE]
> **Low Risk**
> Linter-only regex and test changes; no runtime, auth, or data-path code. Residual risk is false positives/negatives in the style gate, which the new tests pin.
> 
> **Overview**
> Stops the pipefail early-close scanner from flagging the house remediation idiom (`cmd || grep -q …`) by treating `||` as **not** a pipeline, while still catching real `|` / `|&` hazards on the same line.
> 
> The matcher now runs on a probe copy where `||` is replaced with a `\001` stand-in that also acts as a regex boundary, so a later `|| grep -q` cannot be borrowed by a plain `| grep`, and glued forms like `head||die` stay flagged. `|&` (pipe both streams) is now recognized as a pipe.
> 
> Also documents and pins a known false positive: `set` inside a multi-line quoted fixture is read as the file’s own options; `# pipefail-guard: allow` remains the workaround.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a0a9d3fabef1f43188d0e79eb79d49c3a2004d20. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/tests/pipefail-early-close.awk
+++ b/scripts/tests/pipefail-early-close.awk
@@ -165,10 +165,18 @@ FNR == 1 {
   # `|&` is bash's pipe-both-streams and is a pipe like any other, so the bar is
   # allowed one optional `&`. It is NOT the `|&` of the terminator class below,
   # which is about what may FOLLOW `head`.
+  #
+  # \001 IS IN THE HEAD TERMINATOR CLASS for the same reason it is in the grep
+  # arms' `[^|\001]*`: the stand-in has to read as a BOUNDARY, not as ordinary
+  # text. `producer | head||die` leaves `head` followed by \001, and without it
+  # in the class the pipe is missed -- a FALSE NEGATIVE this fix introduced and
+  # `develop` did not have (Arturo, client#777). The neutralisation and the
+  # boundary are one change; doing half of it moves the bug rather than fixing
+  # it, which is exactly what happened here twice.
   # The terminator class matters: `head` can be followed by a CLOSING delimiter,
   # not just whitespace or end-of-line. `x="$(cmd | head)"` ends at `)` and then
   # `"`, and requiring space-or-EOL missed exactly that shape (Bugbot #763).
-  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
+  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&\001])/ \
       || probe ~ /\|&?[[:space:]]*grep[^|\001]*[[:space:]]-[a-zA-Z]*q/ \
       || probe ~ /\|&?[[:space:]]*grep[^|\001]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
     sub(/^[[:space:]]+/, "", line)
--- a/scripts/tests/pipefail-early-close.awk
+++ b/scripts/tests/pipefail-early-close.awk
@@ -115,16 +115,32 @@ FNR == 1 {
 
   if (!(e_on && p_on)) next
 
+  # `||` IS NOT A PIPE, and must be neutralised before the hazard test below.
+  # Otherwise the SECOND bar of a `||` reads as a pipeline, and the scanner
+  # flags the very form this file recommends:
+  #     out=$(producer) || true
+  #     grep -q needle <<<"$out"        # fine
+  #     cmd || grep -q needle <<<"$out" # was reported as an offender
+  # Found while converting the fleet for backend#2264 -- the gate rejected the
+  # output of its own remediation, which is how a gate teaches people to
+  # disable it. \001 cannot occur in a shell source line, so it is a safe
+  # stand-in; the ORIGINAL line is still what gets printed.
+  probe = line
+  gsub(/\|\|/, "\001\001", probe)
+
   # The hazard: a pipe into a reader that closes early.
   #   `| head`        — closes after N lines
   #   `| grep -q`     — closes on the first match
   #   `| grep -m N`   — closes after N matches
+  # `|&` is bash's pipe-both-streams and is a pipe like any other, so the bar is
+  # allowed one optional `&`. It is NOT the `|&` of the terminator class below,
+  # which is about what may FOLLOW `head`.
   # The terminator class matters: `head` can be followed by a CLOSING delimiter,
   # not just whitespace or end-of-line. `x="$(cmd | head)"` ends at `)` and then
   # `"`, and requiring space-or-EOL missed exactly that shape (Bugbot #763).
-  if (line ~ /\|[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*q/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
+  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
+      || probe ~ /\|&?[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*q/ \
+      || probe ~ /\|&?[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
     sub(/^[[:space:]]+/, "", line)
     print curfile ":" FNR ": " line
   }
--- a/scripts/tests/pipefail-early-close.awk
+++ b/scripts/tests/pipefail-early-close.awk
@@ -39,6 +39,24 @@
 # `hazardous` (from pipefail-early-close.sh) seeds files that inherit both
 # options from a sourcing script; those start with the state already on.
 #
+# KNOWN LIMITATION -- a `set` line inside a MULTI-LINE QUOTED STRING is read as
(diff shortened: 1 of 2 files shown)


LukasWodka:
bugbot run

scripts/tests/pipefail-early-close.awk
```diff
@@ -115,16 +115,32 @@ FNR == 1 {
...
+  # which is about what may FOLLOW `head`.
   # The terminator class matters: `head` can be followed by a CLOSING delimiter,
   # not just whitespace or end-of-line. `x="$(cmd | head)"` ends at `)` and then
   # `"`, and requiring space-or-EOL missed exactly that shape (Bugbot #763).
-  if (line ~ /\|[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*q/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
+  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
+      || probe ~ /\|&?[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*q/ \
+      || probe ~ /\|&?[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
```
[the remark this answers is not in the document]
LukasWodka: Correct, and caught one level deeper than my fix went. Reproduced before fixing:

```
producer | grep needle && cmd || grep -q x <<<"$y"    → flagged
```

The `|` characters of a `||` were not just noise — they were the **stop** for `grep[^|]*`. Replacing them with a character the class permitted let the plain `| grep` span the rest of the line and reach the `-q` of the later `|| grep -q`: the same false positive the commit set out to remove.

Fixed by adding `\001` to the class on both grep arms (`grep[^|\001]*`), so the stand-in neutralises the pipe **and** keeps the boundary. Your note that the `head` arm does not span that way is also right — it takes no `[^|]*`, so it is untouched.

Three tests added. Two cover the cases you named; the third is the discrimination that makes them mean anything — a **real** `| grep -q` on a line of that exact shape is still flagged, so widening the stand-in until everything is spared would not pass.

One more thing this surfaced: I also shrank the stand-in from `\001\001` to a single `\001`, assuming the width mattered. Its mutation **survived** — the class does all the work — so that half proved nothing and I reverted it. Only the class change ships.

scripts/tests/pipefail-early-close.awk
```diff
@@ -115,16 +133,44 @@ FNR == 1 {
...
+  # `|&` is bash's pipe-both-streams and is a pipe like any other, so the bar is
+  # allowed one optional `&`. It is NOT the `|&` of the terminator class below,
+  # which is about what may FOLLOW `head`.
   # The terminator class matters: `head` can be followed by a CLOSING delimiter,
   # not just whitespace or end-of-line. `x="$(cmd | head)"` ends at `)` and then
   # `"`, and requiring space-or-EOL missed exactly that shape (Bugbot #763).
-  if (line ~ /\|[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*q/ \
-      || line ~ /\|[[:space:]]*grep[^|]*[[:space:]]-[a-zA-Z]*m[[:space:]]*[0-9]/) {
+  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&])/ \
```
aptracebloc: The `||` neutralisation opens a **false negative in the head arm** when `head` is glued directly to `||` with no space:

```bash
producer | head||die
```

`gsub` rewrites `||` → `\001\001`, so `head` is now followed by `\001` — which isn't in the terminator class `[)"'`;|&]`, and the pipe is missed. `develop` flags this line; this PR silently drops it (verified by running the head-SHA scanner on both).

This is the exact mirror of the boundary fix you already made for the grep arms — "neutralising the pipe was only half the job; the boundary has to survive." The `\001` was added to `grep[^|\001]*`, but the head terminator class didn't get the same treatment, so `\001` reads as "not a terminator" instead of "a boundary." Adding `\001` to the class restores it:

```suggestion
  if (probe ~ /\|&?[[:space:]]*head([[:space:]]|$|[)"'\''`;|&\001])/ \
```

`head||true` / `head|| :` stay spared by the earlier `|| true` guard, and the common spaced form `head || die` and the `)`-terminated `$(… | head)||…` were never affected — so this is a narrow shape, but it is a real hazard the base gate caught and this one no longer does.

— drafted with Claude Code

LukasWodka: Confirmed and fixed — your diagnosis was exactly right, including the framing.

Reproduced against both scanners before touching anything:

```
producer | head||die
  develop scanner : /tmp/ar.sh:3  ← flagged
  this PR scanner : (nothing)     ← dropped
```

So this was a **regression I introduced**, not a pre-existing gap. Applied your suggestion verbatim.

What stings is that this is the *same* mistake twice on the same three-line edit. Bugbot caught the grep arms spanning past the stand-in; I fixed those with `[^|\001]*` and wrote in the header that "neutralising the pipe is only half the job; the boundary has to survive" — and then did not carry that to the head arm sitting directly above it. The stand-in read as a boundary on one arm and as ordinary text on the other. You quoted my own comment back at me, which is fair.

Four tests added, **42 green**. Two cover the glued form (`head||die`, `|& head||die`); two are the discrimination that keeps those honest — `head||true` and `head|| :` must **still** be spared, or the fix degenerates into "flag anything containing head".

Mutation-proved: dropping `\001` from the class reddens **exactly tests 39 and 40**, nothing else. Anchor asserted to match one line before running, so an inert mutation cannot pass for coverage.

I also added the check that would have caught this without you: running develop's scanner and this one over every tracked `.sh` and diffing the offender lists. They are byte-identical (both 0 on the real tree). A whole-tree parity diff is the cheap way to notice "the base flags something I no longer do", and I should have run it when I first touched the arms rather than only testing the case I was thinking about.

Thanks — this one was a genuine catch, and the suggestion block saved a round trip.

LukasWodka:
bugbot run

aptracebloc (approved):
Re-review — approving. Fixes two operator misclassifications in `pipefail-early-close.awk` (the shell-lint early-close gate): `||` was wrongly read as a pipe, and `|&` was wrongly not.

The false-negative regression I flagged last pass — `producer | head||die` silently dropped because the `||`→`\001` rewrite left `head` before a `\001` not in its terminator class — is fixed in `a0a9d3fab`: `\001` was added to the head terminator class (`[)"'\`;|&\001]`), mirroring the grep arms' `[^|\001]*`. Verified in code; confirmed the `|| (true|:)` guard still runs first so `head||true`/`head|| :` aren't newly flagged; and confirmed the author added the pinning tests my finding said were missing (`head||die` and `|& head||die` flagged, `head||true`/`head|| :` spared) — derived, mutation-proof, specific. My thread is resolved, CI green, 0 open threads, MERGEABLE.

— drafted with Claude Code

PR #785, opened 2026-08-21
feat(#1528): wire DB_BOOTSTRAP_* into jobs-manager to re-parent the mint off edgeuser (S3)

saqlainsyed007:
## What

**backend#1528 S3** — the chart half of the bootstrap re-parent. The runtime seam already merged (**client-runtime#310**: `_bootstrap_db_config` resolves the minting identity from `DB_BOOTSTRAP_USER` / `DB_BOOTSTRAP_PASSWORD`, both-or-neither, falling back to `edgeuser` when unset). This PR injects those into the **jobs-manager** deployment, gated on a new S3 flag, so the account-minting DDL (`CREATE USER` / `GRANT` for `tb_credmgr` / `tb_meta` / `tb_ingest`) authenticates as **MySQL `root`** instead of the root-equivalent `edgeuser`.

While the mint authenticates *as* `edgeuser`, `edgeuser` cannot be `REVOKE`d / `DROP`ped. Re-parenting it onto `root` (which already holds the global `CREATE USER` + `GRANT OPTION` the mint needs) is the prerequisite that unblocks the rest of S3.

## Changes

- **`values.yaml`** — `bootstrapDbReparent` (operator override) + `bootstrapDbReparentByEnv{dev,stg,prod: false}` + `bootstrapDbPassword`. Default-**off** everywhere; this is the last, `edgeuser`-retiring step, flipped dev-first per fleet only once `serviceDbAccounts` + `perExperimentDbCreds` are verified and the S0 `SHOW GRANTS` snapshot is captured.
- **`_helpers.tpl`** — `tracebloc.bootstrapDbReparent`, resolving exactly like `tracebloc.serviceDbAccounts` (override → `…ByEnv` → normalized `CLIENT_ENV`), so the fleet's S3 posture reads out of one place and shares the backend#1723 alias normalization.
- **`secrets.yaml`** — `DB_BOOTSTRAP_PASSWORD` resolved **2-tier: pin → existing Secret → `fail`**. Deliberately **no** generate-random tier: jobs-manager *authenticates* as `root` and never runs `ALTER USER root`, so a random value the server was never told to expect would lock the minter out and CrashLoop it. **No** alphanumeric constraint (unlike the credmgr/tb_* passwords it is never interpolated into DDL — only a mysql-connector connection parameter; the baked root password `Edg9@Tr@ce` isn't alphanumeric anyway). Placeholder-guarded.
- **`jobs-manager-deployment.yaml`** — `DB_BOOTSTRAP_USER=root` + `DB_BOOTSTRAP_PASSWORD` (secretKeyRef), gated, both-or-neither.
- **`values.schema.json`** — three new keys. **`Chart.yaml`** 1.9.51 → 1.9.52 (+ `appVersion` lockstep).
- **Tests** — +8 helm-unittest cases across `secrets_test.yaml` / `jobs_manager_test.yaml`.

## Design notes (both confirmed with the reviewer route on the ticket)

- **Require-pin, fail-fast** over a baked `Edg9@Tr@ce` fallback — keeps the literal out of the chart (doesn't fight backend#947) and fits the operator-driven, prod-irreversible-adjacent rollout.
- **ByEnv gate** mirroring `serviceDbAccounts` — the backend#1752 lesson that a plain global bool caused a silent fleet break.

## Verification

- `helm unittest` full suite: **499 passed** (33 suites).
- `helm template` (ci/eks base): **default render = 0 `DB_BOOTSTRAP` refs (byte-identical)**; gate-on+pin wires the Secret (`DB_BOOTSTRAP_PASSWORD`) and both env vars; gate-on **without** pin fails fast with the guidance message; `bootstrapDbReparentByEnv.dev=true` wires it while `prod` (default) doesn't; placeholder pin rejected.
- `helm lint` clean; `chart-version-guard.sh` satisfied.

**Additive and default-off — nothing renders until the S3 rollout flips the flag per fleet.**

## Not in this PR (remaining S3)
- Narrow `client-runtime/mysql-client-initdb/10-edgeuser-bridge-grants.sql` once the mint no longer runs as edgeuser.
- Ops-gated, prod-irreversible: `REVOKE` edgeuser → `DROP USER`, fleet-staged dev→stg→prod, gated on the S0 snapshot.
- The literal sweep (backend#947).

Refs: backend#1528 (D10 close-out, S3) · client-runtime#310 (runtime seam) · backend#1752 (ByEnv rollout) · backend#947 (literal removal).

---

> [!NOTE]
> **Medium Risk**
> Handles MySQL root credentials for account-minting DDL, but the gate is default-off everywhere so existing installs stay byte-identical until an operator flips it. A wrong pin with the flag on would CrashLoop the minter.
> 
> **Overview**
> Wires **backend#1528 S3**: when `bootstrapDbReparent` is on, jobs-manager mints `tb_*` accounts as MySQL **root** (`DB_BOOTSTRAP_USER` / `DB_BOOTSTRAP_PASSWORD`) so `edgeuser` can later be revoked. Default installs stay byte-identical.
> 
> The gate mirrors `serviceDbAccounts` (operator override, else `bootstrapDbReparentByEnv` keyed on resolved `CLIENT_ENV`; all envs **false**). `bootstrapDbPassword` is 2-tier only (pin → existing Secret → **fail**): no random default, because jobs-manager never `ALTER USER root`. Placeholders are rejected; non-alphanumeric is allowed.
> 
> Chart **1.9.51 → 1.9.52**. Helm-unittest covers default-off, gate-on wiring, ByEnv, missing pin, and placeholder fail.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 18ba65a1dac6844e93f4aed2182f7ce646b14ac1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client/Chart.yaml
+++ b/client/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: client
 description: A unified Helm chart for tracebloc on AKS, EKS, bare-metal, and OpenShift
 type: application
-version: 1.9.51
-appVersion: "1.9.51"
+version: 1.9.52
+appVersion: "1.9.52"
 keywords:
   - tracebloc
   - kubernetes
--- a/client/templates/_helpers.tpl
+++ b/client/templates/_helpers.tpl
@@ -631,6 +631,26 @@ Usage: {{ include "tracebloc.ingestorDigest" . }}
 {{- end -}}
 {{- end }}
 
+{{/*
+  Whether to re-parent the account-minting bootstrap off edgeuser onto MySQL
+  root (backend#1528 S3). Resolves identically to tracebloc.serviceDbAccounts —
+  operator override first, else the per-environment default — so the fleet's S3
+  posture reads out of one place the same way, and the CLIENT_ENV normalization
+  (backend#1723) is shared. This is the LAST, edgeuser-retiring step, so
+  bootstrapDbReparentByEnv is false everywhere by default; flip dev first only
+  after serviceDbAccounts + perExperimentDbCreds are verified on the fleet.
+*/}}
+{{- define "tracebloc.bootstrapDbReparent" -}}
+{{- $override := (default dict .Values).bootstrapDbReparent -}}
+{{- if not (kindIs "invalid" $override) -}}
+{{- if $override }}true{{ end -}}
+{{- else -}}
+{{- $env := include "tracebloc.clientEnv" . -}}
+{{- $byEnv := default dict .Values.bootstrapDbReparentByEnv -}}
+{{- if get $byEnv $env }}true{{ end -}}
+{{- end -}}
+{{- end }}
+
 {{- define "tracebloc.clientEnv" -}}
 {{- $raw := (default dict .Values.env).CLIENT_ENV | default "prod" -}}
 {{- $aliases := dict "development" "dev" "staging" "stg" "production" "prod" -}}
--- a/client/templates/jobs-manager-deployment.yaml
+++ b/client/templates/jobs-manager-deployment.yaml
@@ -358,6 +358,24 @@ spec:
               name: {{ include "tracebloc.secretName" . }}
               key: TB_INGEST_PASSWORD
         {{- end }}
+        {{- if (include "tracebloc.bootstrapDbReparent" .) }}
+        # backend#1528 S3 (RFC-0003 D10 close-out): re-parent the account-minting
+        # bootstrap DDL OFF the root-equivalent edgeuser onto MySQL root, so
+        # edgeuser can be REVOKEd / DROPped. jobs-manager authenticates as this
+        # identity ONLY to CREATE USER / GRANT the scoped accounts (tb_credmgr /
+        # tb_meta / tb_ingest) — privileges those accounts deliberately lack.
+        # client-runtime#310's _bootstrap_db_config reads DB_BOOTSTRAP_USER and
+        # DB_BOOTSTRAP_PASSWORD together (both-or-neither), so both render here or
+        # neither. Rendered only when the S3 gate is on; a default install stays
+        # byte-identical (the mint keeps running as edgeuser).
+        - name: DB_BOOTSTRAP_USER
+          value: "root"
+        - name: DB_BOOTSTRAP_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "tracebloc.secretName" . }}
+              key: DB_BOOTSTRAP_PASSWORD
+        {{- end }}
         # Ingestor image wiring for the POST /internal/submit-ingestion-run
         # endpoint. jobs-manager spawns each ingestion Job from these values
         # (see client-runtime submit_ingestion_run._build_image_reference):
--- a/client/templates/secrets.yaml
+++ b/client/templates/secrets.yaml
@@ -90,6 +90,39 @@
 {{-     $tbIngestPassword = randAlphaNum 48 -}}
 {{-   end -}}
 {{- end -}}
+{{- /*
+  Bootstrap re-parent password (backend#1528 S3). DB_BOOTSTRAP_PASSWORD is the
+  password jobs-manager authenticates with to run the account-minting DDL
+  (CREATE USER / GRANT for tb_credmgr / tb_meta / tb_ingest). S3 re-parents that
+  mint OFF the root-equivalent edgeuser onto MySQL root (client-runtime#310), so
+  edgeuser can finally be REVOKEd / DROPped.
+
+  Resolution differs DELIBERATELY from the accounts above — jobs-manager
+  AUTHENTICATES as this identity, it never mints or ALTERs it:
+    1. explicit .Values.bootstrapDbPassword (the real root password — `Edg9@Tr@ce`
+       on an unrotated image, or the rotated value for the S3 window), else
+    2. the value already stored in the live Secret (preserve across upgrades), else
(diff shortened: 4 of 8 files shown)


saadqbal (commented):
Holding the review rather than posting one: this is `CONFLICTING` / `mergeStateStatus: DIRTY`, so the diff I'd be reading isn't the diff that will land, and reviewing the wrong three-way merge wastes both our time.

**The cause is a stale base, not a single unlucky collision.** The Chart.yaml hunk reads:

```
-version: 1.9.51
+version: 1.9.52
```

while `origin/develop` is at **1.9.58** — seven chart versions ahead. The PR was opened today, so the branch was cut from a checkout that had gone stale rather than from the current tip; `#764`, `#777`, `#779`, `#780`, `#781` and `#782` have all landed in between, and `#779` and `#782` both touch files this PR touches.

`git fetch && git merge origin/develop`, re-bump to whatever is current at that point, then re-request me — I'll pick it up on the next pass. Worth doing before anything else, because two of those intervening PRs changed `jobs-manager-deployment.yaml` and `secrets.yaml`, so the merge may well change what this PR needs to say rather than just where its lines sit.

LukasWodka (commented):
Picked this up as reviewer. **It needs a `develop` merge before I can review it properly** — `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`, confirmed on two consecutive reads so it isn't GitHub's "not yet computed" state.

It isn't only the version bump. Every file this PR touches has also moved on `develop` since your merge-base:

```
client/Chart.yaml
client/templates/_helpers.tpl
client/templates/jobs-manager-deployment.yaml
client/values.schema.json
client/values.yaml
```

`Chart.yaml` is the loudest — you set **1.9.52**, `develop` is at **1.9.60** — but `jobs-manager-deployment.yaml` and `_helpers.tpl` are the ones that matter, because the telemetry work landed in both today. I'm deliberately **not** doing the correctness pass yet: with those two re-resolved, I'd be reviewing lines that won't be what ships, and a review of a superseded diff is worse than no review because it reads as coverage.

## A systemic note, since this is the third one today

`chart-version-guard` requires a `Chart.yaml` bump whenever chart content changes — correct, and I verified it's a real required gate. But it means **any two concurrent chart PRs conflict by construction**, and the one that lands second always has to re-cut. Today that has cost:

- **client#765** — closed outright for exactly this (its 1.9.58→1.9.59 became a certain conflict once #784 took 59 and #787 took 60);
- **client#785** — this PR, now dirty;
- and #784/#787 themselves had to serialize.

That's not an argument against the guard. It's an argument that the bump wants to be the *last* thing computed rather than a hand-picked number carried on the branch — a release-time or merge-queue-time bump would remove the whole class. Worth a ticket; it's now demonstrably eating PRs rather than just being untidy.

## One question that survives the rebase

The two design decisions you flagged as confirmed on the ticket both look right to me, and neither depends on the merge: **require-pin/fail-fast over a baked `Edg9@Tr@ce` fallback** (keeps the literal out of the chart, doesn't fight backend#947), and the **ByEnv gate mirroring `serviceDbAccounts`** (the backend#1752 lesson that a plain global bool broke the fleet silently). Default-off everywhere with `helm template` showing a byte-identical default render is the right shape for a prod-irreversible-adjacent step.

What I'd want to understand is the **credential lifetime**. As described, `DB_BOOTSTRAP_USER=root` and `DB_BOOTSTRAP_PASSWORD` land as env vars on the jobs-manager deployment — so MySQL `root` is present in the pod environment for the whole process lifetime, not only while the account-minting DDL runs. The mint is a startup-time operation; the credential is permanent.

I'm not saying that's wrong — re-parenting onto `root` is clearly the right call, and your reasoning for why `edgeuser` can't revoke itself while the mint authenticates as it is correct. But it does trade "root-equivalent `edgeuser` in the pod" for "actual `root` in the pod", and the S3 endgame is `DROP USER edgeuser`. So: is the intent that jobs-manager holds `root` permanently from then on, or is a narrower minting identity (global `CREATE USER` + `GRANT OPTION`, nothing else) the eventual target? If the latter, worth saying so in the PR so the next reader doesn't take permanent-root as the settled destination.

Happy to do the full pass as soon as it's merged up — no need to re-request, it'll come back to me automatically.

saadqbal (commented):
Still holding — the re-request came without the merge. Head is unchanged at `18ba65a1`, the same commit I read on 08-21, and the PR is still `CONFLICTING` / `mergeStateStatus: DIRTY`. Re-requesting a review doesn't clear a conflict, so there's nothing new here for me to read.

The drift has grown since I flagged it: `origin/develop` is now at **1.9.60**, this branch is at **1.9.52** — eight chart versions, not seven. `#784` and `#787` landed in the meantime.

Concretely, five of your eight files have also moved on `develop`:

```
client/Chart.yaml
client/templates/_helpers.tpl
client/templates/jobs-manager-deployment.yaml
client/values.schema.json
client/values.yaml
```

`#784` ("the chart wires jobs-manager to the Collector's token") edits the same `jobs-manager-deployment.yaml` this PR is wiring `DB_BOOTSTRAP_*` into — which is why I'd rather not review the pre-merge diff. The merge is likely to change *what* this PR needs to say, not just where its lines sit.

Unchanged ask: `git fetch && git merge origin/develop`, re-bump off whatever the tip is at that point, push, then re-request. I'll review the merged diff on the next pass.

